PR #552, opened 2020-04-07
Created Group Target For All Tests

garethellis0:
### Checklist


garethellis0:
We can just have all the tests for a given board built into a single executable, which should work fine, given how little code we have.

PR #553, opened 2020-04-07
Moved The Generated CAN Code To The Build Folder

garethellis0:
This way we can stop manually deleting `auto_generated` whenever we want
to do a clean build.... :)

### Checklist


garethellis0:
Ah yes you're right. Doesn't delete the `auto_generated` folder but does delete the files inside of it. Regardless, I think that it's best practices to only have user-modifiable code outside the build folder, because less code makes it easier to find whatever you're trying to change.

garethellis0:
I'd advocate for dropping `auto_generated` from the path because the user really shouldn't need to worry about whether it's auto-generated or not, and it's fairly simple for them to check. I see why we might want to have it to make sure the user doesn't think they need to change it, but I don't think it's worth the mental overhead. Just my two cents.

thekenu:
Hm yeah I think I agree if we're

1. Dropping `auto_generate` from `#include`, and
2. Moving the generated code into the build output folder, and
3. Having code warning about the files being auto-generated

, then it should be obvious for the user know that these file are not meant to be modified.

Feel free to drop `auto_generate` from `#include` as well as the output path :+1:

thekenu:
Replaced by #560

PR #554, opened 2020-04-07
Fix error where Travis CI doesn't report error when ctest fails

thekenu:
### Summary

CI was not failing when `ctest` returns non-zero exit code :cry: 

### Checklist

### DIFF
--- a/scripts/travis_ci/travis_script.sh
+++ b/scripts/travis_ci/travis_script.sh
@@ -28,7 +28,7 @@ if [ "$RUN_X86_TESTS" = "true" ]; then
     BUILD_DIR=boards/x86_build
     travis_run cmake -S boards -B $BUILD_DIR -DPLATFORM=x86
     travis_run make --directory=$BUILD_DIR
-    travis_run cd $BUILD_DIR && ctest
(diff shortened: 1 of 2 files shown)


scripts/travis_ci/travis_script.sh
```diff
@@ -28,7 +28,14 @@ if [ "$RUN_X86_TESTS" = "true" ]; then
     BUILD_DIR=boards/x86_build
     travis_run cmake -S boards -B $BUILD_DIR -DPLATFORM=x86
     travis_run make --directory=$BUILD_DIR
-    travis_run cd $BUILD_DIR && ctest && cd -
+    travis_run "cd $BUILD_DIR && ctest"
+
+    if [ $? -ne 0 ]; then
```
garethellis0, line 33: Nit; can we capture the return value in a variable and check that instead? I think it makes the code more readable and less error-prone:
```
CTEST_RETURN_CODE=$?
if [ $CTEST_RETURN_CODE -ne 0 ]; then
...
```

thekenu:
Re-opening this in #557 since I messed up the forks.

PR #555, opened 2020-04-08
Add a simple IMD unit test

thekenu:
### Summary

Add IMD test case to verify the conversion from frequency to IMD condition is correct.

### Checklist


boards/BMS/Inc/App/App_Imd.h
```diff
@@ -2,21 +2,55 @@
...
  * @get_pwm_duty_cycle: A function that can be called to get the duty cycle of
  *                      the IMD's PWM output
  * @get_seconds_since_power_on: A function that can be called to get the time of
  *                              the system, in milliseconds
  * @return A pointer to the created IMD, whose ownership is given to the caller
  */
 struct Imd *App_Imd_Create(
-    struct BmsCanTxInterface *can_tx,
     float (*get_pwm_frequency)(void),
+    float (*get_pwm_frequency_tolerance)(void),
```
thekenu, line 53: Transferring the unresolved comment from https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/536#discussion_r405608157

boards/BMS/Src/App/states/App_DriveState.c
```diff
@@ -1,15 +1,87 @@
...
 }
 
 static void DriveStateRunOnTick(struct StateMachine *state_machine)
 {
-    UNUSED(state_machine);
+    struct BmsWorld *world = App_SharedStateMachine_GetWorld(state_machine);
+    struct BmsCanTxInterface *can_tx = App_BmsWorld_GetCanTx(world);
+    struct Imd *              imd    = App_BmsWorld_GetImd(world);
+
+    App_DriveState_SetImdPeriodicCanSignals(can_tx, imd);
```
thekenu, line 84: Transferring unresolved comment from https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/536#discussion_r405611007

boards/BMS/Test/Inc/imd_fff.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
```
thekenu, line 1: Transferring unresolved comment from https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/536#discussion_r405192994

garethellis0: Seems reasonable to me, I'd say go for it.

thekenu:
@garethellis0 Addressed all the comments you left in #536, but let's keep all the new comments in this PR now since this is the PR that will contain the changes i made!

boards/BMS/Test/Src/state_machine_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+    // We need to tick twice, once to run the `Init` state, and once more
+    // to have the state machine transition to the `Run` state.
+    App_SharedStateMachine_Tick(state_machine);
+    App_SharedStateMachine_Tick(state_machine);
+
+    EXPECT_EQ(
+        App_GetDriveState(),
+        App_SharedStateMachine_GetCurrentState(state_machine));
+}
```
garethellis0, line 71: Should we test that we're properly setting all the CAN messages associated with the IMD?

boards/DIM/CMakeLists.txt
```diff
@@ -31,19 +31,21 @@ set(STM32_DEVICE_DEFINE STM32F302xC)
...
-        "${ARM_BINARY_IO_SRCS}"
-        "${ARM_BINARY_IO_INCLUDE_DIRS}"
-        "${ARM_BINARY_IO_SYSCALLS_SRC}"
-        "${ARM_LINKER_SCRIPT}"
-        "${STM32_DEVICE_DEFINE}"
-        "${STM32CUBEMX_DRIVER_SRCS}"
-        "${STM32CUBEMX_DRIVER_INCLUDE_DIRS}"
-        "${GOOGLETEST_TEST_SRCS}"
-        "${GOOGLETEST_MAIN_SRC}"
+    "DIM"
```
thekenu, line 37: nit; Just realized the indentation here was messed up

boards/BMS/Test/Src/state_machine_test.cpp
```diff
@@ -62,6 +57,16 @@ class BmsStateMachineTest : public testing::Test
             App_SharedStateMachine_GetCurrentState(state_machine));
     }
 
+    std::vector<const struct State *> GetAllStates(void)
```
thekenu, line 60: Came up with this helper function for us to test something in all states. I felt that if we were to test whether a CAN message is sent correctly in all 4 states, splitting that across 4 `TEST_F` would be quite redundant.

boards/BMS/Test/Src/imd_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+    void SetPwmFrequencyTolerance(float tolerance)
+    {
+        ASSERT_TRUE(imd != NULL);
+        App_Imd_Destroy(imd);
+        imd = App_Imd_Create(
+            get_pwm_frequency, tolerance, get_pwm_duty_cycle,
+            get_seconds_since_power_on);
+    }
+
+    void SetImdCondition(enum Imd_Condition condition)
```
thekenu, line 42: @garethellis0  Had to write this function twice, once in `imd_test.cpp` and once in `state_machine_tests.cpp`. Is there an elegant way to re-use it?

thekenu:
@garethellis0 Added a bunch of tests to exercise the CAN signals. Have a look and let me know what you think!

boards/BMS/Test/Src/imd_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    for (auto &condition : conditions)
+    {
+        SetImdCondition(condition);
+
+        // From ISOMETER® IR155-3203/IR155-3204 manual:
+        //     Insulation Resistance =
+        //         (90% * 1200kOhm) / (duty_cycle - 5%) - 1200kOhm
+
+        constexpr float MAX_VALID_DUTY_CYCLE = 95.0f;
```
thekenu, line 73: General testing strategy for all the tests: Define the `MIN/MAX` duty cycle at the top. Then exercise `0.1` below `MIN`, `MIN`, `(MIN+MAX)/2`, `MAX`, `0.1` above`MAX`.

boards/BMS/Test/Src/imd_test.cpp
```diff
@@ -0,0 +1,267 @@
+#include "fff.h"
+#include <gtest/gtest.h>
+#include <utility>
```
garethellis0, line 3: nit; this should be `Imd_test.cpp` to be consistent with the naming of our other files.

boards/BMS/Test/Src/imd_test.cpp
garethellis0: Yep, you can define a class `BmsTest` that extends `public testing::Test`, then both `ImdTest` and `StateMachineTest` can extend `public BmsTest`. Then you can put shared code in `BmsTest`.

boards/BMS/Test/Src/imd_test.cpp
thekenu: Yep I can do that. However, right now the names are kinda all over  the place
```
./DIM/Test/Src/seven_seg_displays_test.cpp
./DIM/Test/Src/state_machine_test.cpp
./PDM/Test/Src/voltage_monitor_tests.cpp
./PDM/Test/Src/state_machine_test.cpp
./FSM/Test/Src/state_machine_test.cpp
./DCM/Test/Src/brake_light_tests.cpp
./DCM/Test/Src/state_machine_test.cpp
./BMS/Test/Src/imd_test.cpp
./BMS/Test/Src/state_machine_test.cpp
```

Might be worth it to clean up everything in one pass after this PR as Test_{ModuleName}.cpp?

That way we have `{App|Io}_{ModuleName}.c` for the application and IO layer code, and `Test_{ModuleName}.cpp` for all test files?

boards/BMS/Test/Src/imd_test.cpp
thekenu: Played around with this slightly but unsure  about a couple things

Points I considered:

- Should `struct Imd* imd` just live in `BmsTest`? Or should `StateMachineTest` and `ImdTest` hold its own copy of `sturct Imd* imd`
- Depending on the last point, I would also need to change the function signature for `SetImdCondition`

```
// If "struct Imd* imd" lives in class BmsTest
void BmsTest::SetImdCondition(enum Imd_Condition condition);

// If StateMachineTest and ImdTest hold their own "struct Imd* imd"
void BmsTest::SetImdCondition(struct Imd* imd, enum Imd_Condition condition)
```
- It gets a little more annoying than that actually because `SetImdCondition` needs to modify `{fff_functio}.return_val`, so the most correct function signatures would be 
``` 
// If "struct Imd* imd" lives in class BmsTest
void BmsTest::SetImdCondition(float& fake_frequency_return_value, enum Imd_Condition condition);

// If StateMachineTest and ImdTest hold their own "struct Imd* imd"
void BmsTest::SetImdCondition(struct Imd* imd, float& fake_frequency_return_value, enum Imd_Condition condition)
```

Anyway, I wrote up a [quick draft](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/555/commits/ee105058bba125775e424cee9ab2feacbdb35330) that works. Feel free to leave any comments there.

boards/BMS/Test/Inc/Bms_Test.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
thekenu, line 1: If we left the class definition in `Test/Src/Bms_Test.cpp`, the build system would get confused because it treats each `.cpp` (that is not `main.cpp`) as its own separate, compilable Gtest binary. I don't imagine this file getting too long anyway so I think leaving it header-only is fine for now.

boards/BMS/Test/Src/imd_test.cpp
garethellis0: Looks good to me!

boards/BMS/Test/Src/imd_test.cpp
garethellis0: Seems reasonable to me.

boards/BMS/Test/Inc/Bms_Test.h
garethellis0: Uhhhh, we should probably not be doing it like that. In general, we should compile all the `test` files into a single executable, and having a single (shared) `gtest_main.cpp` file that has the `main` function in it. Does that make sense?

Looks like you don't even need to create `gtest_main.cpp`, gtest provides one for you: https://stackoverflow.com/a/21479008

thekenu: Oh ok yeah I get what you mean. Can we have one `main` function per board? Or did you mean one `main` function for every board (i.e. Entire project)?

boards/BMS/Test/Src/imd_test.cpp
thekenu: Ok cool I would prefer to do the clean up after this PR then since it's getting a bit long :+1:

thekenu: I changed my mind and decided to just clean up all the test files in this PR anyway.

boards/CMakeLists.txt
```diff
@@ -253,37 +253,36 @@ function(can_code_generation
...
-            PUBLIC
-                -Wall
-                -Werror
-                )
-        target_link_libraries(${TEST_EXECUTABLE_NAME} gtest_main)
-        add_test(NAME ${TEST_EXECUTABLE_NAME}
-                COMMAND ${TEST_EXECUTABLE_NAME})
-    endforeach()
+
+    set(TEST_EXECUTABLE_NAME "${BOARD_NAME}_test")
```
thekenu, line 264: We now only compile 
```
DCM_test
BMS_test
PDM_test
DIM_test
FSM_test
```
Rather than one executable per `_test.cpp`.

boards/FSM/Test/Src/Test_StateMachine.cpp
```diff
@@ -1,13 +1,12 @@
-#include "fff.h"
-#include "gtest/gtest.h"
+#include <fff.h>
```
thekenu, line 1: Decided to include fff.h and gtest.h using `<>` becuase they're code we don't modify.

boards/CMakeLists.txt
```diff
@@ -430,10 +428,10 @@ function(create_arm_binary_or_tests_for_board
     git_hash_code_generation(${GIT_HASH_HEADER})
 
     if ("${PLATFORM}" STREQUAL "x86")
-        compile_googletest_executables(
+        compile_googletest_executable(
```
thekenu, line 431: Renamed to a singular "executable", because while, we only compile one executable per board now. (Just making it clear that the renaming was intentional)

boards/BMS/Test/Inc/Bms_Test.h
thekenu: I went ahead and created one test executable per board. All the `main.cpp` are purged as well.

garethellis0: Yep, that's what I was thinking of.

PR #558, opened 2020-04-09
Add application code for displaying SoC on 7-segment displays

thekenu:
### Summary

- Add application code for the 3 x 7-segment display on DIM with some basic unit tests.
- Add skeleton IO code to be filled in later

### Checklist


boards/DCM/Src/main.c
```diff
@@ -25,7 +25,7 @@
 /* Private includes ----------------------------------------------------------*/
 /* USER CODE BEGIN Includes */
 #include "Io_SharedSoftwareWatchdog.h"
-#include "Io_SharedConstants.h"
+#include "App_SharedConstants.h"
```
thekenu, line 28: Constants should never really be platform-specific. So changing it to `App_`

thekenu:
@garethellis0 Can you give this a review pass without the jdocs? i.e. Just review the functional aspect of this PR? Then I'll add in the jdocs afterwards.

boards/DIM/Inc/App/App_SevenSegDisplays.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+struct SevenSegDisplay;
+
+// Do NOT take ownership of individual displays
+struct SevenSegDisplays *App_SevenSegDisplays_Create(
```
thekenu, line 6: I was having a hard time deciding what should be put into this struct.

```
DIM-9 | 7-segment | - The DIM must display the SoC as percentage on the 7-segment displays if no faults are present.  - If a fault has occurred the DIM must stop displaying the SoC and instead display any faults onto the 7-segment displays. - The first 7-segment must display the board ID while the remaining two must display the fault ID.   - If there are more than one fault active, the DIM must cycle through displaying each present fault at 1Hz.
```

I thought, the DIM must display the SoC so I will add a getter function for that. However, the 7-seg displays must also be smart enough to display SoC instead of faults when needed. How should I be tweaking `struct SevenSegDisplays` for it to be smart enough?

boards/DIM/Src/Io/Io_SevenSegDisplays.c
```diff
@@ -0,0 +1,175 @@
+#include "App_SharedAssert.h"
+#include "App_SharedConstants.h"
+
+#include "Io_SevenSegDisplays.h"
+
+float Io_SevenSegDisplay_GetStateOfCharge(void)
```
thekenu, line 6: The functions are left intentionally unfilled for now

boards/DIM/Test/Src/seven_seg_displays_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+        App_SevenSegDisplays_Destroy(seven_segment_displays);
+    }
+
+    struct SevenSegDisplay * left_display;
+    struct SevenSegDisplay * middle_display;
+    struct SevenSegDisplay * right_display;
+    struct SevenSegDisplays *seven_segment_displays;
+};
+
+TEST_F(SevenSegDisplaysTest, write_max_state_of_chrage)
```
thekenu, line 49: Happy to add more unit tests. But can't think of any more things to test for now. Suggestions are welcomed.

boards/DIM/Test/Src/seven_seg_displays_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+    ASSERT_EQ(1, set_left_hex_digit_fake.arg0_history[0]);
+    ASSERT_EQ(0, set_middle_hex_digit_fake.arg0_history[0]);
+    ASSERT_EQ(0, set_right_hex_digit_fake.arg0_history[0]);
+
+    ASSERT_EQ(1, get_state_of_charge_fake.call_count);
+    ASSERT_EQ(1, set_left_hex_digit_fake.call_count);
+    ASSERT_EQ(1, set_middle_hex_digit_fake.call_count);
+    ASSERT_EQ(1, set_right_hex_digit_fake.call_count);
+
+    ASSERT_EQ(fff.call_history[0], (void *)get_state_of_charge);
```
thekenu, line 64: Wasn't sure if the `call_history` is excessive or not :thinking:

boards/DIM/Inc/App/App_SevenSegDisplays.h
thekenu: I think eventually I want to remove `WriteStateOfCharge` and just have `Tick` function

boards/DIM/Src/main.c
```diff
@@ -77,7 +80,19 @@ void        StartDefaultTask(void const *argument);
 int main(void)
 {
     /* USER CODE BEGIN 1 */
+    struct SevenSegDisplay *left_seven_seg_display =
+        App_SevenSegDisplay_Create(Io_SevenSegDisplay_SetLeftHexDigit);
 
+    struct SevenSegDisplay *middle_seven_seg_display =
+        App_SevenSegDisplay_Create(Io_SevenSegDisplay_SetLeftHexDigit);
```
markomarkusi, line 87:
```suggestion
        App_SevenSegDisplay_Create(Io_SevenSegDisplay_SetMiddleHexDigit);
```

boards/DIM/Src/main.c
```diff
@@ -77,7 +80,19 @@ void        StartDefaultTask(void const *argument);
...
+
+    struct SevenSegDisplay *right_seven_seg_display =
+        App_SevenSegDisplay_Create(Io_SevenSegDisplay_SetLeftHexDigit);
```
markomarkusi, line 90:
```suggestion
        App_SevenSegDisplay_Create(Io_SevenSegDisplay_SetRightHexDigit);
```

boards/DIM/Src/Io/Io_SevenSegDisplays.c
```diff
@@ -0,0 +1,264 @@
...
+static_assert(HEX_DIGIT_E == 14, "Hex enum must match its numeric value.");
+static_assert(HEX_DIGIT_F == 15, "Hex enum must match its numeric value.");
+
+float Io_SevenSegDisplay_GetStateOfCharge(void)
+{
+    // TODO: Read the state of charge from CAN
+    return 0.0f;
+}
+
+void Io_SevenSegDisplay_SetLeftHexDigit(uint8_t hex_digit)
```
markomarkusi, line 29: With how we have the shift registers/ seven segs wired on the DIM we can not just change one seven segment at a time. The shift registers are daisy chained together you need to shift 3 x 8 bits over spi to update them all. We will need a LUT which will map each hex number to the corresponding bits needed to represent that number on seven segment display. After we have the bit representation of each digit, we would assign the elements in SPI tx data buffer, for example:

```
uint8_t spi_tx_data[3];
...
spi_tx_data[0] = right_hex_digit;
spi_tx_data[1] = middle_hex_digit;
spi_tx_data[2] = left_hex_digit;

HAL_SPI_Transmit(&hspi1, spi_tx_data, 3U, 100U);

Comment (markomarkusi) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c:29]: On second thought this function could be used to update the left hex digit in the tx buffer:
`spi_tx_data[2] = left_hex_digit;`
and then the state machine would periodically be calling 
`HAL_SPI_Transmit(&hspi1, spi_tx_data, 3U, 100U);`
which in turn would update that specific display?

Comment (thekenu) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c:29]: Yep that could be one way to do it. I think the approach is to write clean application-level code that is independent of hardware technicalities such as shift registers). The we will write IO-level code to accomdate the application-level code.

```
Task100Hz {
  App_SevenSegDisplays_Tick(); //Update spi_tx_data[0-2] using Io_SevenSegDisplay_Set{Left|Middle|Right}HexDigit
  Io_SevenSegDisplays_Tick(); // HAL_SPI_Transmit
}
```

Comment (markomarkusi) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c:29]: Ya, that sounds good then👍

Comment (thekenu) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c:29]: I'll tweak the code a bit to work with the SPI data better

File: boards/DIM/Src/Io/Io_SevenSegDisplays.c [hunk 9]
```diff
@@ -0,0 +1,93 @@
...
+    0x61, // 0xE
+    0x71, // 0xF
+};
+
+void Io_SevenSegDisplays_Init(SPI_HandleTypeDef hspi)
+{
+    _hspi = hspi;
+}
+
+void Io_SevenSegDisplays_WriteCommands(void)
```
thekenu, line 60: @garethellis0 Where would this function go? Registering it in `App` seems to go against the abstraction because this is closely tied to hardware

boards/DIM/Src/Io/Io_SevenSegDisplays.c
```diff
@@ -0,0 +1,93 @@
...
+{
+    // TODO: Read the state of charge from CAN
+    return 0.0f;
+}
+
+void Io_SevenSegDisplays_SetLeftHexDigit(uint8_t hex_digit)
+{
+    shared_assert(hex_digit < NUM_HEX_DIGITS);
+
+    commands.left = command_lookup_table[hex_digit];
```
thekenu, line 78: Since we can't just set each 7-seg individually, I decided to store the commands in an array. The commands would wait to be transmitted via SPI at a later time.

boards/DIM/Src/Io/Io_SevenSegDisplays.c
thekenu: @markomarkusi Feel free to take a look at the code I rewrote and see if it makes more sense now with the shift register design

boards/DIM/Src/App/App_SevenSegDisplays.c
```diff
@@ -0,0 +1,55 @@
+#include <stdlib.h>
+
+#include "App_SevenSegDisplays.h"
+#include "App_SharedAssert.h"
+
+struct SevenSegDisplays
+{
+    float (*get_state_of_charge)(void);
+    void (*set_left_hex_digit)(uint8_t);
```
thekenu, line 9: Previously, I created struct for **individual** 7-seg displays. And instead of `void (*set_left_hex_digit)(uint8_t);` I had `struct SevenSegDisplay* left_display` but later I realized `struct SevenSegDisplay` only has one method function.

Considering how much boilerplate code I'd have to add for `struct SevenSegDisplay`, I decided to just replace that with a simple setter function pointer here.

boards/DIM/Inc/App/App_SevenSegDisplays.h
garethellis0: If this is an abstraction around the 7-segment display I don't think it should know anything about the SoC. In general I would expect to see something like:
```
App_SevenSegDisplay_SetValue(seven_seg_display, App_SoC_GetSoCAsString(soc_struct));
```
probably in the state machine (somewhat pseudo-code, but you get the idea). So really the SevenSeg just can set any characters we give it, and we get the SoC as a string and set it. In the case where there's a fault we should just set the corresponding characters. It probably makes sense to have some sort of abstraction around setting faults, just not in the SevenSeg code. 

More generally, the point I'm trying to make here isn't about interface details, but more about what information the SevenSeg logically encapsulates.

Does that make sense? 


edit: After looking through this more, a more reasonable interface might be `App_SevenSegDisplay(struct SevenSeg, HexDigit chars[], size_t num_chars)`.

boards/DIM/Src/Io/Io_SevenSegDisplays.c
garethellis0: Yeah, seems like this should live in IO.

boards/DIM/Src/Io/Io_SevenSegDisplays.c
```diff
@@ -0,0 +1,93 @@
...
+    // can't update them individually. Instead, we must update the 7-segment
+    // displays all at once.
+    HAL_SPI_Transmit(_hspi, (uint8_t *)&commands, 3U, 100U);
+}
+
+float Io_SevenSegDisplays_GetStateOfCharge(void)
+{
+    // TODO: Read the state of charge from CAN
+    return 0.0f;
+}
```
garethellis0, line 72: Similar comment here, the SevenSeg doesn't need to know anything about the SoC.

boards/DIM/Test/Src/seven_seg_displays_test.cpp
garethellis0: I think it's probably excessive. Let's consider what we really want to check here. We just want to check that the `left` `middle` and `right` setters were called with the values that we expect. The rest of the stuff we're checking here could change and we'd still consider the function "correct", which is a good indicator that we shouldn't be checking it.

boards/DIM/Test/Src/seven_seg_displays_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+        "");
+}
+
+TEST_F(SevenSegDisplaysTest, write_underflow_state_of_charge)
+{
+    get_state_of_charge_fake.return_val = -1.0f;
+
+    ASSERT_DEATH(
+        { App_SevenSegDisplays_WriteStateOfCharge(seven_segment_displays); },
+        "");
```
garethellis0, line 117: Do we really want to hardfault if we try to write an invalid SoC to the seven seg? It seems to me that a more reasonable course for something non-critical like this would be to return a value that we can check, and then log a CAN message if the returned value indicates the sevenseg could not write what we asked it to.

boards/DIM/Src/App/App_SevenSegDisplays.c
garethellis0: Depends if you think this will ever change or not. Personally I'd be inclined to have a struct per-display, but I'm fine with this as well.

boards/DIM/Inc/App/App_SevenSegDisplays.h
thekenu: I think I get what you're saying. Just one small question though: Is `App_SevenSegDisplay(struct SevenSeg, HexDigit chars[], size_t num_chars)` supposed to represent all three 7-segment displays? And we would have to check in the function that num_chars is `<= 3` because we don't have the hardware to write more than three characters?

boards/DIM/Test/Src/seven_seg_displays_test.cpp
thekenu: Yeah I think you are right about this one. Removing all things call history related.

boards/DIM/Test/Src/seven_seg_displays_test.cpp
thekenu: I think you're right on this one and it really speaks on how we avoid error codes like the plague in our codebase. Something I've really wanted to start doing is introducing generic error codes that can be shared across the modules.

Something like
```
// StatusCodes for various errors, keep these generic. Never assume their order is fixed so refer to
// them by name only.
typedef enum {
  STATUS_CODE_OK = 0,
  STATUS_CODE_UNKNOWN,
  STATUS_CODE_INVALID_ARGS,
  STATUS_CODE_RESOURCE_EXHAUSTED,
  STATUS_CODE_UNREACHABLE,
  STATUS_CODE_TIMEOUT,
  STATUS_CODE_EMPTY,
  STATUS_CODE_OUT_OF_RANGE,
  STATUS_CODE_UNIMPLEMENTED,
  STATUS_CODE_UNINITIALIZED,
  STATUS_CODE_INTERNAL_ERROR,
  NUM_STATUS_CODE
```
And then we can `ASSERT_EQ` on return codes rather than `DEATH`. What do you think? If you think it's sane I would like to add `error_code.h` in this PR.

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Hmm it's one extra level of indrection, i.e.
```
SevenSegDisplays
  SevenSegDisplay
    Io_SeLeftDigit
```
v.s. 
```
SevenSegDisplays
    Io_SeLeftDigit
```
But I guess the former is more logically sound? I'll just sit on this a little more and maybe revert the changes. But yeah I'm fine with either option too atm.

boards/DIM/Src/Io/Io_SevenSegDisplays.c
thekenu: Removed!

boards/DIM/Src/Io/Io_SevenSegDisplays.c
```diff
@@ -0,0 +1,93 @@
...
+    0x1F, // 0x7
+    0x01, // 0x8
+    0x09, // 0x9
+    0x11, // 0xA
+    0x1C, // 0xB
+    0x63, // 0xC
+    0x85, // 0xD
+    0x61, // 0xE
+    0x71, // 0xF
+};
```
markomarkusi, line 53: Noice this is what I was thinking as well. The bits should be inverted though as a 1 represents a segment being on. I also believe the bit order should be reversed (0 should be 0xC0) unless i'm getting confused with MSB/ LSB first😵

image

thekenu: Hmm ahh yeah I just copied these values from the older PR. Can you help me calculate the new lookup table? Then I'll add it to this PR. And I'll also leave a comment saying that these values aren't tested yet. We can do that at a later time.

boards/DIM/Inc/App/App_SevenSegDisplays.h
thekenu: I took my own assumptions and wrote 

```
void App_SevenSegDisplays_SetValue(
    const struct SevenSegDisplays *seven_seg_displays,
    uint8_t                        chars[],
    size_t                         num_chars);
```

Feel free to take a look when you get the chance. I also took this function and tested every one of the possible 16 * 16 * 16 = 4096 values.

boards/DIM/Src/Io/Io_SevenSegDisplays.c
markomarkusi, line 53:
```suggestion
static const uint8_t command_lookup_table[NUM_HEX_DIGITS] = {
    0x3F, // 0x0
    0x06, // 0x1
    0x5B, // 0x2
    0x4F, // 0x3
    0x64, // 0x4
    0x6D, // 0x5
    0x7D, // 0x6
    0x07, // 0x7
    0x7F, // 0x8
    0x67, // 0x9
    0x77, // 0xA
    0x7C, // 0xB
    0x39, // 0xC
    0x5E, // 0xD
    0x79, // 0xE
    0x71, // 0xF
};
```

markomarkusi: Did we want to place this in a function?
```
uint8_t get_command_data(uint8_t hex_digit){
    static const uint8_t command_lookup_table[NUM_HEX_DIGITS] = {
        0x3F, // 0x0
        0x06, // 0x1
        0x5B, // 0x2
        0x4F, // 0x3
        0x64, // 0x4
        0x6D, // 0x5
        0x7D, // 0x6
        0x07, // 0x7
        0x7F, // 0x8
        0x67, // 0x9
        0x77, // 0xA
        0x7C, // 0xB
        0x39, // 0xC
        0x5E, // 0xD
        0x79, // 0xE
        0x71, // 0xF
        };
    return command_lookup_table[hex_digit];
}

Comment (markomarkusi) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c]: To set the decimal point the command (ie: 0x3F, // 0x0) would need to be or'd with 0x80. If we put the LUT in a getter function and then have a add decimal point boolean parameter we could or the LUT value with 0x80 at before returning if we want a decimal point?

Comment (thekenu) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c]: Don't feel very strongly about it because I don't see what value it adds at the moment and would rather keep it simple :thinking:

Comment (markomarkusi) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c]: Ya sure. Where would we want to enable the decimal point then?

Comment (thekenu) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c]: I was thinking about this too. Do we actually need to use the decimal point? I was reading the specs but didn't see an explicit need for it

Comment (markomarkusi) [reply boards/DIM/Src/Io/Io_SevenSegDisplays.c]: Ya we could get away without it, it would only be used for SoC. When cycling between displaying the SoC and speed, the decimal point on the SoC could help distinguishing between the two, but again it's not necessary.

File: boards/DIM/Src/App/App_SevenSegDisplays.c (hunk 11 above)
Comment (garethellis0) [reply boards/DIM/Src/App/App_SevenSegDisplays.c]: Yeah, I prefer the first option here.

File: boards/DIM/Test/Src/seven_seg_displays_test.cpp (hunk 13 above)
Comment (garethellis0) [reply boards/DIM/Test/Src/seven_seg_displays_test.cpp]: Sure, that's fine by me, this is essentially our equivalent of `HAL_Status`

File: boards/DIM/Src/App/App_SevenSegDisplays.c [hunk 15]
```diff
@@ -0,0 +1,51 @@
...
+    const struct SevenSegDisplays *const seven_seg_displays,
+    const uint8_t                        chars[],
+    size_t                               num_chars)
+{
+    shared_assert(num_chars <= NUM_SEVEN_SEG_DISPLAYS_DIGITS);
+
+    for (size_t i = 0; i < num_chars; i++)
+    {
+        shared_assert(chars[i] < NUM_HEX_DIGITS);
+        seven_seg_displays->set_hex_digit[i](chars[i]);
```
garethellis0, line 49: What if I only give one character? Do the other ones remain unchanged?

boards/DIM/Test/Src/seven_seg_displays_test.cpp
```diff
@@ -0,0 +1,108 @@
...
+        input[i] = NUM_HEX_DIGITS;
+        ASSERT_DEATH(
+            {
+                App_SevenSegDisplays_SetValue(
+                    seven_segment_displays, input,
+                    NUM_SEVEN_SEG_DISPLAYS_DIGITS);
+            },
+            "");
+    }
+}
```
garethellis0, line 108: Can we add a test for setting all characters, and then setting only a single character? Just want to make sure this has well-defined behavior.

boards/shared/Inc/App/App_SharedConstants.h
```diff
@@ -0,0 +1,33 @@
...
+    HEX_DIGIT_8,
+    HEX_DIGIT_9,
+    HEX_DIGIT_A,
+    HEX_DIGIT_B,
+    HEX_DIGIT_C,
+    HEX_DIGIT_D,
+    HEX_DIGIT_E,
+    HEX_DIGIT_F,
+    NUM_HEX_DIGITS,
+};
```
garethellis0, line 33: A comment that order is important here would be good.

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Oh yeah that's a good point. I am unsure what the best approach is. I can thinking we can either:
1. Leave the others unchanged
2. Default the unspecified digits to be zero
3. Default the unspecified digits to be turned off
4. The user has to specify exactly 3 chars each time

Which one do you think makes more sense?

boards/DIM/Test/Src/seven_seg_displays_test.cpp
thekenu: For sure. Let's reach an agreement on what I mentioned [here](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/558#discussion_r406854415) then i'll go ahead and add the tests.

boards/shared/Inc/App/App_SharedConstants.h
thekenu: Added

boards/DIM/Src/Io/Io_SevenSegDisplays.c
thekenu: Let's cross that bridge when you get there and we can add decimal point is there a proven use-case

boards/DIM/Src/App/App_SevenSegDisplays.c
garethellis0: 3

thekenu: I think if we want to go with 3, we will need another `HEX_DIGIT` for turning a display off (i.e., HEX_DIGIT_OFF). That way I can write
```
 ASSERT_EQ(
                    HEX_DIGIT_OFF,
                    set_left_hex_digit_fake.arg0_history[count - 1]);
```

But then that kinda breaks `enum HexDigit` because now we have 17 digits (`0-15`, `OFF`) to handle. What's an elegant way to get around this issue? Maybe give `Io_SevenSegDisplay` its own unique `enum SevenSegHexDigit` that has 17 digits rather than relying on `enum HexDigit` in `SharedConstant.h`?

garethellis0:
```
struct SevenSegHexDigit {
 bool enabled;
 HexDigit value;
}
```
maybe?

thekenu: Hmm i think that's about as elegant as we can get while having a generic hex enum. I'll try adding this in and make sure the comment is clear about what `enabled` means :v:

boards/DIM/Test/Src/seven_seg_displays_test.cpp
thekenu: I added tests that write either 1 value, 2 values, or 3 values. Can you take a look and see if that's what you wanted to test?

boards/DIM/Test/Src/seven_seg_displays_test.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include "gtest/gtest.h"
+
+extern "C"
+{
+#include "App_SevenSegDisplays.h"
+#include "App_SevenSegDisplay.h"
+#include "App_ErrorCode.h"
+
+    DEFINE_FFF_GLOBALS;
+    FAKE_VOID_FUNC(set_right_hex_digit, struct SevenSegHexDigit);
```
thekenu, line 20: Wasn't sure if the argument should be a pointer or not. Maybe it doesn't matter too much because `struct SevenSegHexDigit` is quite small?

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Added `SevenSegDisplay` back in. You can see in `main.c` that I had to create left, middle, and right displays individually.

boards/DIM/Test/Src/seven_seg_displays_test.cpp
thekenu: Added some error codes under `App_ErrorCode.h`. Later I'd like to add some simple macros to log some extra information (e.g. `__LINE__`, custom message describing the error, etc) but that's for another PR

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Added the enabled bit. Essentially, if `enabled` is `false`, then `value ` is ignored entirely and we just turn off the display.

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -45,6 +45,9 @@ SG_ Frequency : 32|32@1- (1,0) [-3.4E038|3.4E038] "Hz" LOGGER
 BO_ 107 BMS_STATE_MACHINE : 1 BMS
 SG_ State : 0|8@1+ (1,0) [0|255] "" LOGGER
 
+BO_ 108 BMS_STATE_OF_CHARGE: 4 BMS
```
thekenu, line 48: Not strictly related to this PR, but we definitely want to display SoC on 7-seg at some point by reading the CAN bus. So I thought why not add this in.

boards/DIM/Inc/App/App_SevenSegDisplay.h
```diff
@@ -0,0 +1,20 @@
...
+    bool          enabled;
+    enum HexDigit value;
+};
+
+struct SevenSegDisplay *
+    App_SevenSegDisplay_Create(void (*set_hex_digit)(struct SevenSegHexDigit));
+
+void App_SevenSegDisplay_Destroy(struct SevenSegDisplay *seven_seg_display);
+
+void App_SevenSegDisplay_SetHexDigit(
```
thekenu, line 18: Renamed `SetValue` and the like to be `SetHexDigit` and `SetHexDigits` because I think it makes it much more obvious what this function is doing and what input it expects (i.e. It only writes hex values and not any ascii value)

boards/DIM/Inc/Io/Io_SevenSegDisplays.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <stdbool.h>
+#include "App_SevenSegDisplay.h"
+
+void Io_SevenSegDisplays_Init(SPI_HandleTypeDef *hspi);
+void Io_SevenSegDisplays_WriteCommands(void);
```
thekenu, line 7: @garethellis0 Got any idea where we should be calling this function? I gave it some thought but wasn't too sure.

thekenu:
Thanks all for the review :v:

PR #561, opened 2020-04-09
Use malloc for Create() and Destroy() in CAN TX/RX interface in the application layer

thekenu:
### Summary

Use `malloc` we can `Create` and `Destroy` in `googletest`

### Checklist


scripts/codegen/CAN/canrx_codegen.py
```diff
@@ -147,7 +151,7 @@ def __init_structs(self):
             'CAN RX interface')
 
     def __generateHeaderIncludes(self):
-        header_names = ['<string.h>',
```
thekenu, line 150: `string.h` is no longer needed because we don't need `size_t` anymore

PR #562, opened 2020-04-11
Bring up CAN hardware on DIM

thekenu:
### Summary

Add just enough code so that we can send and receive CAN messages on DIM.

### Testing Done

Hooked up a CAN dongle to manually send/receive messages.

### Checklist


boards/DIM/Src/main.c
```diff
@@ -68,14 +80,27 @@ static void MX_ADC2_Init(void);
...
 
 /* USER CODE BEGIN PFP */
 
 /* USER CODE END PFP */
 
 /* Private user code ---------------------------------------------------------*/
 /* USER CODE BEGIN 0 */
+static void CanRxQueueOverflowCallBack(size_t overflow_count)
+{
+    shared_assert(can_tx != NULL);
```
garethellis0, line 93: In keeping with the discussion we had recently about checking for NULL args, is this the sort of the place we need to do it? IMO it doesn't seem critical enough.

Perhaps add it to the auto-generated code for `CanTx` and `CanRx` so we don't have to worry about it?

boards/DIM/Src/main.c
```diff
@@ -437,11 +480,35 @@ void RunTaskCanTx(void const *argument)
...
+
+    /* Infinite loop */
+    for (;;)
+    {
+        Io_CanTx_EnqueuePeriodicMsgs(
+            can_tx, osKernelSysTick() * portTICK_PERIOD_MS);
+        osDelayUntil(&PreviousWakeTime, period_ms);
+    }
+    /* USER CODE END StartTask1kHz */
+}
```
garethellis0, line 510: Please correct me if I'm wrong, but generally we have dedicated tasks for can RX and TX, no? Any particular reason we don't here?

boards/DIM/Src/main.c
thekenu: Oh yeah I totally had the same thought while porting this code over. Let me clean this up for all the boards in this PR.

boards/DIM/Src/main.c
thekenu: We do! 

```
void        RunTask100Hz(void const *argument);
void        RunTaskCanRx(void const *argument);
void        RunTaskCanTx(void const *argument);
void        StartTask1kHz(void const *argument);
```

The task order is just different on this board haha. I can go edit the `.ioc` so the CanRx and CanTx tasks are at the bottom like the other projects.

thekenu: Unless you're asking about something else? :thinking:

garethellis0: Nope, sorry, somehow just totally missed them here.

PR #563, opened 2020-04-12
Implement assert_func and replace App_SharedAssert.h with assert.h

thekenu:
### Summary

I wrote `shared_assert` but I really hate it. TIL you can override `__assert_func` so we can finally write `assert` instead of `shared_assert`. I also took the chance to remove some clearly excessive `assert`.

The general rule of thumb I used for deciding which assert should be deleted is:

- `malloc` result always needs to be asserted
- Function arguments of pointer type is not checked **UNLESS** it's in `_Create` or `_Init` function (I don't think it would be wise if we were to create a `Imd` with  NULL method functions)

### Checklist


thekenu:
Definitely one of those moments when I look at the code and thought to myself, wtf did I write .........

PR #569, opened 2020-04-20
Update CAN period for heartbeats in DBC

thekenu:
### Summary

DIM-10, FSM-1, etc all specify heartbeaet to be 10Hz

### Checklist


boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -310,4 +310,4 @@ TEST_F(
             123, App_CanTx_GetPeriodicSignal_SECONDS_SINCE_POWER_ON(
                      can_tx_interface));
     }
-};
```
thekenu, line 313: nit fix not relevant to this PR but stumbled upon it

PR #570, opened 2020-04-21
Display state of charge on 7-segment displays

thekenu:
### Summary

Fulfill part of DIM-9 by reading SoC from CAN and displaying it on 7-seg

### Testing Done

- Wrote unit tests
- Manually tested on DIM by setting different SoC using `App_CanRx_BMS_STATE_OF_CHARGE_SetSignal_STATE_OF_CHARGE` 

### Checklist


boards/DIM/Test/Src/Test_Dim.cpp
```diff
@@ -1,3 +1,35 @@
 #include "Test_Dim.h"
 
 DEFINE_FFF_GLOBALS;
+
+void DimTest::SetUp()
```
thekenu, line 5: Take advantage of defining a common Setup() and Teardown() in the base class to reduce code bloat.

boards/DIM/Src/main.c
```diff
@@ -516,6 +517,8 @@ void RunTask100Hz(void const *argument)
     /* Infinite loop */
     for (;;)
     {
+        App_SharedStateMachine_Tick(state_machine);
+        Io_SevenSegDisplays_WriteCommands();
```
thekenu, line 521: Could possibly define this as a callback function that  gets called at the end of `App_SevenSegDisplays_SetHexDigits`. Haven't really decided if it's better that way though.

garethellis0: Yeah I think registering it at a callback makes more sense.

boards/DIM/Test/Src/Test_SevenSegDisplays.cpp
```diff
@@ -6,46 +6,8 @@ DEFINE_FAKE_VOID_FUNC(set_left_hex_digit, struct SevenSegHexDigit);
 
 class SevenSegDisplaysTest : public DimTest
 {
```
garethellis0, line 8: Hmm, I would think that this class would have all the seven seg setup and then the `Dim` test should extend this class. Otherwise this class gets everything in the dim test class, which doesn't really make sense with regards to making sure each class only gets what it needs and nothing more.

thekenu: whups i think you're totally right. I think this means StateMachineTeste will just inherit from multiple classes in the future?

```
class DimTest : public SevenSegDisplaysTest, public LedTest, public CanTest
```

I think i just forgot multiple inheritance was a thing ...

garethellis0: Correct, and lmao.

boards/DIM/Src/main.c
thekenu: Registered a callback function that gets invoked after we set a value to the 7-seg displays (e.g. `SetHexDigits` or `SetUnsignedBase10Value`)

scripts/travis_ci/travis_script.sh
```diff
@@ -25,19 +25,28 @@ if [ "$RUN_ARM_BUILD" = "true" ]; then
 fi
 
 if [ "$RUN_X86_TESTS" = "true" ]; then
-    BUILD_DIR=boards/x86_build
-    # Use Release build for testing because the heartbeat monitor relies on it
-    # for code logic
-    travis_run cmake -S boards -B $BUILD_DIR -DPLATFORM=x86 -DCMAKE_BUILD_TYPE=Release
-    travis_run make --directory=$BUILD_DIR
-    travis_run "cd $BUILD_DIR && ctest --verbose"
+    # The heartbeat monitor logic is different in Release and Debug build type,
```
thekenu, line 28: Not so relevant to this PR but since I was cleaning up the test files I thought I'd fix this too

thekenu:
@garethellis0 Ready for another review. I took some time to clean up the test files such that each file only gets what it needs.

boards/DIM/Inc/App/App_SevenSegDisplays.h
```diff
@@ -43,11 +46,24 @@ void App_SevenSegDisplays_Destroy(struct SevenSegDisplays *seven_seg_displays);
...
     const uint8_t                  hex_digits[],
     size_t                         num_hex_digits);
+
+/**
+ * Set an unsigned base-10 value on the given group of 7-segment displays
+ * @note The value that can be set is constrained by the number of 7-segment
+ *       displays. Use the return code to see if the requested value is valid.
+ * @param seven_seg_displays The group of 7-segment displays to set base-10
+ *                           value for
+ * @param value The unsigned base-10 value to set
```
garethellis0, line 64: nit; might be to better to say "to display" then "to set", since that's the effect the user of this is likely aiming for.

boards/DIM/Src/App/App_SevenSegDisplays.c
```diff
@@ -77,5 +81,40 @@ ErrorCode App_SevenSegDisplays_SetHexDigits(
             seven_seg_displays->displays[i], hex_digit);
     }
 
-    return ERROR_CODE_OK;
+    if (seven_seg_displays->set_value_callback != NULL)
+    {
+        seven_seg_displays->set_value_callback();
+    }
```
garethellis0, line 87: IMO we should just put a check for this in the constructor, I don't see any use case for this where we give it a `NULL` callback.

boards/DIM/Src/App/App_SevenSegDisplays.c
```diff
@@ -77,5 +81,40 @@ ErrorCode App_SevenSegDisplays_SetHexDigits(
...
+
+        if (value == 0)
+        {
+            App_SevenSegDisplays_SetHexDigits(
+                seven_seg_displays, digits, num_digits);
+            break;
+        }
+    }
+
+    return EXIT_CODE_OK;
```
garethellis0, line 119: Just to be clear, are we displaying all digits/error codes in base 10? Just want to make sure we're OK with giving up the extra values that displaying in hex gives us.

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Fine with me :v:

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Yeah so i was reviewing the specs on what exactly we need to display, and there are

- The SoC: This has to be in base 10 since it goes from 0 to 100
- The error codes follow the scheme [One digit for Board ID] [Two digits for Error Code], leaving 99 error codes per board which should be more than enough. I feel like we should change `App_SevenSegDisplays_SetHexDigits` to a static function because I don't see any use-case (arguably we could even just remove it but I'd leave it in for now).

boards/DIM/Inc/App/App_SevenSegDisplays.h
thekenu: But then should the function name be "DisplayUnsignedBase10"? I do agree that "to set" is a bit vague and weak though

thekenu: Changed only "to set" to "to display" for now

boards/DIM/Src/App/App_SevenSegDisplays.c
thekenu: Actually I thought it is strictly better to hide `SetHexDigits` so I turned it into a static function d7dc5c1af2e46993ccda13e0673bdac2190e7b10

thekenu: Uggh the existing tests for `SetHexDigits` can't compile anymore because well, it's a private function now. What should we do in this case @garethellis0 ?

garethellis0: Leave them public?

thekenu: Didn't want to leave them public if we're not using it but I guess it's more important to test the function

PR #572, opened 2020-04-21
Implement application code for regen paddle

thekenu:
### Checklist


boards/DIM/Inc/App/App_RegenPaddle.h
```diff
@@ -0,0 +1,20 @@
...
+ *                            pressed at all and 100% means pressed all the way
+ * @param lower_deadzone The paddle position threshold of lower deadzone
+ * @param upper_deadzone The paddle position threshold of upper deadzone
+ * @return The created regen paddle, whose ownership is given to the caller
+ */
+struct RegenPaddle *App_RegenPaddle_Create(
+    float (*get_paddle_position)(void),
+    float lower_deadzone,
+    float upper_deadzone);
+void  App_RegenPaddle_Destroy(struct RegenPaddle *regen_paddle);
```
thekenu, line 18: I actually am starting to fill like the application code is shaping up nicely and the boilerplate javadoc is becoming not necessary, especially for all the `Destroy`, `Tick`, and getter functions.

boards/DIM/Test/Src/Test_RegenPaddle.cpp
```diff
@@ -0,0 +1,63 @@
...
+        regen_paddle = NULL;
+    }
+    const float         DEFAULT_LOWER_DEADZONE = 5.0f;
+    const float         DEFAULT_UPPER_DEADZONE = 95.0f;
+    struct RegenPaddle *regen_paddle;
+};
+
+TEST_F(RegenPaddleTest, check_lower_deadzone)
+{
+    get_paddle_position_fake.return_val = std::nextafter(
```
thekenu, line 33: using `nextafter` as a consistent way to test "threshold value - epsilon", where epsilon is a small change in value.

thekenu:
@garethellis0 there's some overlapping changes to the struct of test files and `ErrorCode.h` but the rest of the changes are quite and ready for review :v:

boards/DIM/Src/App/App_RegenPaddle.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+ErrorCode App_RegenPaddle_Tick(struct RegenPaddle *const regen_paddle)
+{
+    const uint32_t paddle_position = regen_paddle->get_paddle_position();
+
+    if (paddle_position > 100)
+    {
+        return ERROR_CODE_OUT_OF_RANGE;
+    }
```
thekenu, line 42: Couldn't use invalid args error code because paddle position isn't an argument. Came up with a more generic out of range error code. I feel like this error code could be rephrased as something more concise so suggestions are welcome.

boards/DIM/Inc/App/App_RegenPaddle.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <stdint.h>
+#include "App_ErrorCode.h"
+
+/**
+ * Allocate and initialize a regen paddle
+ * @param get_paddle_position A function that can be called to get the paddle
+ *                            position as a percentage, where 0% means not
+ *                            pressed at all and 100% means pressed all the way
```
garethellis0, line 10: could we change `0%` and `100%` to just `0` and `100` here? IMO it's potentially ambiguous what number `100%` maps to here.

boards/DIM/Inc/App/App_RegenPaddle.h
```diff
@@ -0,0 +1,22 @@
...
+ * @param lower_deadzone The paddle position threshold of lower deadzone
+ * @param upper_deadzone The paddle position threshold of upper deadzone
```
garethellis0, line 12: This is a bit ambiguous. Do we ignore values above this? Is this value mapped to 100%? Same goes for the lower deadzone.

boards/DIM/Src/App/App_RegenPaddle.c
```diff
@@ -0,0 +1,74 @@
...
+    else
+    {
+        regen_paddle->regen =
+            100 *
+            (regen_paddle->paddle_position - regen_paddle->lower_deadzone) /
+            (regen_paddle->upper_deadzone - regen_paddle->lower_deadzone);
+    }
+
+    return ERROR_CODE_OK;
+}
```
garethellis0, line 63: Why is this a `tick` function? Why not `ErrorCode App_RegenPaddle_GetPaddlePosition(struct RegenPaddle* paddle, unint32_t* returned_paddle_position)`?

boards/DIM/Src/App/App_RegenPaddle.c
```diff
@@ -0,0 +1,74 @@
...
+uint32_t App_RegenPaddle_GetRegen(const struct RegenPaddle *const regen_paddle)
+{
+    return regen_paddle->regen;
+}
+
+uint32_t
+    App_RegenPaddle_GetPosition(const struct RegenPaddle *const regen_paddle)
+{
+    return regen_paddle->paddle_position;
+}
```
garethellis0, line 74: IMO there should only be `GetRegen` and `GetPosition`, no `tick`, because ticking is really an implementation detail we don't care about here.

boards/DIM/Inc/App/App_RegenPaddle.h
thekenu: Ah yeah that's true, fixed

boards/DIM/Src/App/App_RegenPaddle.c
thekenu: Hmm maybe I am misunderstanding what `Tick` is but I think of it as a step to update the internal states (e.g. `paddle_position` and `regen`) using the underlying IO functions (e.g. `get_pedal_position`).

Whereas `GetPosition` and `GetRegen` just returns the value the internal state, which is essentially the calculations from the most recent update step. Does that line of reasoning seem sane to you?

boards/DIM/Src/App/App_RegenPaddle.c
thekenu: Maybe I misunderstood when Tick function should be used:
https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/572#discussion_r414047203

boards/DIM/Inc/App/App_RegenPaddle.h
thekenu: Good point :+1: I rephrased it so let me know if it's still ambiguous. Also clarified that paddle position not in the range of 0 - 100 is invalid.

boards/DIM/Src/App/App_RegenPaddle.c
garethellis0: I mean, I understand it, but IMO we should just have `GetPosition` and `GetRegen` call the io functions directly and not have a `tick` function, because the abstraction around the padel doesn't actually have anything to do with ticking, that's a hardware-specific detail.

thekenu: To clarify, this would mean removing these internal variables right? Because we can just call the underlying io function each time to get the value.
```
uint32_t paddle_position;
uint32_t regen;
```
Also what kind of application would require a `tick` function then? Just for my future reference.

garethellis0: >To clarify, this would mean removing these internal variables right? Because we can just call the underlying io function each time to get the value.

Yep.

> Also what kind of application would require a tick function then? Just for my future reference.

Hmm, I suppose anything that needs to run periodically. IMO this isn't periodic because we ask it when we want information. In general I avoid tick functions unless needed because they explicitly rely on side effects (generally) which makes them much harder to reason about and test.

thekenu: I guess getter functions that request an underlying hardware driver (e.g. ADC for regen paddle) for information should probably never have a tick function? I may have told @pasitlaothamatas to use a Tick for flow meter (which reads PWM frequency). I guess we should remove that `Tick` too? 

https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/master/boards/FSM/Src/App/App_FlowMeter.c

boards/DIM/Test/Inc/Test_SevenSegDisplays.h
```diff
@@ -13,7 +13,7 @@ DECLARE_FAKE_VOID_FUNC(set_middle_hex_digit, struct SevenSegHexDigit);
 DECLARE_FAKE_VOID_FUNC(set_left_hex_digit, struct SevenSegHexDigit);
 DECLARE_FAKE_VOID_FUNC(display_value_callback);
 
-class SevenSegDisplaysTest : public testing::Test
+class SevenSegDisplaysTest : public virtual testing::Test
```
thekenu, line 16: @garethellis0 Ran into the diamond problem where `class SevenSegDisplaysTest ` and `class RegenPaddleTest` inherit from the same base class `testing::Test`, then `class StateMachineTest` uses multiple inheritance. I did some quick research but honestly I don't know what's the right thing to do here: redesign the inheritance or just slap on a virtual inheritance?

boards/DIM/Src/App/App_RegenPaddle.c
garethellis0: Yeah, in general it makes things easier to misuse, because now all the users of whatever code you built need to understand that you must first call `tick` before calling any of the getter functions in order to get new information. It's not much, but it is mental overhead that makes it easier to create bugs.

thekenu: Awesome, I'll fix up flow meter later then :+1:

boards/DIM/Test/Inc/Test_SevenSegDisplays.h
garethellis0: virtual inheritance seems fine to me here, this is one of the textbook cases for it.

PR #573, opened 2020-04-23
Rename mapped regen paddle position to regen request

thekenu:
### Checklist


specs.md
```diff
@@ -233,6 +233,6 @@ DIM-3 | Drive mode switch | The DIM must transmit the drive mode position of the
 DIM-4 |  Start, traction control, torque vectoring switches | For each of the switches, the DIM must: <br/> - Transmit the on/off switch status of over CAN at 1kHz. <br/> - Set the corresponding green status LEDs when the switch is on.
 DIM-5 | IMD LED | The DIM must turn on the IMD LED when it receives IMD fault status from BMS over CAN. | EV.8.5.5
 DIM-6 | BSPD LED | The DIM must turn on the BSPD LED when it receives BSPD fault status from FSM over CAN.
-DIM-7 | Mapped regen paddle percentage reporting | The DIM must report the mapped regen paddle percentage over CAN at 1kHz.
```
thekenu, line 236: @garethellis0 I found the term "mapped regen paddle percentage" slightly awkward and wanted to rename it. Since it's describing the mapping between regen paddle and the amount of desired regen, I thought it's more clear to call it "regen request". However, @kirons96 believes it could be misinterpreted as a command directly sent to the inverter because it sounds similar to "torque request". We couldn't reach a consensus because I think the specifications already define "regen request" quite clearly.

What's your take on whether this re-phrasing is necessary ?:octocat:

garethellis0: I agree with Kiron. Specs should be verbose and specific, there's no real need to have them be eloquent (though of course if you can without sacrificing their principle purpose, all the better), and IMO this is more ambiguous.

thekenu: Kk let's close this PR then, must have just been me who found mapped regen paddle confusing  :+1:

PR #574, opened 2020-04-24
Refactor voltage monitor into a generic InRangeCheck to be used for voltage and current

thekenu:
### Summary

Range check for
- Voltages: vbat, 24v acc, 24v aux
- Currents: aux1, aux2, left inv, right inv, energy meter, can glv, air shutdown

### Checklist


boards/BMS/Inc/App/App_BmsWorld.h
```diff
@@ -8,11 +8,8 @@
 struct BmsWorld;
 
 /**
- * Allocate and initialize a world in the application layer
- * @param can_tx_interface The CAN TX interface to register to the world
- * @param can_rx_interface The CAN RX interface to register to the world
- * @param imd The IMD to register to the world
- * @param heartbeat_monitor The heartbeat monitor to register to the world
+ * Allocate and initialize a world in the application layer by registering
```
thekenu, line 11: The javaco for world `World_Create` is getting a bit out of hand since I added like 8 new ones for PDM. Decided to just remove the `@param` altogether.

boards/PDM/Inc/App/App_InRangeCheck.h
```diff
@@ -0,0 +1,45 @@
...
+ * the given min/max range
+ * @get_voltage A function that can be called to get the value
+ * @min_value Minimum value in the range, inclusive
+ * @max_value Maximum value in the range, inclusive
+ * @error_callback The function to call when the value is out of the specified
+ *                 range
+ * @return A pointer to the created in-range check, whose ownership is given
+ * to the caller
+ */
+struct InRangeCheck *App_InRangeCheck_Create(
```
thekenu, line 23: Replaced voltage monitor with a more generic `InRangeCheck` that can be used for current too

boards/PDM/Inc/App/App_InRangeCheck.h
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Get the value for the given in-range check
+ * @param in_range_check The in-range check to get value for
+ * @param value_buffer The buffer to store the value in
+ * @return VALUE_IN_RANGE if the value was in the specified range
+ *         VALUE_UNDERFLOW if the value was below the specified range
+ *         VALUE_OVERFLOW if the value was above the specified range
+ */
+enum InRangeCheck_Status App_InRangeCheck_GetValue(
```
thekenu, line 43: @garethellis0 I am guessing you would prefer a no `Tick` design? Previously I had `Tick` in this

boards/PDM/Test/Src/Test_StateMachine.cpp
```diff
@@ -176,3 +240,9 @@ TEST_F(PdmStateMachineTest, check_air_closed_state_is_broadcasted_over_can)
         CANMSGS_PDM_STATE_MACHINE_STATE_AIR_CLOSED_CHOICE,
         App_CanTx_GetPeriodicSignal_STATE(can_tx_interface));
 }
+
+TEST_F(
+    PdmStateMachineTest,
+    check_vbat_voltage_is_broadcasted_over_can_in_all_states)
+{
```
thekenu, line 247: Plenty of unit tests for  stting `can_tx` signals, to be added later once everything else looks good

boards/PDM/Src/App/states/App_AirOpenState.c
```diff
@@ -12,7 +12,63 @@ static void AirOpenStateRunOnEntry(struct StateMachine *const state_machine)
 
 static void AirOpenStateRunOnTick(struct StateMachine *const state_machine)
 {
-    UNUSED(state_machine);
+    struct PdmWorld *world = App_SharedStateMachine_GetWorld(state_machine);
+
```
garethellis0, line 16: I might be missing something here, but it seems like there is a significant amount of duplication between the various states here? Could we we merge things into shared function(s)?

boards/PDM/Src/Io/Io_CurrentCheck.c
```diff
@@ -0,0 +1,139 @@
...
+    enum InRangeCheck_Status status)
+{
+    if (status == VALUE_UNDERFLOW)
+    {
+        App_CanTx_SetPeriodicSignal_UNDERCURRENT_AIR_SHUTDOWN(_can_tx, true);
+    }
+    else if (status == VALUE_OVERFLOW)
+    {
+        App_CanTx_SetPeriodicSignal_OVERCURRENT_AIR_SHUTDOWN(_can_tx, true);
+    }
```
garethellis0, line 138: bit of a nit, but generally I prefer `switch` statements instead of `if` statements for enums because you can do a compile-time check that all cases of a given enum are being properly handled (as long as you don't have a `default` case).

garethellis0: This really is a nit though, I don't feel that strongly about this, up to you if you want to change it.

boards/PDM/Src/Io/Io_VoltageCheck.c
```diff
@@ -0,0 +1,63 @@
...
+{
+    if (status == VALUE_UNDERFLOW)
+    {
+        App_CanTx_SetPeriodicSignal_UNDERVOLTAGE_24_V_ACC(_can_tx, true);
+    }
+    else if (status == VALUE_OVERFLOW)
+    {
+        App_CanTx_SetPeriodicSignal_OVERVOLTAGE_24_V_ACC(_can_tx, true);
+    }
+}
```
garethellis0, line 63: Another thought I had here, it doesn't seem like it would be too hard to make a quick macro to reduce the amount of duplication in this file. Same goes for the current checks. Something like:
```
#define CHOOSE_A_BETTER_MACRO_NAME(NAME1, NAME2) \
void Io_VoltageInRangeCheck_##NAME1##ErrorCallback(enum InRangeCheck_Status status) \
{ \
    if (status == VALUE_UNDERFLOW) \
    { \
        App_CanTx_SetPeriodicSignal_UNDERVOLTAGE_##NAME2(_can_tx, true); \
    } \
    else if (status == VALUE_OVERFLOW) \
    { \
        App_CanTx_SetPeriodicSignal_OVERVOLTAGE_##NAME2(_can_tx, true); \
    } \
}
```
It's not pretty, but  pretty effective for reducing duplication here. 

Anyhow, totally your call.

boards/PDM/Inc/App/App_InRangeCheck.h
garethellis0: yep.

boards/PDM/Src/Io/Io_VoltageCheck.c
thekenu: Oh yeah I would prefer this. I was getting tired of copying-and-pasting for sure.

thekenu: Went for a static function instead :v:

boards/PDM/Src/App/App_SetPeriodicCanSignals.c
```diff
@@ -0,0 +1,17 @@
+#include "App_SetPeriodicCanSignals.h"
+#include "App_CanTx.h"
+#include "App_InRangeCheck.h"
+#include "App_ErrorCode.h"
+
+void App_SetPeriodicCanSignals_InRangeCheck(
```
thekenu, line 6: I started with one function for each inRangeCheck
```
App_SetPeriodicCanSignals_VbatInRangeCheck
App_SetPeriodicCanSignals_24vAuxInRangeCheck
...
```

But it seemed a bit redundant still so I just created one function for all the `struct InRangeCheck`

boards/PDM/Src/App/states/App_AirOpenState.c
thekenu: Wasn't sure how aggressive we should get with the clean-up, but here's what I ended up with for now: https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/574/files#r415273335

garethellis0: It looks like we're still duplicating all the "InRange" checks between the states?

boards/PDM/Src/Io/Io_VoltageCheck.c
garethellis0: Nice, that's better then what I suggested.

boards/PDM/Src/App/states/App_AirOpenState.c
thekenu: Ah I think I may misunderstood what you were looking for. Were  you thinking something more along the lines of:

```
// App_AllStates.c
void AllStatesRunOnTick(state_machine)
{
    struct PdmWorld *world = App_SharedStateMachine_GetWorld(state_machine);
    struct PdmCanTxInterface *can_tx = App_PdmWorld_GetCanTx(world);
    struct InRangeCheck *     vbat_voltage_check =
        App_PdmWorld_GetVbatInRangeCheck(world);
    struct InRangeCheck *_24v_aux_voltage_check =
        App_PdmWorld_Get24vAuxInRangeCheck(world);
    struct InRangeCheck *_24v_acc_voltage_check =
        App_PdmWorld_Get24vAccInRangeCheck(world);
    struct InRangeCheck *aux1_current_check =
        App_PdmWorld_GetAux1CurrentCheck(world);
    struct InRangeCheck *aux2_current_check =
        App_PdmWorld_GetAux2CurrentCheck(world);
    struct InRangeCheck *left_inverter_current_check =
        App_PdmWorld_GetLeftInverterCurrentCheck(world);
    struct InRangeCheck *right_inverter_current_check =
        App_PdmWorld_GetRightInverterCurrentCheck(world);
    struct InRangeCheck *energy_meter_current_check =
        App_PdmWorld_GetEnergyMeterCurrentCheck(world);
    struct InRangeCheck *can_current_check =
        App_PdmWorld_GetCanCurrentCheck(world);
    struct InRangeCheck *air_shutdown_current_check =
        App_PdmWorld_GetAirShutdownCurrentCheck(world);

    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, vbat_voltage_check, App_CanTx_SetPeriodicSignal_VBAT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, _24v_aux_voltage_check, App_CanTx_SetPeriodicSignal__24_V_AUX);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, _24v_acc_voltage_check, App_CanTx_SetPeriodicSignal__24_V_ACC);

    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, aux1_current_check,
        App_CanTx_SetPeriodicSignal_AUXILIARY1_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, aux2_current_check,
        App_CanTx_SetPeriodicSignal_AUXILIARY2_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, left_inverter_current_check,
        App_CanTx_SetPeriodicSignal_LEFT_INVERTER_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, right_inverter_current_check,
        App_CanTx_SetPeriodicSignal_RIGHT_INVERTER_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, energy_meter_current_check,
        App_CanTx_SetPeriodicSignal_ENERGY_METER_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, can_current_check, App_CanTx_SetPeriodicSignal_GLV_CURRENT);
    App_SetPeriodicCanSignals_InRangeCheck(
        can_tx, air_shutdown_current_check,
        App_CanTx_SetPeriodicSignal_AIR_SHUTDOWN_CURRENT);
}
// App_AirOpenState.c
static void AirOpenStateRunOnTick(struct StateMachine *const state_machine) {
  AllStatesRunOnTick(state_machine);
 ...
}

// App_AirClosedState.c
 static void AirClosedStateRunOnTick(struct StateMachine *const state_machine) {
  AllStatesRunOnTick(state_machine);
 ...
}
```

boards/PDM/Src/Io/Io_CurrentCheck.c
thekenu: Turned into switch statements !

boards/PDM/Src/App/states/App_AirOpenState.c
garethellis0: Yeah, basically. If we're doing the exact same thing in multiple states then we might as well put them in a function to reduce duplication. Just wasn't sure if I was missing something here.

boards/PDM/Src/App/App_InRangeCheck.c
```diff
@@ -0,0 +1,58 @@
...
+
+    return in_range_check;
+}
+
+void App_InRangeCheck_Destroy(struct InRangeCheck *const in_range_check)
+{
+    free(in_range_check);
+}
+
+enum InRangeCheck_Status App_InRangeCheck_GetValue(
```
thekenu, line 35: I ended up removing the callback functions for `InRangeCheck`  because it was quite hard to test (maybe that's a sign it wasn't a good design?). The problem was that the callback functions weren't part of the state machine logic and thus couldn't be tested.

boards/PDM/Inc/App/App_CurrentLimits.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#define AUX1_MIN_CURRENT 0.0f
```
thekenu, line 3: I needed an elegant way to share these constants between embedded and gtest targets so i didn't have to hard code the same value twice. This was the simplest way I could think of.

boards/PDM/Test/Src/Test_StateMachine.cpp
```diff
@@ -189,3 +366,30 @@ TEST_F(PdmStateMachineTest, check_air_closed_state_is_broadcasted_over_can)
         CANMSGS_PDM_STATE_MACHINE_STATE_AIR_CLOSED_CHOICE,
         App_CanTx_GetPeriodicSignal_STATE(can_tx_interface));
 }
+
+TEST_F(PdmStateMachineTest, check_vbat_voltage_can_signals_in_all_states)
```
thekenu, line 370: So the idea here is  to have 
- 1 test for each voltage (because the bheaviour is the same in every state)
- 2 tests for each current (because the behaviour in INIT state is different than other states)

scripts/codegen/CAN/cantx_codegen.py
```diff
@@ -62,18 +62,33 @@ def __init_functions(self, function_prefix):
...
+        lst = []
+
+        for signal in self._periodic_cantx_signals:
+            if signal.is_float:
+                lst.append(Function(
+                    'void %s_SetPeriodicSignal_%s(struct %sCanTxInterface* can_tx_interface, %s value)' % (
+                    function_prefix, signal.uppercase_name, self._sender.capitalize(), signal.type_name),
+                    '',
+                    '''\
+    if (App_CanMsgs_{msg_snakecase_name}_{signal_snakecase_name}_is_in_range(value) || isnanf(value))
```
thekenu, line 74: Previously our setter functions didn't accept NAN for floating pointer values

boards/PDM/Test/Src/Test_StateMachine.cpp
```diff
@@ -61,25 +60,42 @@ class PdmStateMachineTest : public testing::Test
...
-        vbat_voltage_monitor = App_VoltageMonitor_Create(
-            GetVbatVoltage, GetVbatMinVoltage, GetVbatMaxVoltage,
-            VbatErrorCallback);
-        _24v_aux_voltage_monitor = App_VoltageMonitor_Create(
-            Get24vAccVoltage, Get24vAccMinVoltage, Get24vAccMaxVoltage,
-            _24vAuxErrorCallback);
-        _24v_acc_voltage_monitor = App_VoltageMonitor_Create(
-            Get24vAccVoltage, Get24vAccMinVoltage, Get24vAccMaxVoltage,
-            _24vAccErrorCallback);
+        can_rx_interface   = App_CanRx_Create();
```
thekenu, line 63: Would be nice if we don't have to write all the `_Create` twice, once in `main` and once in `Setup`. What would be a nice way to do this @garethellis0 ?

garethellis0: Sorry, not seeing where you're setting this up twice? `Setup` will run before each test, so that should be sufficient?

thekenu: Oh sorry when I said `main` I mean the main function in the ARM embedded target.
```
// main.c (ARM)
int main(void) {
  ..
  x = App_XXX_Create();
  y = App_YYY_Create();
  App_PdmWorld_Create(x, y);
  ..
}
```

```
// Test_StateMachine.cpp
void Setup override {
  ..
  x = App_XXX_Create();
  y = App_YYY_Create();
  App_PdmWorld_Create(x, y);
  ..
}
```
We're writing the same code for the embedded and  test targets

garethellis0: Oh, I see. That's a bit of a tricky one, because we're not doing _exactly_ the same thing, in that we're passing in different IO functions in the tests versus in the arm target. I can't think of a neat solution here that wouldn't obfuscate things further.

boards/PDM/Src/App/App_SetPeriodicCanSignals.c
```diff
@@ -0,0 +1,116 @@
...
+        break;
+        case VALUE_OVERFLOW:
+        {
+            overflow_setter(can_tx, true);
+        }
+        break;
+    }
+
+    can_signal_setter(can_tx, value);
+}
```
garethellis0, line 36: As a thought, this whole function could be a macro `APP_CREATE_PERIODIC_CAN_SIGNALS_IN_RANGE_CHECK(CAN_INTERFACE_TYPE)` so that we can share it between boards.

thekenu: Oh yeah let's do that before I merge so @pasitlaothamatas can use it for his FSM wheel speed PR as well. How do we want to design the macro to paste the `BoardName` in to `CanTxInterface` though? Or we could even just leave it as a void pointer so we can leave it as a function instead of a macro?

garethellis0: I am fairly strongly against tossing away typing with void pointers in basically all cases. I think pasting the `BoardName` is fine, ie:
```
APP_CREATE_PERIODIC_CAN_SIGNALS_IN_RANGE_CHECK(CAN_INTERFACE_TYPE) \
static void App_SetPeriodicCanSignals_InRangeCheck( \
    struct CAN_INTERFACE_TYPE *const  can_tx, \
    const struct InRangeCheck *const in_range_check, \
    void (*const can_signal_setter)(struct CAN_INTERFACE_TYPE *, float), \
    void (*const underflow_setter)(struct CAN_INTERFACE_TYPE *, uint8_t), \
    void (*const overflow_setter)(struct CAN_INTERFACE_TYPE *, uint8_t)) \
{ \
    float                    value; \
    enum InRangeCheck_Status status = \
        App_InRangeCheck_GetValue(in_range_check, &value); \
 \
    switch (status) \
    { \
        case VALUE_IN_RANGE: \
        { \
            underflow_setter(can_tx, false); \
            overflow_setter(can_tx, false); \
        } \
        break; \
        case VALUE_UNDERFLOW: \
        { \
            underflow_setter(can_tx, true); \
        } \
        break; \
        case VALUE_OVERFLOW: \
        { \
            overflow_setter(can_tx, true); \
        } \
        break; \
    } \
\
    can_signal_setter(can_tx, value); \
}

```

PR #578, opened 2020-04-27
Ken/add faults

thekenu:
### Checklist


scripts/codegen/CAN/canrx_codegen.py
```diff
@@ -232,9 +266,15 @@ def __init_functions(self, function_prefix):
                     msg_uppercase_name=msg.snake_name.upper())
 
             signal_setters_fmt = '''
-            App_CanRx_{msg_uppercase_name}_SetSignal_{signal_name_uppercase}(
+            App_CanRx_{msg_uppercase_name}_SetPeriodicSignal_{signal_name_uppercase}(
```
thekenu, line 269: To make the `_add_critical_fault` cleaner. I split `CanRx_SetSignal` up to become  `CanRx_SetPeriodicSignal` and `CanRx_SetNonPeriodicSignal` to it matches `CanTx_SetPeriodicSignal` better. That's the reason for all the changes in this script.

boards/BMS/Src/App/App_FaultManager.c
```diff
@@ -0,0 +1,81 @@
...
+    _fsm_can = fsm_can;
+    _dcm_can = dcm_can;
+    _pdm_can = pdm_can;
+    _dim_can = dim_can;
+    _bms_can = bms_can;
+}
+
+bool App_FaultManager_HasFsmCriticalFaultManager(void)
+{
+    void *can = _fsm_can;
```
thekenu, line 33: Could add `_fsm_can` as another argument to `add_critical_fault` but I thought it makes the macro more clunky to use

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -211,6 +219,7 @@ BA_ "GenMsgCycleTime" BO_ 413 10;
 BA_ "GenMsgCycleTime" BO_ 500 100;
 BA_ "GenMsgCycleTime" BO_ 501 5000;
 BA_ "GenMsgCycleTime" BO_ 503 10;
+BA_ "GenMsgCycleTime" BO_ 505 1000;
```
thekenu, line 222: Forgot to add period to DIM_ERRORS earlier and it was causing compilation issues

boards/shared/Src/App/App_FaultManager.c
```diff
@@ -0,0 +1,81 @@
...
+    if (PREFIX##_GetPeriodicSignal_##FAULT(can)) \
+        return true;
+
+#define add_critical_fault(PREFIX, FAULT) _add_critical_fault(PREFIX, FAULT)
+
+static void *_fsm_can;
+static void *_dcm_can;
+static void *_pdm_can;
+static void *_dim_can;
+static void *_bms_can;
```
thekenu, line 15: Could leave the user to define each pointer type but I feel like that's a bit much. I think I am okay with sacrificing type-checking for ease-of-use so I went with void pointers.

boards/shared/Inc/App/App_FaultManager.h
```diff
@@ -0,0 +1,37 @@
...
+#error "Please define PDM_CAN"
+#endif
+
+#ifndef DIM_CAN
+#error "Please define DIM_CAN"
+#endif
+
+#ifndef BMS_CAN
+#error "Please define BMS_CAN"
+#endif
```
thekenu, line 24: Could combine this into 

```
#if !defined(FSM_CAN) || !defined(DCM_CAN) || !defined(PDM_CAN) || !defined(DIM_CAN) || !defined(BMS_CAN)
#error "Some message"
#endif
```

Bit of a nit, but yeah.

boards/shared/Src/App/App_FaultManager.c
```diff
@@ -0,0 +1,81 @@
...
+    _fsm_can = fsm_can;
+    _dcm_can = dcm_can;
+    _pdm_can = pdm_can;
+    _dim_can = dim_can;
+    _bms_can = bms_can;
+}
+
+bool App_FaultManager_HasFsmCriticalFaultManager(void)
+{
+    void *can = _fsm_can;
```
thekenu, line 33: Could add `_fsm_can` as another argument to add_critical_fault (which means we could remove `void* can`) but I decided against it (for now) because it makes the macro more clunky to use

boards/BMS/Inc/App/App_FaultManagerConfig.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#define FSM_CAN App_CanRx_FSM_ERRORS
+#define DCM_CAN App_CanRx_DCM_ERRORS
+#define PDM_CAN App_CanRx_PDM_ERRORS
+#define DIM_CAN App_CanRx_DIM_ERRORS
+#define BMS_CAN App_CanTx
```
garethellis0, line 7: Why is the `BMS_CAN` different from all the others here?

boards/shared/Src/App/App_FaultManager.c
```diff
@@ -0,0 +1,81 @@
...
+}
+
+bool App_FaultManager_HasBmsCriticalFaultManager(void)
+{
+    void *can = _bms_can;
+    add_critical_fault(BMS_CAN, CRITICAL_FAULT);
+    return false;
+}
+
+bool App_FaultManager_HasCriticalFaultManager(void)
```
garethellis0, line 67: Given that these functions are suffixed like `HasCriticalFaultManager`, I'm not sure what they're returning?

thekenu:
@garethellis0 Yep let's hold this off until you are free then! I'll just leave this PR open as-is for now :penguin:

boards/BMS/Inc/App/App_FaultManagerConfig.h
thekenu: It's because BMS_ERRORS is TX from the point of view of BMS board.

If I were writing the config header for DCM, it would look like

```
#define FSM_CAN App_CanRx_FSM_ERRORS
#define DCM_CAN App_CanTx
#define PDM_CAN App_CanRx_PDM_ERRORS
#define DIM_CAN App_CanRx_DIM_ERRORS
#define BMS_CAN App_CanRx_BMS_ERRORS
```

boards/shared/Src/App/App_FaultManager.c
thekenu: It was a find-and-replace mistake :facepalm: I fixed it now. The correct function sigantures should be

```
bool App_FaultManager_HasFsmCriticalFault(void);
bool App_FaultManager_HasDcmCriticalFault(void);
bool App_FaultManager_HasPdmCriticalFault(void);
bool App_FaultManager_HasDimCriticalFault(void);
bool App_FaultManager_HasBmsCriticalFault(void);
bool App_FaultManager_HasCriticalFault(void);
```

garethellis0:
Moving design discussion from slack to here. We should:
- Change CAN generation so each board also receives it's own errors
- This allows us to write 
```
#define FSM_CAN App_CanRx_FSM_ERRORS
#define DCM_CAN App_CanRx_DCM_ERRORS
#define PDM_CAN App_CanRx_PDM_ERRORS
#define DIM_CAN App_CanRx_DIM_ERRORS
#define BMS_CAN App_CanRx_BMS_ERRORS
```
*once* in the `shared` code.
- We should *not* use `void*`, but instead should typedef the can interface, ie. `#typedef CanTx BMSCanTx` in each specific board

garethellis0:
Also we really should rename things like `FSM_CAN` to something like `CAN_FSM_ERRORS`

garethellis0:
whoops, didn't mean to close`

thekenu:
@garethellis0 So it looks like the CAN protocol doesn't allow a node to receive its own message because a message can only be ACK'd by a different node (Read more [here](https://electronics.stackexchange.com/questions/224022/can-bus-test-via-loopback-possible-or-not/224469)). Any alternative design to work around this limitation?

thekenu:
Closing this in favor of #615

PR #581, opened 2020-04-28
Add app and io code for drive mode rotary switch

thekenu:
### Summary

Add app and io code for drive mode rotary switch 

### Testing Done

For the io code, I was able to take some jumper wires and short each drive mode pin then observe that the correct DRIVE MODE is reported on the CAN traffic.

### Checklist


boards/DIM/Inc/App/App_RegenPaddle.h
```diff
@@ -27,6 +27,7 @@ void App_RegenPaddle_Destroy(struct RegenPaddle *regen_paddle);
  * Get the raw paddle position for the given regen paddle, where 0 means not
  * pressed at all and 100 means pressed all the way
  * @param regen_paddle The regen paddle to get raw paddle position for
+ * @param returned_raw_paddle_position Buffer to hold the raw paddle position
```
thekenu, line 30: Unrelated to drive mode switch but noticed that I missed some `@param`

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -176,6 +176,9 @@ BO_ 505 DIM_REGEN_PADDLE: 8 DIM
 SG_ Raw_Paddle_Position : 0|32@1+ (1,0) [0|100] "%" LOGGER
 SG_ Mapped_Paddle_Position : 32|32@1+ (1,0) [0|100] "%" LOGGER
 
+BO_ 506 DIM_ROTARY_SWITCH: 1 DIM
+SG_ Drive_Mode : 0|3@1+ (1,0) [0|0] "" DCM
```
pasitlaothamatas, line 180: Is the range supposed to be specified here?

boards/DIM/Inc/App/App_RegenPaddle.h
```diff
@@ -38,6 +39,8 @@ ExitCode App_RegenPaddle_GetRawPaddlePosition(
  * Get the mapped paddle position for the given regen paddle, where 0 means no
  * regen and 100 means maximum regen
  * @param regen_paddle The regen paddle to get mapped paddle position for
+ * @param returned_mapped_paddle_position Buffer to hold the mapped paddle
+ *                                        position
```
garethellis0, line 43: When I think of a `buffer` I generally think of 1 or more values, but this just looks like a single value? Perhaps `This will be set to the mapped paddle position` might be a bit more explicit?

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -2,6 +2,63 @@
...
+        case 5:
+        {
+            App_CanTx_SetPeriodicSignal_DRIVE_MODE(
+                can_tx,
+                CANMSGS_DIM_ROTARY_SWITCH_DRIVE_MODE_DRIVE_MODE_INVALID_CHOICE);
+        }
+        break;
+        default:
+        {
+            // Should never reach here
```
garethellis0, line 57: should we set it to `CANMSGS_DIM_ROTARY_SWITCH_DRIVE_MODE_DRIVE_MODE_INVALID_CHOICE` here as well?

scripts/codegen/CAN/App_CanMsgs.dbc
thekenu: Yep fixed :+1:

boards/DIM/Inc/App/App_RegenPaddle.h
thekenu: Fine with me :+1:

boards/DIM/Src/App/states/App_DriveState.c
thekenu: Hmm I think if we were to do that I'd add another enum. `case 5` is still a valid switch position that we can physically switch to (It's "invalid" only because `specs.md` says we only need 5 drive modes, not 6), but `case 6` and above would not be physically possible is likely a result of a programming or hardware error.

thekenu:
Addressed all the comments and ready for another round of review

boards/DIM/Src/App/states/App_DriveState.c
garethellis0: OK, I'm find with it as-is then.

thekenu: 👍 I'll also leave in a comment about `case 5` to further clarify what `INVALID` means

PR #587, opened 2020-05-03
Clean up CAN libraries

thekenu:
### Summary

- Added const pointer as needed
- Cleaned up sharedCan lib slightly

### Checklist


boards/shared/Src/Io/Io_SharedCan.c
```diff
@@ -1,27 +1,14 @@
-#include "App_CanTx.h"
+#include <assert.h>
+
 #include "Io_CanRx.h"
```
thekenu, line 3: Got rid of the nonsense headers here

PR #588, opened 2020-05-03
Replaced shared logging with printf

thekenu:
### Summary

The sbrk provided by STM32 is actually known to be broken for newlib and is documented in various forum posts. Someone provided a solution for this: http://www.nadler.com/embedded/newlibAndFreeRTOS.html because unfortunately STM32 isn't prioritizing providing a fix.

With this fix, we can actually use printf as opposed to using SharedLogging.

### Testing Done

- `printf` before scheduler starts, inside tasks. 
- Make sure that `assert` prints out its message appropriately as well

### Checklist


boards/CMakeLists.txt
```diff
@@ -186,6 +186,7 @@ function(create_arm_binary
             -Wl,-Map=${PROJECT_BINARY_DIR}/${BOARD_NAME}.map
             -Wl,-gc-sections,--print-memory-usage
             -Wl,-T ${ARM_LINKER_SCRIPT}
+            --specs=nano.specs
```
thekenu, line 189: Required when compiling newlib with configUSE_NEWLIB_REENTRANT set to 1

boards/BMS/Src/syscalls.c
```diff
@@ -113,7 +113,7 @@ __attribute__((weak)) int _write(int file, char *ptr, int len)
     return len;
 }
 
-caddr_t _sbrk(int incr)
+__attribute__((weak)) caddr_t _sbrk(int incr)
```
thekenu, line 116: The default implementation provided by STM32 is broken for multithreaded systems. Leave it as weak so it can be overridden by the patch.

boards/DIM/Inc/FreeRTOSConfig.h
```diff
@@ -174,8 +174,8 @@ standard names. */
 #define xPortSysTickHandler SysTick_Handler
 
 /* USER CODE BEGIN Defines */
-/* Section where parameter definitions can be added (for instance, to override
- * default ones in FreeRTOS.h) */
+// Required for thread-safety of newlib sprintf, strtok, etc...
```
markomarkusi, line 177:
```suggestion
// Integrates the Tracealyzer recorder with FreeRTOS. Tracealyzer recommends
// that this is inserted at the end of FreeRTOSConfig.h.
#if (configUSE_TRACE_FACILITY == 1)
#include "trcRecorder.h"
#endif

// Required for thread-safety of newlib sprintf, strtok, etc...
```

markomarkusi: `#define configUSE_TRACE_FACILITY 1`
Isn't defined in this file either

thekenu: Yeah i'm gonna do that in another PR

thekenu: And fully test the tracing before i make the changes

PR #591, opened 2020-05-06
Clean up GoogleTest files 

thekenu:
### Summary

- Removed multiple inheritance in `Test_StateMachine` as a means to share `Setup()` and `TearDown()` because the design falls apart when we try to instantiate more than one object from a class (e.g. When we instantiate two flow meters from `App_FlowMeter_Creat` in `Test_StateMachine`)
- Use public static function to share helper functions between classes (For an example, see: `static void SetImdCondition(...)`
- Stop hard coding values in `Test_StateMachine`. Instead, create `Config` headers to share values between the embedded and gtest targets and they are placed under `Inc/App/configs`.
- Simplify the way we write `TearDown` by creating 
```
template <typename T>
void TearDownObject(T *&object, void (*destroy_func)(T *object))
{
    ASSERT_TRUE(object != NULL);
    destroy_func(object);
    object = NULL;
}
```
- The fake functions in `Test_StateMachine` can have name collision with other `Test.cpp` (e.g. `get_pwm_frequency`). To get around this, I guarded `Test_StateMachine` with its own namespace rather than prefixing arbitrary words like `prv_` to function names as a means to avoid name collision.

### Checklist


boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -68,38 +76,19 @@ class BmsStateMachineTest : public ImdTest
...
-        ASSERT_TRUE(state_machine != NULL);
-        App_SharedStateMachine_Destroy(state_machine);
-        state_machine = NULL;
+        TearDownObject(world, App_BmsWorld_Destroy);
+        TearDownObject(state_machine, App_SharedStateMachine_Destroy);
+        TearDownObject(can_tx_interface, App_CanTx_Destroy);
+        TearDownObject(can_rx_interface, App_CanRx_Destroy);
+        TearDownObject(imd, App_Imd_Destroy);
+        TearDownObject(heartbeat_monitor, App_SharedHeartbeatMonitor_Destroy);
+        TearDownObject(rgb_led_sequence, App_SharedRgbLedSequence_Destroy);
```
garethellis0, line 85: Very nice :smiley:

boards/DCM/Test/Inc/Test_Dcm.h
```diff
@@ -2,3 +2,11 @@
...
 #include <fff.h>
 #include <gtest/gtest.h>
+
+template <typename T>
+void TearDownObject(T *&object, void (*destroy_func)(T *object))
+{
+    ASSERT_TRUE(object != NULL);
+    destroy_func(object);
+    object = NULL;
+}
```
garethellis0, line 12: Is this duplicated with the other boards?

thekenu: I thought about only defining this function once and sharing it across the other boards but wasn't sure where to put it. The most obvious choice seems to be creating a `Test_Shared.h`, but I felt like that was up for debate because the test files under `shared` are sort of standalone. That is, we create a target named `SHARED_test` that can be compiled on its own without other `BOARD_test` targets. I felt like the abstraction would become kind of weird if I start including `Test_Shared.h` in all the `BOARD_test` targets like:

```
// Test_Shared.h
void TearDownObject(...) {}
// Test_Bms.h
#include "Test_Shared.h" // Does this include make sense if we
                                         // can compile SHARED_test on its own, 
                                         // independent of BMS?
```

Maybe i'm just overthinking it, but that's why i defined the function multiple times.

garethellis0: I would make a folder called `shared/test_utils` and put it in there to make it clear that it's not part of the shared tests, it's just shared testing code that anything (including the shared tests) can use.

thekenu: I like it, nice and simple. Added it in !

PR #593, opened 2020-05-06
Add brake light to DCM world

thekenu:
### Checklist


boards/DCM/Inc/App/App_BrakeLight.h
```diff
@@ -2,12 +2,6 @@
 
 #include <stdbool.h>
```
thekenu, line 4: Did some slight clean up in this PR

- Removed `enum` and replaced it with a simpler boolean
- Cleaned up some unncessary `const`
- Touch up the test files according to the last 2 points

boards/DCM/Src/main.c
```diff
@@ -116,6 +118,17 @@ static void CanTxQueueOverflowCallBack(size_t overflow_count)
     App_CanTx_SetPeriodicSignal_TX_OVERFLOW_COUNT(can_tx, overflow_count);
 }
 
+static bool IsBrakeActuated(void)
```
thekenu, line 121: @garethellis0 I was just about to merge this but I realized this design is hard to test in `Test_StateMachine` because the logic for checking brake actuation and regen active is _not_ directly inside the state machine. I feel like I want to write something like
```
// Test_StateMachine.cpp
Test(turn_on_brake_light_when_brake_is_active) {
  App_CanRx_FSM_BRAKE_SetSignal_BRAKE_IS_ACTUATED(can_tx, true);
  App_SharedStateMachine_Tick(state_machine);
  ASSERT_EQ(App_BrakeLight_IsTurnedOn(brake_light), true);
}
```

But with the current design, I would write
```
// Test_StateMachine.cpp
FAKE_VALUE_FUNC(bool, is_brake_active)
Test(turn_on_brake_light_when_brake_is_active) {
  is_brake_active_fake.return_vale = true;
  App_SharedStateMachine_Tick(state_machine);
  ASSERT_EQ(App_BrakeLight_IsTurnedOn(brake_light), true);
}
```

What do you think about this?

garethellis0: I think the first test should definitely work if the trigger for the brake light turning on is supposed to be the CAN message. Seems like we're going to have to implement that functionality one way or the other eventually?

thekenu: Oh the problem is that `IsBrakeActuated` is IO code and doesn't get compiled into the test target. I guess the problem is to move `IsBrakeActuated` to the App layer?

thekenu: Following up on this, I think moving `IsBrakeAcutaed` and `IsRegenActive` to some `App` file is the right move. What should we name this file though? Can't do `App_BrakeLight.c` because that's taking already? Maybe we can have `App/Src/configs/App_BrakeLightConfig.c` like how I have `App/Inc/configs/XXXConfig.h` like I have in #591 right now?

garethellis0: `IsBrakeActuated` should just be checking a CAN signal right? If so I'd say yeah, that should be in the App layer.

garethellis0: Why do you think it should have a `Config` suffix? 

We could just rename `App_BrakeLight` to `App_Brake` and put everything in there?

thekenu: Maybe the `Config` suffix was a poor example, but my understanding for what belongs in `App_BrakeLight` is the brake light `class` itself. Things like `IsBrakeActuated` and `IsRegenActive` belongs to a specific instance of brake light class, so it seems weird to me to mix those functions in one file.

```
// App_Brake.h

// These belong to a specific _instance_ of the brake light class
App_Brake_IsBrakeActive();
App_Brake_IsRegenActive();

// The following functions make up the brake light class
App_BrakeLight_Create();
App_BrakeLight_Destroy();
App_BrakeLight_Tick();
App_BrakeLight_IsTurnedOn();
```

Also I just realized if we were to write 
```
App_Brake_IsBrakeActive();
App_Brake_IsRegenActive();
```
We would need another function to register `can_tx` because `IsBrakeActive` and `IsRegenActive` rely on it.
```
App_Brake_Init(can_tx);
```
For this last reason alone, I feel like it would be nicer to move the logic into the state machine so we don't need a dedicated init function just for registering `can_tx`

garethellis0: I'm fine with moving logic into the state machine, and also have `App_Brake` and `App_BrakeLight`. Neither `App_Brake` nor `App_BrakeLight` should have any knowledge of CAN.

thekenu: That sounds good but I'm still have a bit of a hard time isolating CAN from `App_Brake` and `App_BrakeLight`. The best solution I've come up so far is to __not__ pass in getter functions `IsBrakeAcuated` and `IsRegenActivated` to `App_BrakeLight_Create`, and instead pass arguments by value into `App_BrakeLight_Tick()`

```
// App_BrakeLight.c
struct BrakeLight *App_BrakeLight_Create(void (*turn_on_brake_light)(void), void (*turn_off_brake_light)(void));
struct BrakeLight* App_BrakeLight_Tick(bool is_brake_actuated, bool is_regen_active);

// App_DriveState.c
void DriveStateRunOnTick(struct StateMachine* state_machine) {
  DcmWorld* world = get_world(state_machine);
  DcmCanTxInterface* can_tx = get_can_tx(world);
  DcmCanRxInterface* can_rx = get_can_tx(world);
  BrakeLight* brake_light = get_brake_light(world);

  brake_light_tick(App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx) == true,
                           App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx) < 0.0f);
}
```

It seems this solution wouldn't even need `App_Brake`, and `App_BrakeLight` has no knowledge of CAN. I don't really know if this is what you had in mind though. Feel free to write out some pseudo-code describing what you had in mind.

garethellis0: Sorry, I wasn't suggesting any particular implementation, merely indicating a conceptual rule we should be following to isolate things for testability. I think what you have above is pretty good, I might write the bottom bit more like:

```
bool brake_actuated = App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx);
bool regen_active = App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx) < 0.0f;

App_BrakeLight_SetLightStatus(brake_actuated, regen_active);
```
The main thing is that `tick` could mean anything, I'd prefer to have a slightly more descriptive name (though I really struggled to come up with anything here).

thekenu: Oh yeah I think anything is more clear than tick here, will make the necessary changes!

boards/DCM/Inc/App/states/App_AllStates.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "App_SharedStateMachine.h"
+
+/**
+ * On-tick function for every state in the given state machine
+ * @param state_machine The state machine to run on-tick function for
+ */
+void App_AllStatesRunOnTick(struct StateMachine *state_machine);
```
thekenu, line 9: @garethellis0 I didn't want to write this code snippet 3 times (once for each state)
```
    struct DcmWorld *world = App_SharedStateMachine_GetWorld(state_machine);
    struct DcmCanTxInterface *can_tx      = App_DcmWorld_GetCanTx(world);
    struct DcmCanRxInterface *can_rx      = App_DcmWorld_GetCanRx(world);
    struct BrakeLight *       brake_light = App_DcmWorld_GetBrakeLight(world);

    bool is_brake_actuated =
        App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx);
    bool is_regen_active =
        App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx) < 0.0f;

    App_BrakeLight_SetLightStatus(
        brake_light, is_brake_actuated, is_regen_active);
```

It didn't really belong in `SetPeriodicCanSignals.c` too, so I created this file as a way to share functions that should run in every state. What do you think?

PR #595, opened 2020-05-07
Split SharedStateMachine Tick Function into 1Hz and 1kHz Tick Functions

garethellis0:
### Checklist


boards/DCM/Inc/App/errors.h
```diff
@@ -0,0 +1 @@
+#include "App_CanRx.h"
```
thekenu, line 1: extra file?

boards/shared/Src/App/App_SharedStateMachine.c
```diff
@@ -10,6 +11,38 @@ struct StateMachine
...
+    if (tick_function == NULL)
+    {
+        // Nothing to do
+        return;
+    }
+
+    tick_function(state_machine);
+
+    // Check if we should transition states
+    if (state_machine->next_state != state_machine->current_state)
```
thekenu, line 34: Can we have write a quick test to make sure that when `Tick1Hz` changes the state from `State A` to `State B`, the next `Tick1kHz` will tick in `State B` as well? In other words, the state change is propagated between the different tick functions.

boards/DCM/Test/Src/Test_StateMachine.cpp
```diff
@@ -109,8 +109,8 @@ TEST_F(
 {
     // We need to tick twice, once to run the `Init` state, and once more
     // to have the state machine transition to the `Run` state.
-    App_SharedStateMachine_Tick(state_machine);
-    App_SharedStateMachine_Tick(state_machine);
+    App_SharedStateMachine_Tick1Hz(state_machine);
```
thekenu, line 112: I wonder how our `Test_StateMachine` is going to accommodate for different tick rates. For example, should this test be ticking `1Hz` or `1Khz` or both? It seems a bit tricky to me.

boards/shared/Src/App/App_SharedStateMachine.c
thekenu: Adding `Test_SharedStateMachine` under `shared/Test/Src` would be great

boards/DCM/Test/Src/Test_StateMachine.cpp
garethellis0: Yeah, we're probably going to just end up testing one or the other.

boards/DCM/Src/App/states/App_FaultState.c
```diff
@@ -23,10 +23,11 @@ static void FaultStateRunOnExit(struct StateMachine *const state_machine)
 const struct State *App_GetFaultState(void)
 {
     static struct State fault_state = {
-        .name         = "FAULT",
-        .run_on_entry = FaultStateRunOnEntry,
-        .run_on_tick  = FaultStateRunOnTick,
-        .run_on_exit  = FaultStateRunOnExit,
+        .name             = "FAULT",
+        .run_on_entry     = FaultStateRunOnEntry,
+        .run_on_tick_1Hz = FaultStateRunOnTick,
```
thekenu, line 28: nit; for this PR can we leave the `OnTick` functions as 1kHz rather 1Hz? I believe all the state machines are ticking at 1kHz at the moment

boards/DCM/Inc/App/errors.h
garethellis0: yep, removed

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -25,7 +25,8 @@ const struct State *App_GetDriveState(void)
     static struct State drive_state = {
         .name         = "DRIVE",
         .run_on_entry = DriveStateRunOnEntry,
-        .run_on_tick  = DriveStateRunOnTick,
+        .run_on_tick_1Hz  = DriveStateRunOnTick,
```
thekenu, line 28: I wanna say it would make sense to have separate on entry and on exit for 1hz and 1khz?

boards/shared/Src/App/App_SharedStateMachine.c
garethellis0: Agreed. Added.

boards/DCM/Src/App/states/App_DriveState.c
garethellis0: I don't think so, can you think of any examples?

boards/DCM/Src/App/states/App_FaultState.c
garethellis0: sure, no problem

boards/DCM/Src/App/states/App_DriveState.c
thekenu: Hmm don't have a concrete example for now so I guess we can leave it out (should be easy to extend later on as the need arises)

boards/shared/CMakeLists.txt
```diff
@@ -42,7 +42,10 @@ set(SHARED_IO_INCLUDE_DIRS ${SHARED_IO_INCLUDE_DIRS} PARENT_SCOPE)
 file(GLOB_RECURSE ARM_BINARY_APP_SRCS "Src/App/*.*")
 set(ARM_BINARY_APP_INCLUDE_DIRS "Inc/App")
 
-file(GLOB GOOGLETEST_TEST_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/*.cpp")
+file(GLOB GOOGLETEST_TEST_SRCS
```
thekenu, line 45: Aren't the GTest source files always going to be in `.cpp` because it's a C++ framework? Not sure why this change is necessary.

boards/shared/Inc/App/App_SharedStateMachine.h
```diff
@@ -15,10 +15,12 @@ struct State
     char name[MAX_STATE_NAME_LENGTH];
 
     void (*run_on_entry)(struct StateMachine *state_machine);
-    void (*run_on_tick)(struct StateMachine *state_machine);
+    void (*run_on_tick_1Hz)(struct StateMachine *state_machine);
+    void (*run_on_tick_1kHz)(struct StateMachine *state_machine);
     void (*run_on_exit)(struct StateMachine *state_machine);
 };
 
+
```
thekenu, line 23: nit; extra line

boards/shared/Inc/App/App_SharedStateMachine.h
```diff
@@ -65,7 +67,13 @@ struct World *
     App_SharedStateMachine_GetWorld(const struct StateMachine *state_machine);
 
 /**
- * Tick the given state machine
+ * Tick the 10hz function of the given state machine
```
thekenu, line 70:
```suggestion
 * Tick the 1hz function of the given state machine
```

boards/shared/Test/Src/App_TestWorld.c
```diff
@@ -0,0 +1,15 @@
+#include "App_TestWorld.h"
+
+#include <stdlib.h>
+
+struct TestWorld {};
```
thekenu, line 5: I believe our compiler flags aren't happy about empty structs (But I assume you'd find out about that eventually anyway)

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,93 @@
...
+        state_machine = App_SharedStateMachine_Create(world, initial_state);
+        ASSERT_TRUE(state_machine != NULL);
+        ASSERT_EQ(
+            initial_state,
+            App_SharedStateMachine_GetCurrentState(state_machine));
+    }
+
+    // We provide our own implementation of the 1hz tick for state_A
+    // here so that we can simulate a state transition in a tick
+    static void state_A_tick_1Hz(struct StateMachine* state_machine){
```
thekenu, line 72: Is there any particular reason for this to be static?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,93 @@
...
+        RESET_FAKE(state_B_tick_1Hz);
+        RESET_FAKE(state_B_tick_1kHz);
+        RESET_FAKE(state_B_exit);
+    }
+
+    void TearDown() override
+    {
+        ASSERT_TRUE(world != NULL);
+        App_TestWorld_Destroy(world);
+
```
thekenu, line 54:
```suggestion
        ASSERT_TRUE(world != NULL);
        App_TestWorld_Destroy(world);
        world = NULL;

```

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -25,7 +25,8 @@ const struct State *App_GetDriveState(void)
     static struct State drive_state = {
         .name         = "DRIVE",
         .run_on_entry = DriveStateRunOnEntry,
-        .run_on_tick  = DriveStateRunOnTick,
+        .run_on_tick_1Hz  = NULL,
+        .run_on_tick_1kHz = DriveStateRunOnTick,
```
thekenu, line 29: Maybe we can start renaming the functions in this PR? Up to you, not super for or against it.
```suggestion
        .run_on_tick_1kHz = DriveStateRunOnTick1kHz,
```

boards/shared/CMakeLists.txt
garethellis0: In this case I had an odd issue where I had to make the `TestWorld.c` file a `.c` otherwise I was getting undefined references, because we include `TestWorld.h` in an `extern C`. I think this is good because it's consistent with how we create and include `World`'s elsewhere, but open to changing it if you feel otherwise.

boards/shared/Inc/App/App_SharedStateMachine.h
garethellis0: removed

boards/shared/Inc/App/App_SharedStateMachine.h
garethellis0: Good catch, changed.

boards/shared/Test/Src/App_TestWorld.c
garethellis0: The test seems to compile fine, and this isn't ever included in an `arm` executable, so I think it should be Ok?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: Yeah, this won't compile if it's a member function because the type won't match (which is definitely not immediately obvious). You'll get:
```
error: cannot convert ‘SharedStateMachineTest::state_A_tick_1Hz’ from type ‘void (SharedStateMachineTest::)(StateMachine*)’ to type ‘void (*)(StateMachine*)’
```

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: Good point, added.

boards/DCM/Src/App/states/App_DriveState.c
garethellis0: I think I'll leave it as-is for now, as there may be things in the `RunOnTick` functions that belong in the `1Hz` task, and thus the name wouldn't be totally correct until we do an audit in a future PR to move stuff over.

boards/shared/Test/Src/App_TestWorld.c
thekenu: Oh yeah you're right, I assumed it's part of the arm target because it's `.c`

boards/shared/CMakeLists.txt
thekenu: Oh I see the dilemma now. I feel like since `TestWorld.c` is an `App` file it would make more sense to append it to `ARM_BINARY_APP_SRCS`, i.e.:

```
list(APPEND ARM_BINARY_APP_SRC "Src/App/App_TestWorld.c)
```
And leave this line as-is. What do you think?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
thekenu: Oh ah that totally makes sense :+1:

boards/shared/Src/App/App_SharedStateMachine.c
```diff
@@ -10,6 +11,38 @@ struct StateMachine
...
 };
 
+/**
+ * Run the given tick function over the given state machine if the tick function
+ * is not null
+ *
+ * @param state_machine The state machine to run the tick function over
+ * @param tick_function The tick function to run over the state machine
+ */
+void App_SharedStateMachine_RunStateTickFunctionIfNotNull(
```
thekenu, line 21: I was a little bit worried about whether changing the state machine to be multithreaded cause any race condition. One potential race condition I could think of:


Assuming the following:
1. We have `OnTick1Khz` and `OnTick1Hz`
2. `Task1Khz` is higher priority than `Task1Hz`
3. We have two states - `INIT` and `FAULT`, and we default to `INIT`

```
OnTick1Khz() {
  SetNextState(INIT);
}

OnTick1Hz() {
  SetNextState(FAULT);
  // An task switch to Task1kHz  occurs before we return from this function
}
```

It seems to me that if `OnTick1Hz` requests a transition to `FAULT` state and then gets preempted by `OnTick1kHz` (which overrides NexState to be `INIT`), then we never transition to `FAULT` state. Obviously this example is contrived but I think you get the scenario I am trying to describe.

garethellis0: Ok, added a semaphore lock here that should prevent this.

garethellis0:
> Another concern that came to me is, how should we adapt our `specs.md` if the state machine ticks at different rates. Some specifications don't have a tick rate. For example,
> 
> > FSM-13 | Entering the AIR-Closed state | The FSM must enter the AIR-Closed state when the BMS closes the contactors.
> 
> Would `OnTick1Hz` handle this or `OnTick1Khz`, or both?

IMO if it's particularly time sensitive it should be phrased as `The FSM must enter the AIR-Closed state when the BMS closes the contactors for more then N ms`. But that's true whether or not we have multiple tick rates, the specs should be independent of that. If there's no duration `N` specified it's up the discretion of the developer.

boards/shared/Src/App/App_SharedStateMachine.c
thekenu: That's quite a big section to guard but I've got no better idea at the moment so let's run with it. I think we should leave the semaphore inside `struct StateMachine` though?

Also you need to initialize it: https://www.freertos.org/xSemaphoreCreateMutexStatic.html

garethellis0: Good catch, added.

boards/shared/Src/App/App_SharedStateMachine.c
```diff
@@ -10,10 +16,59 @@ struct StateMachine
...
+{
+#ifdef __arm__
+    xSemaphoreTake(state_tick_mutex, portMAX_DELAY);
+#endif
+
+    if (tick_function == NULL)
+    {
+        // Nothing to do
+        return;
+    }
```
thekenu, line 43: Shouldn't grab the semaphore if the tick function is NULL because we'd never release it! Noticed it after testing it on the actual hardware.
```suggestion
    if (tick_function == NULL)
    {
        // Nothing to do
        return;
    }
    
#ifdef __arm__
    xSemaphoreTake(state_tick_mutex, portMAX_DELAY);
#endif
```

boards/shared/CMakeLists.txt
thekenu: @garethellis0 Think you missed this comment

boards/shared/Src/App/App_SharedStateMachine.c
thekenu: Looks like the semaphore is initialized properly now, but it's still not part of `struct StateMachine`. Was that a deliberate choice? It seems to make more sense to have a per-state machine mutex. Suppose we ever have multiple state machines, you would want an unique mutex for each of them?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,95 @@
...
+    void TearDown() override
+    {
+        ASSERT_TRUE(world != NULL);
+        App_TestWorld_Destroy(world);
+        world = NULL;
+
+        ASSERT_TRUE(state_machine != NULL);
+        App_SharedStateMachine_Destroy(state_machine);
+        state_machine = NULL;
```
thekenu, line 58: nit; Could start using the `TearDownObject` I created in #591

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    void SetInitialState(const struct State *const initial_state)
+    {
+        App_SharedStateMachine_Destroy(state_machine);
```
thekenu, line 63: Same as https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/595/files#r422556483

boards/shared/CMakeLists.txt
garethellis0: But that's a test file, it should never be built into the arm binary, no?

boards/shared/Src/App/App_SharedStateMachine.c
garethellis0: Whoops, I did actually get part way through doing that design, but bamboozled myself into thinking each mutex needed a static storage duration..... anyhow you're totally right. Moved the mutex and it's storage to the `StateMachine` struct and changed everything else to work with that. 

:+1:

boards/shared/CMakeLists.txt
garethellis0: Anyhow, I toyed with this a bit more and got it to work as a `.cpp`, so removed the `*.c` match.

boards/shared/Src/App/App_SharedStateMachine.c
garethellis0: Nice catch. I added mutex's via `pthread` on unix so I was able to write a unit test to reproduce this as well. Fixed!

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: Nice :smile: changed.

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: changed

garethellis0:
Ok, would appreciate another pass, really don't want to go changing all the other boards until we're as certain as we can be that there aren't other issues I'll have to go change across all the boards after.

boards/shared/CMakeLists.txt
thekenu: Oh yeah I see what you're saying, i'm fine with the either way then: either replacing `.c` with `.cpp` or adding `"${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/*.c"` back.

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        TearDownObject(world, App_TestWorld_Destroy);
+        TearDownObject(state_machine, App_SharedStateMachine_Destroy);
+    }
+
+    void SetInitialState(const struct State *const initial_state)
+    {
+        App_SharedStateMachine_Destroy(state_machine);
+        state_machine = App_SharedStateMachine_Create(world, initial_state);
+        ASSERT_TRUE(state_machine != NULL);
```
thekenu, line 70: nit; This might be slightly better? I later realized `_Create` already fails if the pointer is NULL so there's no need to do another NULL check.
```suggestion
        TearDownObject(state_machine, App_SharedStateMachine_Destroy);
        state_machine = App_SharedStateMachine_Create(world, initial_state);
```

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <fff.h>
+
+#include "Test_Utils.h"
+#include "App_TestWorld.h"
+
+extern "C"
+{
+#include "App_SharedStateMachine.h"
+}
+
```
thekenu, line 11:
```suggestion
```

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,115 @@
+#include <gtest/gtest.h>
+#include <fff.h>
+
+#include "Test_Utils.h"
+#include "App_TestWorld.h"
```
thekenu, line 5: It is possible to move this include inside `extern C` to better match other test binaries? Or would that break compilation?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
```diff
@@ -0,0 +1,115 @@
...
+    static void state_A_tick_1Hz(struct StateMachine *state_machine)
+    {
+        App_SharedStateMachine_SetNextState(state_machine, &state_B);
+    }
+
+    struct World *       world;
+    struct StateMachine *state_machine;
+};
+
+TEST_F(
```
thekenu, line 87: Maybe add another basic test that ticks a NULL as  well as a non-NULL function, and assert on the `call_count`?

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: That will break compilation, this is basically the issue we discussed above.

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: Sorry, could you be a bit more specific in terms of what functionality you're trying to test here that's not covered in the existing tests?

thekenu: Oh just checking that we don't tick if it's NULL

thekenu:
Whoops forgot to follow up, but i couldn't find anymore things to change so I think we can port over to other boards

boards/shared/Test/Src/Test_SharedStateMachine.cpp
garethellis0: What exactly would you be asserting? If the tick function is NULL we can't very well check that it's never called, can we?

thekenu: Oh yeah of course, whoops :facepalm:

PR #596, opened 2020-05-08
wip

markomarkusi:
### Summary
Io and App code for interfacing with the Efuses over SPI. Current sensing and Voltage sensing Io layer will be added after this is merged in.

### Testing Done
Tested writing to and reading from the Efuses. The Efuse's registers contained the correct configurations after being read back from. The status and fault registers correctly relayed the running state of the Efuse. Read current's from the Efuse's current-sense pin; measurements were within 5% of the actual value. 

### Checklist


thekenu:
Since this is your first PR, can we break it down into smaller pieces? One for clang format, then one for io, and then app?

markomarkusi:
> Since this is your first PR, can we break it down into smaller pieces? One for clang format, then one for io, and then app?

Sure

PR #604, opened 2020-05-12
Add specification ID to test case names

thekenu:
### Summary

Where applicable, add  specification ID to test case names

### Checklist


boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -179,7 +179,7 @@ class DimStateMachineTest : public testing::Test
     struct RotarySwitch *     drive_mode_switch;
 };
 
-TEST_F(DimStateMachineTest, check_drive_state_is_broadcasted_over_can)
+TEST_F(DimStateMachineTest, check_drive_state_is_broadcasted_over_can_DIM_12)
```
thekenu, line 182: I thought about leaving the ID as a comment, but it seems more concise to just leave the ID in the test case name

boards/PDM/Test/Src/Test_StateMachine.cpp
```diff
@@ -390,7 +396,9 @@ TEST_F(PdmStateMachineTest, check_24v_acc_voltage_can_signals_in_all_states)
         CANMSGS_PDM_ERRORS__24_V_ACC_VOLTAGE_OUT_OF_RANGE_OVERFLOW_CHOICE);
 }
 
-TEST_F(PdmStateMachineTest, check_aux1_current_can_signals_in_non_init_states)
+TEST_F(
+    PdmStateMachineTest,
+    check_aux1_current_can_signals_in_non_init_states_PDM_11_13)
```
thekenu, line 401: This means this test applies to two IDs - 11 and 13

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -113,7 +113,7 @@ class BmsStateMachineTest : public testing::Test
     struct RgbLedSequence *   rgb_led_sequence;
 };
 
-TEST_F(BmsStateMachineTest, check_init_state_is_broadcasted_over_can)
+TEST_F(BmsStateMachineTest, check_init_state_is_broadcasted_over_can_BMS_31)
```
garethellis0, line 116: What do you think about `BMS_31__check_init_state_is_broadcasted_over_can` or `check_init_state_is_broadcasted_over_can__BMS_31`? Just trying to make it a bit more clear that `BMS_31` is somewhat seperate.

boards/PDM/Test/Src/Test_StateMachine.cpp
garethellis0: I think I'd prefer `PDM_11_to_13`, just to be super explicit?

boards/DIM/Test/Src/Test_StateMachine.cpp
garethellis0: Either one is fine by me.

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -244,7 +244,46 @@ TEST_F(
...
+    App_SharedStateMachine_Tick(state_machine);
+    ASSERT_EQ(
+        100,
+        App_CanTx_GetPeriodicSignal_MAPPED_PADDLE_POSITION(can_tx_interface));
+    get_raw_paddle_position_fake.return_val = 96;
+    App_SharedStateMachine_Tick(state_machine);
+    ASSERT_EQ(
+        100,
+        App_CanTx_GetPeriodicSignal_MAPPED_PADDLE_POSITION(can_tx_interface));
+}
```
garethellis0, line 282: This doesn't seem like part of this PR?

boards/BMS/Test/Src/Test_StateMachine.cpp
thekenu: I think that's a good point, let me add the double underscore.

boards/DIM/Test/Src/Test_StateMachine.cpp
thekenu: Ah yes forgot to leave a comment but as i was adding the specification ID I realized we forgot to write a test case for this particular ID. Slightly out-of-scope but I just quickly wrote a test for it.

boards/PDM/Test/Src/Test_StateMachine.cpp
thekenu: Yeah that's a good point i'll fix that

boards/DIM/Test/Src/Test_StateMachine.cpp
thekenu: Actually after considering the ambiguity you pointed out (e.g. single vs double underscore, misinterpreting multiple IDs as a contiguous range of IDs), I think it's better to leave things as comments.

thekenu:
@garethellis0 Ready for another review :octocat:

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -188,6 +189,7 @@ TEST_F(DimStateMachineTest, check_drive_state_is_broadcasted_over_can)
         App_CanTx_GetPeriodicSignal_STATE(can_tx_interface));
 }
 
+// DIM_9
```
garethellis0, line 192: We should consistent about using `-` vs `_` in these specs eh? I think `-` better matches the specs markdown doc.

thekenu: Definitely unintentional, will fix

thekenu: Should be fixed now!

PR #606, opened 2020-05-12
Updated Io_SharedAdc

pasitlaothamatas:
### Summary

Updated shared ADC

### Changelist 

- Updated the previous version of Io_Shared_Adc to allow more than one ADC peripheral to be initialised. This is done in `Io_SharedAdc_Init` in `Io_SharedAdc.c`

### Checklist


boards/FSM/Inc/Io/Io_SharedFreqOnlyPwmInputConfig.h
```diff
@@ -1,3 +1,3 @@
 #pragma once
 
-#define MAX_NUM_OF_FREQ_ONLY_PWM_INPUTS 2
```
pasitlaothamatas, line 3: Unrelated to the PR but increasing this due to the addition of Io_WheelSpeedSensor_Init

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
-/** @brief Number of VREFINT channel. VREFINT is the internal voltage reference 
- *         that provides a stable (bandgap) voltage output for the ADC. */
-#define NUM_VREFINT_CHANNEL 1
+struct AdcInput;
 
 /**
- * @brief  Initialize ADC in DMA mode and the correct max ADC value
- * @param  hadc ADC handle
- * @param  vrefint_regular_rank The Regular Rank of VREFINT channel configured
- *         in STM32CubeMX (Starts at 1)
```
pasitlaothamatas, line 31: image

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+ * Allocate and initialize an ADC with direct memory access
+ *
+ * @note The user must ensure that only one of the two ADCs is converting
+ * VREFINT as it is forbidden to have several ADCs converting
+ * VREFINT at the same time. Create the ADC instance containing VREFINT first
```
pasitlaothamatas, line 19: Should I refer to AdcInput as "ADC Instance" ? I feel like the functionality of AdcInput could be better conveyed

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+struct AdcInput;
 
 /**
- * @brief  Initialize ADC in DMA mode and the correct max ADC value
- * @param  hadc ADC handle
- * @param  vrefint_regular_rank The Regular Rank of VREFINT channel configured
- *         in STM32CubeMX (Starts at 1)
+ * Allocate and initialize an ADC with direct memory access
+ *
+ * @note The user must ensure that only one of the two ADCs is converting
```
pasitlaothamatas, line 17: image

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    if (alloc_index == 0U)
     {
-        status = false;
-        Error_Handler();
+        // Initialize ADC instance containing VREFINT before initializing other
+        // ADCs
+        assert(vrefint_index <= adc_input->num_active_channels);
+
+        adc_inputs->vrefint_index = vrefint_index - 1;
+        // Store the pointer pointing to the measured VREFINT value
```
pasitlaothamatas, line 81: https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/606/files#r423530092

boards/shared/Inc/Io/Io_SharedAdc.h
garethellis0: This seems fine to me.

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
-    if (InitializeAdcMaxValue(hadc) != SUCCESS)
+    if (alloc_index == 0U)
     {
-        status = false;
-        Error_Handler();
+        // Initialize ADC instance containing VREFINT before initializing other
+        // ADCs
+        assert(vrefint_index <= adc_input->num_active_channels);
+
+        adc_inputs->vrefint_index = vrefint_index - 1;
```
garethellis0, line 80: This is a bit confusing to me. It seems like there are two different `vrefint_index`'s here? Why are we subtracting one?

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
-// Memory address where you can find the internal reference voltage calibration 
-// values: Raw ADC data acquired at temperature of 30 degC with VDDA = 3.3V 
-// during the manufacturing process
-#define VREFINT_CALIBRATION_ADDRESS		(uint16_t *)(0x1FFFF7BA)
-
-// Number of microcontroller pins that are configured to be ADC inputs
-#ifndef NUM_ADC_CHANNELS
-    #error "Please define the number pins configured as ADC inputs using NUM_ADC_CHANNELS!"
+#ifndef MAX_NUM_OF_ADC
+#define MAX_NUM_OF_ADC 1
```
garethellis0, line 6: Why not `#error` here so that the user is forced to specify the number of ADC's?

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+#endif
+#ifndef MAX_NUM_OF_ADC_CHANNELS
+#define MAX_NUM_OF_ADC_CHANNELS 16
```
garethellis0, line 9: Similar comment here, why not `#error`?

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+ * Allocate and initialize an ADC instance with direct memory access
+ *
+ * @note The user must ensure that only one of the two ADCs is converting
+ * VREFINT as it is forbidden to have several ADCs converting
+ * VREFINT at the same time. Create the ADC instance containing VREFINT first
+ * while initializing the ADCs
+ *
+ * @note To set up the given ADC with a DMA controller, it must be initialized
+ * with (1) Continuous Conversion Mode enabled,(2) DMA Continuous Requests
+ * enabled, (3) DMA buffer with Circular Mode selected
```
garethellis0, line 24: nit, but could each numbered item here be on a new line? 
```
 * with:
 *     (1) Continuous Conversion Mode enabled
 *     (2) DMA Continuous Requests enabled
 *     (3) DMA buffer with Circular Mode selected
```

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+ *
+ * @param adc_handle The handle for the given ADC
+ * @param vrefint_rank The rank which is used to indicate the index of VREFINT
+ * @return A pointer to the allocated and initialized ADC instance
  */
-void Io_SharedAdc_StartAdcInDmaMode(
-    ADC_HandleTypeDef *hadc,
-    uint32_t           vrefint_regular_rank);
+struct AdcInput *
+    Io_SharedAdc_Init(ADC_HandleTypeDef *adc_handle, size_t vrefint_rank);
```
garethellis0, line 31: `vrefint_rank` is called `vrefint_index` in the implementation of this function. Personally I strongly prefer `index` to `rank`, but we should at least have them match either way.

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    //
+    // ACTUAL_VDDA is the vdda value computed in step (1)
+    // ADCx_DATA is the value of the channel measured by the ADC
+    // ADC_MAX_VALUE is the maximum ADC value
 
-    return (float32_t)(
-        3.3f * (float32_t)(GetVrefintCalibrationValue()) /
-        (float)adc_values[vrefint_index]);
+    const uint16_t *const VREFINT_CALIBRATION_ADDRESS =
+        (uint16_t *)(0x1FFFF7BA);
```
garethellis0, line 122: Could we add a quick comment indicating where we got this address from / if it's specific to a particular board or set of boards?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
-    // VDDA = ---------------------
-    //           VREFINT_DATA
+    //                    3.3V x VREFINT_CAL
+    //  ACTUAL_VDDA = ---------------------------
+    //                      VREFINT_MEASURED
     //
-    // Where:
-    // - VREFINT_CAL is the VREFINT calibration value
-    // - VREFINT_DATA is the actual VREFINT output value converted by the ADC
+    // VREFINT_CAL is the VREFINT calibration value (stored at 0x1FFF7BA)
```
garethellis0, line 108:
``` suggestion
    // VREFINT_CAL is the VREFINT calibration value
```
In general we want to avoid duplicating constants, and in particular in comments where they can quickly become out of date.

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
-    return (float32_t)(
-        3.3f * (float32_t)(GetVrefintCalibrationValue()) /
-        (float)adc_values[vrefint_index]);
+    const uint16_t *const VREFINT_CALIBRATION_ADDRESS =
+        (uint16_t *)(0x1FFFF7BA);
+    const float ACTUAL_VDDA =
+        (float)(3.3f * (uint16_t)(*VREFINT_CALIBRATION_ADDRESS) / *measured_vrefint_address);
+
+    return (
+        float)(ACTUAL_VDDA * adc_input->adc_readings[adc_rank - 1] / adc_input->adc_max_value);
```
garethellis0, line 127: Could we break this out a bit? Something like:
```
        float A = (float)ACTUAL_VDDA * adc_input->adc_readings[adc_rank - 1];
        float B = A / (float)adc_input->adc_max_value;
        return B;
```
Where `A` and `B` are replaced with more descriptive names (_we don't need to split it up exactly like this, just trying to understand what we're calculating here_).

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu: ADC input makes it seem like a specific ADC channel though, I think ADC instance is less ambiguous

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu, line 31: This could probably be renamed to `InitDmaMode`

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
 
 /**
- * @brief  Get the array of ADC readings transferred over DMA request
- * @return Pointer to the array of ADC readings
+ * Gets the pointer to the adc handle from the given ADC instance
+ * @param adc_input The given ADC instance to acquire the adc handle from
+ * @return The ADC handle for the specific ADC instance
  */
-const uint32_t *Io_SharedAdc_GetAdcValues(void);
+ADC_HandleTypeDef *Io_SharedAdc_GetAdcHandle(const struct AdcInput *adc_input);
```
thekenu, line 48: What's an actual use case for `Io_SharedAdc_GetAdcHandle` and `Io_SharedAdc_GetNumActiveChannel`?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+ * @return adc_max_resolution The maximum ADC resolution for the given ADC
+ * handle
  */
-static void CheckAdcInitialized(void);
+static ErrorStatus
+    Io_GetAdcResolution(ADC_HandleTypeDef *hadc, uint16_t *adc_max_resolution);
 
-static ErrorStatus InitializeAdcMaxValue(ADC_HandleTypeDef *hadc)
+static ErrorStatus
+    Io_GetAdcResolution(ADC_HandleTypeDef *hadc, uint16_t *adc_max_resolution)
```
thekenu, line 28: I feel like this function can be removed now if it's only used in one place?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+            &adc_input->adc_readings[adc_input->vrefint_index];
     }
-
-    // Only set adc_initialized to true if all initialization steps are
-    // error-free
-    if (status == true)
+    else
     {
-        adc_initialized = true;
+        UNUSED(vrefint_index);
```
thekenu, line 87: Can't this else-clause just be removed?

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu: Or honestly just ADC is fine too imo

boards/shared/Inc/Io/Io_SharedAdc.h
markomarkusi: @thekenu I think we don't want to use _Continuous Conversion Mode_
It should be triggered periodically by a Timer or launched by software in a periodic task.
I believe @garethellis0 observed that when operating the ADC in _Continuous Conversion Mode_ that the ADC conversions would happen too frequently and it would interfere with other tasks?

boards/shared/Src/Io/Io_SharedAdc.c
thekenu: Can we actually just have the user pass in the calibration value in `_Create`?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    adc_input->num_active_channels = hadc->Init.NbrOfConversion;
+    adc_input->adc_readings        = (uint16_t *)adc_readings[alloc_index];
+
+    HAL_ADCEx_Calibration_Start(hadc, ADC_SINGLE_ENDED);
+    HAL_ADC_Start_DMA(
+        hadc, (uint32_t *)adc_input->adc_readings,
+        adc_input->num_active_channels);
 
-    if (InitializeAdcMaxValue(hadc) != SUCCESS)
+    if (alloc_index == 0U)
```
thekenu, line 74: Why is this == 0 check needed?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
-        status = false;
-        Error_Handler();
-    }
+    adc_input->hadc                = hadc;
+    adc_input->num_active_channels = hadc->Init.NbrOfConversion;
+    adc_input->adc_readings        = (uint16_t *)adc_readings[alloc_index];
+
+    HAL_ADCEx_Calibration_Start(hadc, ADC_SINGLE_ENDED);
+    HAL_ADC_Start_DMA(
+        hadc, (uint32_t *)adc_input->adc_readings,
```
thekenu, line 71: Is this casting necessary?
```suggestion
        hadc, adc_input->adc_readings,
```

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    assert(Io_GetAdcResolution(hadc, &adc_input->adc_max_value) == SUCCESS);
 
-    if (HAL_ADC_Start_DMA(hadc, &adc_values[0], NUM_ADC_CHANNELS) != HAL_OK)
-    {
-        status = false;
-        Error_Handler();
-    }
+    adc_input->hadc                = hadc;
+    adc_input->num_active_channels = hadc->Init.NbrOfConversion;
+    adc_input->adc_readings        = (uint16_t *)adc_readings[alloc_index];
```
thekenu, line 67: If this is supposed to be u16 size, why isn't adc_readings just an array of u16 to begin with?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
- * @brief  Configure the maximum ADC value based on ADC resolution
- * @param  hadc ADC handle
- * @return ErrorStatus
- */
-static ErrorStatus InitializeAdcMaxValue(ADC_HandleTypeDef *hadc);
+struct AdcInput
+{
+    ADC_HandleTypeDef *hadc;
+    size_t             vrefint_index;
+    uint32_t           num_active_channels;
```
markomarkusi, line 10:
```suggestion
    uint8_t           num_active_channels;
```

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu: I super vaguely recall that was a problem but I don't remember all the details now. If the ADC continuous conversion takes up too many CPU cycles then we should consider not using it.

thekenu: Update: I (vaguely) remember the problem was that the ADC callback keeps getting called even though we don't do anything in it. And that's what hogged all the CPU cycle. If that's the case we can maybe investigate a way to disable the callback altogether since we don't do meaningful work in it?

markomarkusi: By default the callback is a weak function:
[__weak void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc)](https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/master/boards/FSM/Drivers/STM32F3xx_HAL_Driver/Src/stm32f3xx_hal_adc.c#L756)
There's no way to disable the call to that function, it is called after the DMA has transferred the ADC data into memory . We do use this callback. `void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc)`  should be implemented in the boards _Io_Adc.c_ file. In `void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc)` that's where you want to convert the ADC readings into values or first filter the results and then convert them

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu: To chime in, `rank` is a hardware-specific term. If you want  to leave it as rank you should make the description more explicit AND tell the user that it should be a value of `ADCEx_regular_rank`

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    return adc_input;
 }
 
-float32_t Io_SharedAdc_GetActualVdda(void)
+float Io_SharedAdc_GetChannelVoltage(
+    const struct AdcInput *const adc_input,
+    size_t                       adc_rank)
 {
-    CheckAdcInitialized();
+    assert(adc_rank > 0 && adc_rank <= adc_input->num_active_channels);
```
thekenu, line 98: There's a macro for this already `IS_ADC_REGULAR_RANK`

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
+    assert(alloc_index < MAX_NUM_OF_ADC);
+    assert(Io_GetAdcResolution(hadc, &adc_input->adc_max_value) == SUCCESS);
 
-    if (HAL_ADC_Start_DMA(hadc, &adc_values[0], NUM_ADC_CHANNELS) != HAL_OK)
-    {
-        status = false;
-        Error_Handler();
-    }
+    adc_input->hadc                = hadc;
+    adc_input->num_active_channels = hadc->Init.NbrOfConversion;
```
thekenu, line 66: Why is it called active channels? I can't find it in the documentation and it seems misleading. I feel like this variable is redundant anyway because you're already saving a copy of hadc. If you need `NbrOfConversion` then just write `adc_iput->hadc->Init.NbrOfConversion`

markomarkusi: I feel like this also falls apart if you do over-sampling. Ie. if you sample each channel four times in a conversion sequence then your `hadc->Init.NbrOfConversion` will be _4*NUM_OF_CHANNELS_

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -1,169 +1,140 @@
...
-{
-    CheckAdcInitialized();
 
-    return (const uint32_t *)adc_values;
+    alloc_index++;
+    return adc_input;
 }
 
-float32_t Io_SharedAdc_GetActualVdda(void)
+float Io_SharedAdc_GetChannelVoltage(
```
thekenu, line 94: I really feel like this should be renamed to

```
/**
 * Get the voltage reading for the given regular rank in the given ADC
 * @param ...
 * @param ...
 * Io_SharedAdc_GetVoltage(adc_input, adc_rank). 
 */
By mentioning channel as well as rank it's just making the code harder to read (though I know what you're trying to convey). You should just stick with using "ranks" imo and remove any mentions of"channel".

File: boards/shared/Src/Io/Io_SharedAdc.c [hunk 24]
```diff
@@ -1,169 +1,140 @@
...
-static void CheckAdcInitialized(void)
-{
-    if (adc_initialized != true)
-    {
-        Error_Handler();
-    }
-}
+    static size_t          alloc_index = 0;
+    static struct AdcInput adc_inputs[MAX_NUM_OF_ADC];
+    static uint32_t adc_readings[MAX_NUM_OF_ADC][MAX_NUM_OF_ADC_CHANNELS];
```
thekenu, line 58: I don't think this library should be allocating the buffer to which the DMA will be transferring data. For two reasons,

1. One of the main advantages of DMA is that we are moving data directly into where we need to. If we are statically allocating buffer and then exposing a getter function, then we probably should've have just used interrupt-based ADC. What we want to with ADC in DMA mode is something like:

```
// Io_CurrentSense.c


uint16_t adc_readings_in_volts[2];

// main.c
extern uint16_t adc_readings_in_volts[2];

int main() {
  Io_SharedAdc_Init(hadc2, VREFINT_RANK, adc_readings_in_volts);
  ..
} 

#define CURRENT_SENSE_CONVERSION_RATIO 1500
1kHzTask() {
  // Convert ADC readings to Amps
  for (int i = 0; i < 2; i++) {
  printf("The current at channel %u is %u\n", i, adc_readings_in_volts[i] * CURRENT_SENSE_CONVERSION_RATIO);  
  }
}


 
2.

**Comment by markomarkusi** [comment]
> @pasitlaothamatas Can you update the summary of the PR explaining the changes you made and why. Im assuming the main motivation is for when we need to use more than one ADC peripheral and in that case we need to call HAL_ADC_Start_DMA() on each ADC peripheral as well as have a buffer array for each ADC

When using two ADCs you could also configure it in [_Multi-mode_](https://www.st.com/resource/en/application_note/dm00069390-stm32f30x-adc-modes-and-application-stmicroelectronics.pdf#page=17):

File: boards/shared/Inc/Io/Io_SharedAdc.h [hunk 25]
```diff
@@ -1,67 +1,55 @@
...
 
-// Memory address where you can find the internal reference voltage calibration 
-// values: Raw ADC data acquired at temperature of 30 degC with VDDA = 3.3V 
-// during the manufacturing process
-#define VREFINT_CALIBRATION_ADDRESS		(uint16_t *)(0x1FFFF7BA)
-
-// Number of microcontroller pins that are configured to be ADC inputs
-#ifndef NUM_ADC_CHANNELS
-    #error "Please define the number pins configured as ADC inputs using NUM_ADC_CHANNELS!"
+#ifndef MAX_NUM_OF_ADC
```
thekenu, line 5: Is this referring to the hardware peripheral? If so can you leave a comment describing it?
```
The number of ADC peripherals in the microcontrollers. For example, STM302F302CCT6 has two: ADC1 and ADC2.
```

boards/shared/Inc/Io/Io_SharedAdc.h
```diff
@@ -1,67 +1,55 @@
...
+#define MAX_NUM_OF_ADC 1
+#endif
+#ifndef MAX_NUM_OF_ADC_CHANNELS
```
thekenu, line 8: What is this actually referring to?  Aren't there 18 channels? A comment to clarifying what this is supposed to be would be helpful

```
/* Note: Depending on devices, some channels may not be available on package  */
/*       pins. Refer to device datasheet for channels availability.           */
#define ADC_CHANNEL_1           ((uint32_t)(ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_2           ((uint32_t)(ADC_SQR3_SQ10_1))
#define ADC_CHANNEL_3           ((uint32_t)(ADC_SQR3_SQ10_1 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_4           ((uint32_t)(ADC_SQR3_SQ10_2))
#define ADC_CHANNEL_5           ((uint32_t)(ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_6           ((uint32_t)(ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_1))
#define ADC_CHANNEL_7           ((uint32_t)(ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_1 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_8           ((uint32_t)(ADC_SQR3_SQ10_3))
#define ADC_CHANNEL_9           ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_10          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_1))
#define ADC_CHANNEL_11          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_1 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_12          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_2))
#define ADC_CHANNEL_13          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_14          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_1))
#define ADC_CHANNEL_15          ((uint32_t)(ADC_SQR3_SQ10_3 | ADC_SQR3_SQ10_2 | ADC_SQR3_SQ10_1 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_16          ((uint32_t)(ADC_SQR3_SQ10_4))
#define ADC_CHANNEL_17          ((uint32_t)(ADC_SQR3_SQ10_4 | ADC_SQR3_SQ10_0))
#define ADC_CHANNEL_18          ((uint32_t)(ADC_SQR3_SQ10_4 | ADC_SQR3_SQ10_1))
```

boards/shared/Inc/Io/Io_SharedAdc.h
thekenu: Are you sure it can't be disabled? My understanding is that the DMA and ADC interrupts are separate. Can you test them and find out?

image

As for the converting the ADC readings, this PR seems to handle that in `Io_SharedAdc_GetChannelVoltage` so that's not really a problem. The idea here is to use DMA to store the raw ADC readings, and when the user wants the convert voltage they would call `Io_SharedAdc_GetChannelVoltage` to convert the raw ADC reading.

thekenu:
It appears if you're using two ADCs simultaneously then you must use the multimode-specific DMA functions. Imo we should just limit this PR to be using _one_ ADC, otherwise this will never end. Just remove `MAX_NUM_OF_ADC` and hard code it as 1, and leave a note in the jdocs saying that this is assuming one ADC without multi-mode.

boards/shared/Src/Io/Io_SharedAdc.c
thekenu: Yeah I was thinking about this too. I was thinking if there's a way to tweak the PR to accomodate for this. But as it stands, this would basically assume no oversampling of any kind and we always store only the latest copy of the ADC reading.

thekenu: For now, let's assume no oversampling in this PR. After everything is cleaned up let's revisit whether it's worth to making a shared library that assumes no oversampling.

boards/shared/Src/Io/Io_SharedAdc.c
pasitlaothamatas: Oh this is my bad actually. I intended to pass in the rank like how we did in the previous version, and store it as an index

boards/shared/Src/Io/Io_SharedAdc.c
garethellis0: That seems like a good option.

pasitlaothamatas:
Imma close this

thekenu:
After offline discussion, this PR is being replaced by #614

PR #610, opened 2020-05-13
 Change 1kHz state machine tick to be 100Hz, and add 1Hz state machine for each board

thekenu:
### Summary

The 1kHz state machine on tick functions has been changed to 100Hz because our CPUs don't quite run fast enough for that. And 100Hz control loop is sufficient for our use case. Also add a generic 1Hz on tick function for each board (To be used by RGB LED sequence later).

### Checklist


boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -0,0 +1,6 @@
+#include "states/App_AllStates.h"
+
+void App_AllStatesRunOnTick1Hz(struct StateMachine *const state_machine)
+{
+    (void)state_machine;
```
thekenu, line 5: Adding some skeleton function to be filled in by the RGB LED sequence later.

garethellis0: OK, let's use the `UNUSED` macro here just to be explicit and we know to remove it when we add actual code?

thekenu: Fixed

PR #612, opened 2020-05-14
Add test to check message transmit frequencies against specs.md

thekenu:
### Summary

Simple tests to check that the CAN TX frequencies match the specifications

### Checklist


boards/DIM/Test/Src/Test_CanMsgs.cpp
```diff
@@ -0,0 +1,45 @@
+#include "Test_Dim.h"
```
thekenu, line 1: Only have DIM for now, once this looks good after review I will add in the other boards.

garethellis0: In that case, should we change the PR name?

boards/DIM/Test/Src/Test_CanMsgs.cpp
```diff
@@ -0,0 +1,38 @@
...
+    CanMsgsTest,
+    start_and_traction_control_and_torque_vectoring_switches_message_frequency)
+{
+    ASSERT_EQ(HZ_TO_MS(100), CANMSGS_DIM_SWITCHES_CYCLE_TIME_MS);
+}
+
+// DIM-7
+TEST(CanMsgsTest, mapped_regen_paddle_percentage_message_frequency)
+{
+    ASSERT_EQ(HZ_TO_MS(100), CANMSGS_DIM_REGEN_PADDLE_CYCLE_TIME_MS);
```
garethellis0, line 37: This seems fine to me, but perhaps we want to check that the cycle time is just _less_ then the required maximum, rather then being exactly equal? This might require re-phrasing specs to be `greater then or equal to...`.

boards/DIM/Test/Src/Test_CanMsgs.cpp
thekenu: Oh yeah good point, changed

boards/DIM/Test/Src/Test_CanMsgs.cpp
thekenu: What about `10Hz or faster`?

garethellis0: Yep, seems fine to me.

PR #614, opened 2020-05-14
Set up ADC in DMA mode for regen paddle and clean up SharedAdc

thekenu:
### Summary

This is meant to replace #606 

Set up ADC2 in DMA mode for regen paddle in DIM. And greatly simplified the shared ADC module. Oh and stopped using VREFINT calibration because we know VDDA.

### Testing Done

- Measured the voltage with a multimeter and confirmed that it matches with what's seen on the CAN bus
- Manually inserted a CAN message in the callback function as a means to check that the ADC is triggered at the desired speed
- Use Tracealzyer to verify that the CPU usage is sane (not overwhelmed by callbacks) 
### Checklist


boards/DIM/Src/Io/Io_Adc.c
```diff
@@ -0,0 +1,28 @@
+#include <stm32f3xx.h>
+#include "Io_SharedAdc.h"
+#include "Io_Adc.h"
+
+enum
+{
+    CHANNEL_12,
```
markomarkusi, line 7: Should this be called REGEN_PADDLE or something more descriptive? I just called it by the channel number in the previous PR because I was to lazy to check what that channel was used for.

thekenu: I don't think this is the right place to include a descriptive name because this file is just related to ADC itself. It should have no knowledge about what other modules are using it.

boards/DIM/Src/main.c
```diff
@@ -510,6 +523,63 @@ static void MX_SPI2_Init(void)
...
+    /* USER CODE END TIM2_Init 0 */
+
+    TIM_ClockConfigTypeDef  sClockSourceConfig = { 0 };
+    TIM_MasterConfigTypeDef sMasterConfig      = { 0 };
+
+    /* USER CODE BEGIN TIM2_Init 1 */
+
+    /* USER CODE END TIM2_Init 1 */
+    htim2.Instance         = TIM2;
+    htim2.Init.Prescaler   = TIM2_PRESCALER - 1;
```
thekenu, line 544: @markomarkusi Is the `-1` needed or not? I always forget

markomarkusi: Yep 👍 that's right a value of 0 = prescaler of 1x and a value of 1 - prescaler of 2x

boards/DIM/Src/Io/Io_Adc.c
markomarkusi: 👍Sounds good

boards/DIM/Src/main.c
thekenu: What about for period?

thekenu: It's the value that it counts down from right? So if my clock is 1Hz and I set period as 2, it would count down from: 2 - 1 - 0 - 2 - 1 - 0. This seems like the `-1` isn't needed?

thekenu: Oh wait maybe you do need the `-1` based on the way I just counted above

markomarkusi: Ya because you count through 0 as well

boards/DIM/Src/Io/Io_Adc.c
garethellis0: Where does the `CHANNEL_12` name come from? I don't see any reason that we wouldn't want a more descriptive name here? I totally agree that we don't want to add _extra_ context to the ADC, but since we need a name here anyways we might as well use a descriptive one?

boards/DIM/Src/Io/Io_Adc.c
```diff
@@ -0,0 +1,28 @@
...
+    NUM_ADC_CHANNLES
```
pasitlaothamatas, line 8: There's a typo. NUM_ADC_CHANNELS

boards/DIM/Src/Io/Io_Adc.c
```diff
@@ -0,0 +1,28 @@
+#include <stm32f3xx.h>
+#include "Io_SharedAdc.h"
+#include "Io_Adc.h"
+
```
pasitlaothamatas, line 4: Should we include a small comment about listing out the elements in the enum in ascending order, exactly like how its specified in Cube?

boards/DIM/Src/Io/Io_Adc.c
pasitlaothamatas: When we specify a channel to a rank in cube we have to specify the channel by its number and not its name. I think channel 12 is okay , imo

boards/DIM/Src/Io/Io_Adc.c
thekenu: What do you mean asecnding order?

boards/DIM/Src/Io/Io_Adc.c
thekenu: Some good concerns were brought up and I added a big block comment clarifying key ADC concepts. Can everyone review it and let me know what you think: bc5a4866db2c4165b430da0e02325ab29763b474. It's meant for someone with limited knowledge about the technical details specific to STM32.

boards/DIM/Src/Io/Io_Adc.c
pasitlaothamatas: Seems like this was addressed in your latest comment

boards/DIM/Src/Io/Io_Adc.c
garethellis0: This seems good to me now, thanks for clarifying.

boards/DIM/Src/Io/Io_Adc.c
```diff
@@ -0,0 +1,49 @@
...
+// The following enum is used to index into raw_adc_values, which means it must
+// be ordered in ascending ranks. If we were writing an enum for the earlier
+// example, it would look like:
+// enum
+// {
+//     CHANNEL_4, // Rank 1
+//     CHANNEL_7, // Rank 2
+//     CHANNEL_2, // Rank 3
+//     NUM_ADC_CHANNELS,
+// }
```
thekenu, line 25:
```suggestion
// example, it would look like:
//
// enum
// {
//     CHANNEL_4, // Rank 1
//     CHANNEL_7, // Rank 2
//     CHANNEL_2, // Rank 3
//     NUM_ADC_CHANNELS,
// };
```

PR #615, opened 2020-05-16
WIP

thekenu:
### Checklist


boards/DIM/Src/App/App_ErrorTable.c
```diff
@@ -0,0 +1,216 @@
+#include <stdlib.h>
```
thekenu, line 1: `{App|Io}_ErrorTable.{c|h}` are meant to be auto-generated and placed inside `cmake-build-{embedded|gtest}`. I hand-wrote it for now as a prototype.

boards/DIM/Src/App/App_ErrorTable.c
```diff
@@ -0,0 +1,216 @@
...
+}
+
+uint32_t App_ErrorTable_HasNonCriticalError(struct ErrorTable* error_table, struct Error errors[])
+{
+    uint32_t num_errors = 0;
+
+    if (error_table->bms_stack_watermark_above_threshold_task1hz)
+    {
+        errors[num_errors].error_id = BMS_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1HZ;
+        errors[num_errors].std_id = 104;
```
thekenu, line 41: `cantools` will allow us to generate the correct std_id

boards/DIM/Inc/App/App_ErrorTable.h
```diff
@@ -0,0 +1,87 @@
...
+ * @param error_table The error table
+ * @param errors Any non-critical errors will be set in this error array
+ * @return The number of non-critical errors
+ */
+uint32_t App_ErrorTable_HasNonCriticalError(struct ErrorTable* error_table, struct Error errors[]);
+
+/**
+ * Check if there is any critical errors in the given error table, and if
+ * necessary, write the critical errors into the given error array
+ * @note The array must hold up to NUM_ERRORS of elements
```
thekenu, line 51: I'm not a big fan of how the user must manually make sure the array is large enough. Might consider generating a struct for the caller to use

```
struct ErrorArray {
  struct Error errors[NUM_ERRORS];
}; 
 uint32_t App_ErrorTable_HasCriticalError(struct ErrorTable* error_table, struct ErrorArray* error_array); 
```

boards/DIM/Src/main.c
```diff
@@ -645,6 +649,7 @@ void RunTaskCanRx(void const *argument)
         Io_SharedCan_DequeueCanRxMessage(&message);
         Io_CanRx_UpdateRxTableWithMessage(
             App_DimWorld_GetCanRx(world), &message);
+        Io_ErrorTable_Update(App_DimWorld_GetErrorTable(world), &message);
```
thekenu, line 652: Wasn't sure if `Io_ErrorTable_Update` should be called inside `Io_CanRx_UpdateRxTableWithMessage`. Left them separate for now.

boards/shared/Src/Io/Io_SharedCan.c
```diff
@@ -297,6 +298,10 @@ void Io_SharedCan_DequeueCanRxMessage(struct CanMsg *message)
         ;
 }
 
+// This extern is a hack for now. Need to rewrite
```
thekenu, line 301: Similar concern as: https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/615/files#r426098243.

There is no nice way to separate Error Table and CAN TX for now though.

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -376,20 +380,43 @@ TEST_F(DimStateMachineTest, imd_led_control_in_drive_state)
...
+    App_CanRx_FSM_NON_CRITICAL_ERRORS_SetSignal_BSPD_FAULT(
+        can_rx_interface, false);
     App_SharedStateMachine_Tick100Hz(state_machine);
     ASSERT_EQ(1, turn_on_bspd_led_fake.call_count);
     ASSERT_EQ(2, turn_off_bspd_led_fake.call_count);
 }
 
+TEST_F(DimStateMachineTest, error_on_7_segment_displays)
+{
+    App_ErrorTable_SetBmsStackWaterMarkAboveThresholdTask1Hz(error_table);
```
thekenu, line 404: I was hoping to replace this line with `App_CanRx_SettBmsStackWaterMarkAboveThresholdTask1Hz`, but since Error Table and Can RX are separate right now I can't do that. Not sure if that's a big deal.

boards/DIM/Src/Io/Io_ErrorTable.c
```diff
@@ -0,0 +1,103 @@
+#include <stdint.h>
+#include <assert.h>
+#include "App_CanRx.h"
+#include "App_CanMsgs.h"
+#include "Io_SharedCanMsg.h"
+#include "Io_ErrorTable.h"
+
+// This should be used in both CAN TX and CAN RX callbacks
+void Io_ErrorTable_Update(
```
thekenu, line 9: If we're generating the code, I was considering generating `Io_ErrorTable_UpdateCanTx` and `Io_ErrorTable_UpdateRx`.

If we're generating code for BMS, the `Io_ErrorTable_UpdateCanTx` would handle only BMS errors and `Io_ErrorTable_UpdateRx` would handle errors from the rest of the boards.

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -174,6 +177,95 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+        for (size_t i = 0; i < num_errors; i++)
+        {
+            uint32_t digits_to_display = errors[i].error_id;
+            assert(digits_to_display < 100);
+
+            switch (errors[i].std_id)
+            {
+                case CANMSGS_BMS_CRITICAL_ERRORS_FRAME_ID:
+                {
+                    digits_to_display += 500;
```
thekenu, line 196: The number after `+=` comes from DIM-9

boards/DIM/Src/App/App_ErrorTable.c
```diff
@@ -0,0 +1,216 @@
...
+
+    return error_table;
+}
+
+void App_ErrorTable_Destroy(struct ErrorTable* error_table)
+{
+    free(error_table);
+}
+
+uint32_t App_ErrorTable_HasNonCriticalError(struct ErrorTable* error_table, struct Error errors[])
```
garethellis0, line 34: Um. I'm a tad confused. This is a function in the `DIM` board (it should be shared?) and it also uses the stack watermark from the BMS for some reason?

thekenu: It was just placed here as an example so I can check that the code compiles. And the stack watermark is just one of the example non-critical error in the error table.

boards/DIM/Inc/App/App_ErrorTable.h
```diff
@@ -0,0 +1,87 @@
...
+
+    NUM_ERRORS = 10, // Number of non-critical and critical errors combined
+                     // across ALL boards
+};
+
+struct Error
+{
+    // CAN message ID
+    uint32_t std_id;
+    // Each error ID is unique in the error table
```
garethellis0, line 31: Do you mean in the `CAN` message? If the error ID is globally unique (ie. across all CAN messages) then there's no point to having the `std_id` here as well?

boards/DIM/Inc/App/App_ErrorTable.h
garethellis0: Yeah, that's fine by me.

boards/DIM/Inc/App/App_ErrorTable.h
```diff
@@ -0,0 +1,87 @@
...
+ * Check if there is any critical errors in the given error table, and if
+ * necessary, write the critical errors into the given error array
+ * @note The array must hold up to NUM_ERRORS of elements
+ * @param error_table The error table
+ * @param errors Any critical errors will be set in this error array
+ * @return The number of critical errors
+ */
+uint32_t App_ErrorTable_HasCriticalError(struct ErrorTable* error_table, struct Error errors[]);
+
+void App_ErrorTable_SetBmsStackWaterMarkAboveThresholdTask1Hz(struct ErrorTable* error_table);
```
garethellis0, line 58: I'm confused why we're setting errors like this, but then define errors using the struct above? I originally proposed using the struct above so that we wouldn't need to have semantic understanding of what each error is.

boards/DIM/Src/App/App_ErrorTable.c
garethellis0: Why are there per-board error tables but also the shared one?

thekenu: The intention was to have only one error table shared across the whole car. I think it became confusing when i decided to put the code in `DIM` rather than in `shared`. Just to be clear, I _don't_ want a per-baord error table anywhere.

boards/DIM/Inc/App/App_ErrorTable.h
thekenu: Good catch. This comment is wrong. The error IDs will be unique only on a per-board basis to fulfill:
image

boards/DIM/Inc/App/App_ErrorTable.h
thekenu: Did you mean `struct Error`? I made these setters becuase I thought it's what you meant by

>// Setters for each possible error message ... we would determine these by regex match on the message ID's from the dbc

I must have misinterpreted what you meant? Maybe `Io_ErrorTable_Update` is what you wanted me to write?

boards/DIM/Inc/App/App_ErrorTable.h
thekenu: Initially I thought about making the error IDs globally unique, but considering the fact that we only have 2 digits to display errors that would leave us with 99 errors in total across 5 PCBs. Which seemed too restrictive.

thekenu: The per-board unique IDs are reflected in the enum at the top of this file

thekenu:
I _think_ I know what you mean by raw CAN message level. If we go down that route, how do we figure out where each error came from and how would a user look up an error's semantic meaning given an error ID?

Also, I moved things under `shared/` and added `Shared` into the function names to indicate that things are meant to be shareable across boards.

PR #617, opened 2020-05-16
Prepare current sense and voltage sense for BMS

thekenu:
### Summary

- Configure CubeMX settings for 4 x ADC channels in DMA mode: Main current 1, Main current 2, AIR shutdown loop current, and TS voltage. They are all triggered by TIM3.
- Write code to convert ADC voltage into meaningful values (e.g. ADC voltage -> TS voltage)
- To support writing uint tests for Io code, I had to update the CMakeLists.txt to divide source files based on whether they are compilable on x86 or not (Rather than just App vs Io).

### Testing Done

- Wrote unit tests

### Checklist


boards/BMS/BMS.ioc
```diff
@@ -1,8 +1,8 @@
 #MicroXplorer Configuration settings - do not modify
```
thekenu, line 1: Unfortunately the BMS board is stuck in the EDC so I can't check that these settings on hardware. But this is more or less the same as #614 so I have reasonable confidence in its correctness.

boards/BMS/Inc/App/App_CurrentSense.h
```diff
@@ -0,0 +1,36 @@
...
+#include "App_ErrorCode.h"
+
+/**
+ * Convert the given ADC voltage to main current 1
+ * @param adc_voltage The ADC voltage to convert
+ * @param main_current_1 This will be set to main current 1, in amps
+ * @return EXIT_CODE_OUT_OF_RANGE if adc_voltage is negative
+ *         EXIT_CODE_INVALID_ARGS if main_current_1 is NULL
+ */
+ExitCode App_CurrentSense_ConvertToMainCurrent1(
```
thekenu, line 12: Originally I wanted `App_{Current|Voltage}Sense.{c|h}` to be in the IO layer because it's code for the current sensors. But I think it's better to leave as much as possible in the App layer so we can write unit tests for them.

thekenu: Also didn't put any of these functions in `BmsWorld` because it's tough to know where they belong right now. So just going to leave them as standalone helper functions for now.

boards/BMS/Inc/App/App_CurrentSense.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "App_ErrorCode.h"
+
+/**
+ * Convert the given ADC voltage to main current 1
+ * @param adc_voltage The ADC voltage to convert
```
garethellis0, line 7: units?

boards/BMS/Inc/App/App_CurrentSense.h
```diff
@@ -0,0 +1,36 @@
...
+ * @param main_current_1 This will be set to main current 1, in amps
+ * @return EXIT_CODE_OUT_OF_RANGE if adc_voltage is negative
+ *         EXIT_CODE_INVALID_ARGS if main_current_1 is NULL
+ */
+ExitCode App_CurrentSense_ConvertToMainCurrent1(
+    float  adc_voltage,
+    float *main_current_1);
```
garethellis0, line 14: I'm a bit confused about what MainCurrent1 is? Is this dictacted by hardware?

thekenu: They're names carried over from the hardware schematic. Basically we call the current in the high voltage path (i.e. 400V battery pack) "Main current". The current sensor for this has two outputs (for two different resolutions), hence, Main Current 1 and Main Current 2.

image

thekenu: The names aren't great. But since I'm not using these currents in the state machine anywhere yet it's hard to come up with a better name. So I thought i'd just leave it the same as the schematic for now.

garethellis0: If the difference between the two is the different resolution, perhaps we could make that part of the name?

boards/BMS/Src/App/App_CurrentSense.c
```diff
@@ -0,0 +1,104 @@
...
+    // Sensitivity = 40mV/A
+
+    const float offset             = 2.5f;
+    const float sensitivity        = 40e-3f;
+    const float hsnbv_d06_output_1 = adc_voltage * (1e3f + 2.2e3f) / 2.2e3f;
+
+    *main_current_1 = (hsnbv_d06_output_1 - offset) / sensitivity;
+
+    return EXIT_CODE_OK;
+}
```
garethellis0, line 42: Just reading through this, it really seems like a very hardware-specific thing. We can still test IO code if we explicitly include it in the gtest, so I don't see a particularly strong reason this shouldn't be in IO? I suppose the idea here is that if we change things like resistor values, we really shouldn't need to change `app` code.

garethellis0: (Sorry for this being a belated comment, was tired when I did the first pass and just took another look at this after a full night of sleep)

boards/BMS/Inc/App/App_CurrentSense.h
thekenu: Yup that's fine with me

boards/BMS/Src/App/App_CurrentSense.c
thekenu: How can we add IO code to gtest since they include the STM32 headers? I wasn't aware testing IO code is an option.

garethellis0: Oh they do? I thought in this case we would just not include any stm32 headers in that particular Io `.c/.h`? Then we could just include it in the files we build into the gtest target.

thekenu: Not all, but most Io code does include stm32 headers.

Did you mean we will just hard code exactly which Io .c/.h should be added to `gtest`?

garethellis0: > Also do you mean we should just hard code exactly which Io .c/.h to add to gtest?

Yeah, that was basically what I was thinking.  Looking at this I think it would require one of:

1. adding an additional argument to `create_arm_binary_or_tests_for_board`
1. just putting the IO files in `GOOGLETEST_TEST_SRCS` 
1. just splitting the files we pass to `create_arm_binary_or_tests_for_board` based on whether or not we can compile them for x86 or not.

Personally I'm a fan of (3), since it means we aren't adding more arguments (as 1 does) and it doesn't feel like a bit of a hack (like 2 does).

thekenu: Do you mind writing some pseudo-code for option 3? Did you mean replacing the following arguments
```
    "${ARM_BINARY_APP_SRCS}"
    "${ARM_BINARY_APP_INCLUDE_DIRS}"
    "${ARM_BINARY_IO_SRCS}"
    "${ARM_BINARY_IO_INCLUDE_DIRS}"
```
with
```
    "${x86_COMPATIBLE_SRCS}" // This can be App or Io
    "${x86_COMPATIBLE_INCLUDE_DIRS}"  // This can be App or Io
    "${ARM_BINARY_SRCS}"  // This must be Io
    "${ARM_BINARY_INCLUDE_DIRS}" // This must be Io
```

garethellis0: Yep basically

thekenu:
@garethellis0 Finally got around to cleaning this up. I have divided up the files into x86 and non-x86 groups like we disussed. Ready for review again :+1:

boards/BMS/CMakeLists.txt
```diff
@@ -37,17 +37,52 @@ list(REMOVE_ITEM GOOGLETEST_TEST_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/main
...
+# compiled on x86. The rest of the Io files, however, can be compiled on x86.
+#
+# These relationships can be summarized as follows:
+#
+# ARM_BINARY_X86_COMPATIBLE_SRCS = App source files + x86-compatible Io files
+# ARM_BINARY_X86_INCOMPATIBLE_SRCS = Io sources files - x86-compatible Io files
+#
+# On the other hand, the App and Io include directories are always
+# x86-compatible because they are paths and not actual files. We just have to
+# place App and Io include directories in ARM_BINARY_INCLUDE_DIRS.
```
garethellis0, line 56: This really feels like we should have it in `boards/CMakeLists.txt`, rather then duplicated in the per-board `CMakeLists.txt`.

thekenu: Do you mean just the comments?

thekenu: I'm not super set on leaving the comments on a per-board basis though. Just wanted to let you know what my thinking process was first.

garethellis0: Yeah, I think in this case the duplication is significant enough that moving it to `boards/CMakeLists.txt` is worth it, even if it might not be totally correct in terms of keeping things isolated.

thekenu: Added it as a `@note` under `create_arm_binary_or_tests_for_board`

PR #618, opened 2020-05-17
Remove alloc_index from Io code with malloc

pasitlaothamatas:
### Summary

After discussing with @thekenu we have decided to use malloc in `Io_x_Create` functions instead of statically allocating memory with `alloc_index` (not including `Io_SharedSoftwareWatchdog`)

### Changelist 

 - Changed from static memory allocation to malloc in `Io_SharedPwmInput_Create.c` and `Io_FreqOnlySharedPwmInput.c`
- Removed `Io_SharedPwmInputConfig.h` and `Io_SharedFreqOnlyPwmInputConfig.h` for all boards.

### Checklist

### DIFF
--- a/boards/shared/Src/Io/Io_SharedFreqOnlyPwmInput.c
+++ b/boards/shared/Src/Io/Io_SharedFreqOnlyPwmInput.c
@@ -49,6 +49,7 @@ struct FreqOnlyPwmInput *Io_SharedFreqOnlyPwmInput_Create(
 
     struct FreqOnlyPwmInput *const pwm_input =
         (struct FreqOnlyPwmInput *)malloc(sizeof(struct FreqOnlyPwmInput));
+    assert(pwm_input != NULL);
 
     pwm_input->frequency_hz        = NAN;
     pwm_input->htim                = htim;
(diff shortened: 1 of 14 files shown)


boards/shared/Src/Io/Io_SharedFreqOnlyPwmInput.c
```diff
@@ -46,12 +46,9 @@ struct FreqOnlyPwmInput *Io_SharedFreqOnlyPwmInput_Create(
     HAL_TIM_ActiveChannel tim_active_channel)
 {
     assert(htim != NULL);
-    static struct FreqOnlyPwmInput pwm_inputs[MAX_NUM_OF_FREQ_ONLY_PWM_INPUTS];
-    static size_t                  alloc_index = 0;
 
-    assert(alloc_index < MAX_NUM_OF_FREQ_ONLY_PWM_INPUTS);
-
-    struct FreqOnlyPwmInput *const pwm_input = &pwm_inputs[alloc_index++];
+    struct FreqOnlyPwmInput *const pwm_input =
```
thekenu, line 50: Missing assert on null pointers after malloc in this PR

pasitlaothamatas: should be good now

boards/shared/Src/Io/Io_SharedFreqOnlyPwmInput.c
```diff
@@ -46,12 +46,10 @@ struct FreqOnlyPwmInput *Io_SharedFreqOnlyPwmInput_Create(
...
+        (struct FreqOnlyPwmInput *)malloc(sizeof(struct FreqOnlyPwmInput));
```
thekenu, line 51: Honestly in C you don't really need to cast malloc https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc/605858#605858. Up to you.

PR #621, opened 2020-05-19
Add error table

thekenu:
### Checklist


boards/shared/Inc/App/App_SharedError.h
```diff
@@ -0,0 +1,154 @@
...
+{
+    BMS,
+    DCM,
+    DIM,
+    FSM,
+    PDM,
+    NUM_ERROR_BOARDS,
+};
+
+struct ErrorBoardList
```
thekenu, line 20: So instead of the original design where the function stores a static array of pointers (e.g. `void ErrorTable_getBoardsWithCriticalErrors(ErrorTable* error_table, ErrorBoard[] boards, int* num_boards), I have the caller provide storage for the pointers. 

I personally prefer this approach a little more. Let me know if you feel strongly about switching about to the static array approach.

boards/shared/Inc/App/App_SharedError.h
```diff
@@ -0,0 +1,154 @@
...
+    // Number of valid entries in "boards"
+    uint32_t num_boards;
+
+    // Storage large enough to store every board
+    enum ErrorBoard boards[NUM_ERROR_BOARDS];
+};
+
+enum ErrorId
+{
+    BMS_NON_CRIT,
```
thekenu, line 31: For now, I still stuck with making up just enough fake errors for writing unit tests. I will add in real errors from the DBC later.

boards/shared/Inc/App/App_SharedError.h
```diff
@@ -0,0 +1,154 @@
+#pragma once
```
thekenu, line 1: Another small note is that I split things up in two files (`SharedError` and `SharedErrorTable` because I felt like it made things much easier to reason about.

boards/shared/Inc/App/App_SharedError.h
```diff
@@ -0,0 +1,139 @@
...
+
+enum ErrorBoard
+{
+    BMS,
+    DCM,
+    DIM,
+    FSM,
+    PDM,
+    NUM_ERROR_BOARDS,
+};
```
garethellis0, line 18: Perhaps it's worth creating `enum Board` in a separate file so that we can share this with other code in the future?

boards/shared/Inc/App/App_SharedError.h
garethellis0: Nah, I like this as well. :+1:

boards/shared/Src/App/App_SharedErrorTable.c
```diff
@@ -0,0 +1,293 @@
...
+
+static void App_ResetBoardList(struct ErrorBoardList *board_list)
+{
+    for (size_t i = 0; i < NUM_ERROR_BOARDS; i++)
+    {
+        board_list->boards[i] = NUM_ERROR_BOARDS;
+    }
+
+    board_list->num_boards = 0;
+}
```
garethellis0, line 19: I'm a bit confused, why do we need this? Is it mostly just something so the lists look a little nicer for debugging?

boards/shared/Src/App/App_SharedErrorTable.c
```diff
@@ -0,0 +1,293 @@
...
+    {
+        return EXIT_CODE_OUT_OF_RANGE;
+    }
+    struct Error *error = error_table->errors[error_id];
+    App_SharedError_SetIsSet(error, is_set);
+
+    return EXIT_CODE_OK;
+}
+
+bool App_SharedErrorTable_HasError(const struct ErrorTable *error_table)
```
garethellis0, line 99: Just to be super clear, perhaps we could call this `HasAnyErrorSet` ?

boards/shared/Src/Io/Io_SharedErrorTable.c
```diff
@@ -0,0 +1,33 @@
+#include "App_SharedErrorTable.h"
+#include "App_CanMsgs.h"
+#include "Io_SharedErrorTable.h"
+
+static void Io_ProcessFsmCriticalErrorMsg(
+    struct ErrorTable *          error_table,
+    struct CanMsgs_fsm_errors_t *data)
+{
+    // TODO: Fill this out once I'm sure about other parts of the design
```
garethellis0, line 9: TODO? This design seems fairly mature to me now.

boards/shared/Src/Io/Io_SharedErrorTable.c
```diff
@@ -0,0 +1,33 @@
...
+    (void)error_table;
+    (void)data;
+}
+
+void Io_SharedErrorTable_SetErrorsBasedOnCanMsg(
```
garethellis0, line 14: as a suggestion: `Io_SharedErrorTable_SetErrorsBasedOnCanMsg` -> `Io_SharedErrorTable_SetErrorsFromCanMsg`

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+    // The board exists in the list but is out-of-bound
+    board_list.boards[0]  = BMS;
+    board_list.boards[1]  = DCM;
+    board_list.boards[2]  = DIM;
+    board_list.boards[3]  = FSM;
+    board_list.boards[4]  = PDM;
+    board_list.num_boards = 4;
+    ASSERT_FALSE(App_SharedError_IsBoardInList(&board_list, PDM));
+}
```
garethellis0, line 157: bit of a nit; but each case here should really be a separate test case. There are several other similar cases below where there are multiple clearly distinct things being tested in a single test case (they're even already effectively named with a comment!).

boards/shared/Inc/App/App_SharedError.h
thekenu: Yeah I was thinking the same too, I will create one under `shared`

boards/shared/Test/Src/Test_ErrorTable.cpp
thekenu: Yeah I kinda felt similarly when I was writing these, I'll spend some time to break things down

boards/shared/Src/App/App_SharedErrorTable.c
thekenu: Yeah I thought that could be nice but on a second though maybe it's no necessary. I removed this function altogether and replaced it with just `board_list->num_boards`.

boards/shared/Src/App/App_SharedErrorTable.c
thekenu: Updated `HasCriticalError` and `HasNonCriticalError` too.

boards/shared/Inc/App/App_SharedError.h
thekenu: Fixed

boards/shared/Src/Io/Io_SharedErrorTable.c
thekenu: Fixed

boards/shared/Src/Io/Io_SharedErrorTable.c
thekenu: Fixed, added for all boards

garethellis0: Is it added for all boards? I don't see it.

thekenu: Refresh one more time? It's there i just double checked

thekenu:
Quick update: I fixed everything _except_ for the test files. I'll send another follow-up message once that's done.

thekenu:
@garethellis0 Ready for review again. Just a heads up, I had to split each `BOARD_ERRORS` message into `BOARD_NON_CRITICAL_ERRORS` and `BOARD_CRITICAL_ERRORS` so there are changes unrelated to the error table files.

boards/shared/Src/Io/Io_SharedErrorTable.c
garethellis0: What files is this in? I see no calls to this function anywhere.

thekenu: Oh I thought I misunderstood what you were asking. I thought you meant if I wrote all the `Io_Process{Board}NonCriticalErrorMsg` function definitions. 

I wanted to call the function in another PR where I add in the callback functions for CAN TX/RX interface. This PR will just be focused on building the error table itself. How does that sound?

garethellis0: Oh that's totally fine, just making sure we're on the same page.

PR #624, opened 2020-05-19
Add basic charger interface

thekenu:
### Summary

Add basic charger interface

### Testing Done

Wrote integration test for test machine sending out charger status

### Checklist


boards/BMS/Inc/App/App_Charger.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Deallocate the memory used by the given charger
+ * @param charger The charger to deallocate
+ */
+void App_Charger_Destroy(struct Charger *charger);
+
+/**
+ * Enable the given charger
+ * @param charger The charger to enable
```
garethellis0, line 28: What does `enable` mean here? Similairly for `disable`?

thekenu: The charger has an on/off pin that must be enabled before charging can begin. It's mentioned a few times in the specs:

>BMS-17 | Charging thermal safety | - The BMS must stop cell balancing once the LTC6813 internal di    e temperature (ITMP) exceeds 115C and throw a non-critical fault. The BMS must re-enable cell bala    ncing once the ITMP decreases below 110C. <br/> - The BMS must disable the charger once the ITMP e    xceeds 120C and throw a non-critical fault. The BMS must re-enable the charger once the ITMP decre    ases below 115C. <br/>  - The BMS must disable the charger when any cell temperature exceeds 43C a    nd throw a non-critical fault.

> BMS-23 | Exiting the charge state and entering the init state | Once charging is complete, the BMS     must disable the charger, disable cell balancing, open the contactors and enter the init state.
etc

garethellis0: perhaps we could suffix `enable` and `disable` with `CellBalancing` to be more clear?

thekenu: Oh it's not actually enable or disabling `CellBalancing`. It's more like a master-switch for the charger.

image

I can see how it's confusing because not every charger would necessarily have a master-switch like this. What do you think we can do here?

garethellis0: After looking at this a bit, I think it's fine as-is.

garethellis0: Can we just somehow put that description (from the image above) in a header-wide comment?

thekenu: Yup will do

PR #628, opened 2020-05-22
Created a steering angle sensor driver

pasitlaothamatas:
### Summary

Created a driver for the steering angle sensor

### Changelist 

- Add Io code to calculate the steering angle sensor
- Configured ADC2 to read analog input value from the steering angle sensor
- Configured TIM3 to trigger ADC conversions at a frequency of 1kHz

### Testing Done

- Add simple unit test to check steering angle can signals in all states

### Checklist


boards/FSM/Test/Src/Test_StateMachine.cpp
```diff
@@ -11,6 +11,7 @@ extern "C"
 #include "configs/App_HeartbeatMonitorConfig.h"
 #include "configs/App_FlowRateThresholds.h"
 #include "configs/App_WheelSpeedThresholds.h"
+#include "configs/App_SteeringAngleConstants.h"
```
thekenu, line 14: Can we name it App_SteeringAngleThresholds to be more consistent?

boards/FSM/Src/App/App_FsmWorld.c
```diff
@@ -68,25 +75,31 @@ struct HeartbeatMonitor *
 struct InRangeCheck *App_FsmWorld_GetPrimaryFlowRateInRangeCheck(
     const struct FsmWorld *const world)
 {
-    return world->primary_flow_rate_in_range_check;
+    return world->primary_flow_meter_in_range_check;
```
thekenu, line 78: Why are the flow meter and wheel speed in range checks renamed again? I thought they were fine as-is.

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
```diff
@@ -0,0 +1,17 @@
+#include "Io_Adc.h"
+#include "configs/App_SteeringAngleConstants.h"
+#include "Io_SteeringAngleSensor.h"
+
+float Io_SteeringAngleSensor_GetAngleDeg(void)
```
thekenu, line 5:
```suggestion
float Io_SteeringAngleSensor_GetAngleDegree(void)
```

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
```diff
@@ -0,0 +1,17 @@
...
+    // Subtract the measured voltage by the steering angle sensor voltage
+    // offset. Multiply the result with VOLTAGE_TO_DEGREES to calculate the
+    // steering angle in degrees.
+    //
+    // The scale factor used to convert the voltage to an angle is given by 360
+    // deg / VDDA (3.3V)
+
+    const float VOLTAGE_TO_DEGREES = 360.0f / 3.3f;
+    return VOLTAGE_TO_DEGREES *
+           (Io_Adc_GetChannel1Voltage() - STEERING_ANGLE_VOLTAGE_OFFSET);
```
thekenu, line 16: `STEERING_ANGLE_VOLTAGE_OFFSET` should just be a const in this function. It does not belong in App.

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
```diff
@@ -0,0 +1,17 @@
...
+float Io_SteeringAngleSensor_GetAngleDeg(void)
+{
+    // Subtract the measured voltage by the steering angle sensor voltage
+    // offset. Multiply the result with VOLTAGE_TO_DEGREES to calculate the
+    // steering angle in degrees.
+    //
+    // The scale factor used to convert the voltage to an angle is given by 360
+    // deg / VDDA (3.3V)
+
+    const float VOLTAGE_TO_DEGREES = 360.0f / 3.3f;
```
thekenu, line 14:
```suggestion
    const float DEGREE_PER_VOLT = 360.0f / 3.3f;
```
This would be more clear?

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
```diff
@@ -0,0 +1,17 @@
...
+#include "configs/App_SteeringAngleConstants.h"
+#include "Io_SteeringAngleSensor.h"
+
+float Io_SteeringAngleSensor_GetAngleDeg(void)
+{
+    // Subtract the measured voltage by the steering angle sensor voltage
+    // offset. Multiply the result with VOLTAGE_TO_DEGREES to calculate the
+    // steering angle in degrees.
+    //
+    // The scale factor used to convert the voltage to an angle is given by 360
```
thekenu, line 11: Where did this scale factor come from?

boards/FSM/Test/Src/Test_StateMachine.cpp
pasitlaothamatas: I named it "constants" as the file also included the offset voltage for the sensor, which isn't a threshold

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
pasitlaothamatas: Ah okay. Ignore my comment on https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/628#discussion_r429518191

boards/FSM/Src/App/App_FsmWorld.c
pasitlaothamatas: Oh I changed the name for consistency as in the .h file it was named `primary_flow_meter_in_range_check` but now I'm realizing that I should've the name in the .h file instead

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
pasitlaothamatas: The scale factor is referring to `VOLTAGE_TO_DEGREES`. Should I be more specific in the javadoc

boards/FSM/Test/Src/Test_StateMachine.cpp
pasitlaothamatas: ^ignore this comment pls

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
pasitlaothamatas: Alright this is fixed

boards/FSM/Src/App/App_FsmWorld.c
pasitlaothamatas: This should be fixed now

pasitlaothamatas:
ready for another round @thekenu

boards/FSM/Src/Io/Io_SteeringAngleSensor.c
thekenu: I think you should just write it as an equation to avoid any misleading comments. Scale factor isn't really a standard terminology and isn't as specific as could be. The current comments is really just describing the formula but in a more verbose way.


```        
// Steering Angle =  (ADC Voltage - Voltage Offset) * Degree Per Volt

const float VOLTAGE_OFFSET = 1.9f;
const float DEGREE_PER_VOLT = 360.9 / 3.3f;

return  (Io_Adc_GetChannel1Voltage() - VOLTAGE_OFFSET ) *  DEGREE_PER_VOLT ;
```

thekenu: `DEGREE_PER_VOLT` seems a bit more clear than `VOLTAGE_TO_DEGREES` on a second though

pasitlaothamatas: Ok should be fixed now

boards/FSM/Inc/Io/Io_SteeringAngleSensor.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+/**
+ * Gets the steering angle in degrees from the steering angle sensor
```
thekenu, line 4: nit
```suggestion
 * Get the steering angle in degrees from the steering angle sensor
```

PR #630, opened 2020-05-23
Send zero torque in DCM fault state

thekenu:
### Summary

image

### Testing Done

Unit test

### Checklist

### DIFF
--- a/boards/DCM/Test/Src/Test_StateMachine.cpp
+++ b/boards/DCM/Test/Src/Test_StateMachine.cpp
@@ -251,8 +251,6 @@ TEST_F(DcmStateMachineTest, zero_torque_request_in_fault_state)
 
     // Start with a non-zero torque request to prevent false positive
     App_CanTx_SetPeriodicSignal_TORQUE_REQUEST(can_tx_interface, 1.0f);
-    ASSERT_EQ(
-        1.0f, App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx_interface));
 
     // Now tick the state machine and check torque request gets zeroed
     App_SharedStateMachine_Tick100Hz(state_machine);
--- a/boards/DCM/Test/Src/Test_StateMachine.cpp
+++ b/boards/DCM/Test/Src/Test_StateMachine.cpp
@@ -244,4 +244,20 @@ TEST_F(DcmStateMachineTest, zero_torque_request_in_init_state)
         0.0f, App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx_interface));
 }
 
+// DCM-16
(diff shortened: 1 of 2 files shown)


boards/DCM/Test/Src/Test_StateMachine.cpp
```diff
@@ -244,4 +244,20 @@ TEST_F(DcmStateMachineTest, zero_torque_request_in_init_state)
...
 
+// DCM-16
+TEST_F(DcmStateMachineTest, zero_torque_request_in_fault_state)
+{
+    SetInitialState(App_GetFaultState());
+
+    // Start with a non-zero torque request to prevent false positive
+    App_CanTx_SetPeriodicSignal_TORQUE_REQUEST(can_tx_interface, 1.0f);
+    ASSERT_EQ(
+        1.0f, App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx_interface));
```
garethellis0, line 255: I don't think we need this `ASSERT_EQ` here, because we should already have tested this in the `CanTx` tests. Restricting asserts to just what we're focused on testing is actually a great way to help guide people as to the intent of the test.

thekenu: But if the default value for torque request is `0`, wouldn't the `CanTx` just check that `App_CanTx_Create` sets the torque request to `0`?   And since we want zero torque request in the fault state that would cause a false positive?

thekenu: Like if I remove this `ASSERT_EQ`, this test would pass even if I remove the code I added to `App_FaultState.c`

garethellis0: Right, so we're checking the functionality of the `CanTx` here yeah? That should be a separate test.

thekenu: Ohhhhh I see and you mean now I thought you meant I should remove line 253 and 254. I gotcha now.

boards/DCM/Test/Src/Test_StateMachine.cpp
```diff
@@ -244,4 +244,20 @@ TEST_F(DcmStateMachineTest, zero_torque_request_in_init_state)
...
+
+    // Now tick the state machine and check torque request gets zeroed
```
thekenu, line 257:
```suggestion

    // Now tick the state machine and check torque request gets zeroed
```

boards/DCM/Test/Src/Test_StateMachine.cpp
thekenu: Fixed!

PR #632, opened 2020-05-26
Enable ctest memcheck

thekenu:
### Summary

Ran Valgrind and realized there was a handful of memory leaks, one of which was especially bad because I malloc'd the wrong size for ErrorTable. Turns out cmake has a `-T memcheck` option so I decided to add it into CI.

### Checklist

### DIFF
--- a/boards/BMS/Test/Inc/Test_Imd.h
+++ b/boards/BMS/Test/Inc/Test_Imd.h
@@ -15,7 +15,7 @@ class ImdTest : public testing::Test
         enum Imd_ConditionName condition_name,
         float &                fake_pwm_frequency_return_val);
     static void
-        SetPwmFrequencyTolerance(struct Imd *&imd_to_set, float tolerance);
+        SetPwmFrequencyTolerance(struct Imd *imd_to_set, float tolerance);
 
   protected:
     void SetUp() override;
--- a/boards/BMS/Test/Inc/Test_Imd.h
+++ b/boards/BMS/Test/Inc/Test_Imd.h
@@ -15,7 +15,7 @@ class ImdTest : public testing::Test
         enum Imd_ConditionName condition_name,
         float &                fake_pwm_frequency_return_val);
     static void
-        SetPwmFrequencyTolerance(struct Imd *imd_to_set, float tolerance);
+        SetPwmFrequencyTolerance(struct Imd *&imd_to_set, float tolerance);
 
   protected:
     void SetUp() override;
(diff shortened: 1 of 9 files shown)


boards/BMS/Test/Inc/Test_Imd.h
```diff
@@ -15,7 +15,7 @@ class ImdTest : public testing::Test
         enum Imd_ConditionName condition_name,
         float &                fake_pwm_frequency_return_val);
     static void
-        SetPwmFrequencyTolerance(struct Imd *imd_to_set, float tolerance);
+        SetPwmFrequencyTolerance(struct Imd *&imd_to_set, float tolerance);
```
garethellis0, line 18: Ok, ya got me, why are we changing this? :thinking:

boards/CMakeLists.txt
```diff
@@ -469,6 +469,14 @@ endfunction()
...
 if ("${PLATFORM}" STREQUAL "x86")
     download_and_unpack_google_test(${CMAKE_SOURCE_DIR}/CMakeLists.txt.in)
+    # For ctest to return a non-zero code when Valgrind fails, we must use
+    # --error-exitcode. On the other hand, --leak-check=full ensures that any
+    # memory leak issue will cause ctest to return a non-zero code.
+    #
+    # This must be set before include(CTest) is called, or else
+    # MEMORYCHECK_COMMAND_OPTIONS won't be overwritten correctly!
+    set(MEMORYCHECK_COMMAND_OPTIONS "--error-exitcode=1 --leak-check=full")
+    include (CTest)
```
garethellis0, line 479: Nice!

boards/PDM/Test/Src/Test_InRangeCheck.cpp
```diff
@@ -79,6 +79,13 @@ TEST_F(InRangeCheckTest, value_overflow)
 
 TEST_F(InRangeCheckTest, invalid_min_and_max)
 {
+#ifdef NDEBUG
+    // Using ASSERT_DEBUG_DEATH with a function that has malloc in it (i.e.
+    // App_InRangeCheck_Create) causes Valgrind to think there's memory leak in
+    // Release Mode but not Debug Mode. To get around this, we're just going to
+    // disable this test in Release Mode for the time being until we find a
+    // better workaround.
```
garethellis0, line 87: :+1:

boards/BMS/Test/Inc/Test_Imd.h
thekenu: I was a bit confused at first too. So the problem is that the new pointer returned by `_Create` in this function is stored in the local variable `imd_to_set` which gets lost after this function returns. 
```
void ImdTest::SetPwmFrequencyTolerance(struct Imd *&imd_to_set, float tolerance)
{
    TearDownObject(imd_to_set, App_Imd_Destroy);
    imd_to_set = App_Imd_Create(
        get_pwm_frequency, tolerance, get_pwm_duty_cycle,
        get_seconds_since_power_on);
}
```

PR #636, opened 2020-05-27
Add logic for handling DCM state transition using start switch position

thekenu:
### Summary

Add DCM-14

### Checklist


thekenu:
Yeah that's a good question. I'm actually not too sure. Thoughts @kirons96 ?

kirons96:
right now, the initial state is init, and to enter the drive state from the init state:

`The start switch must be switched in an upwards direction. If the start switch was already in the upwards position, it must be re-switched into the upwards position.`

to get to the drive state, the start switch will definitely be in the upwards position. to be on the safe side we might as well transition to the init state whenever the switch is in the downwards position. we can reword the spec as:

`When the start switch is in a downwards position, the DCM must transition from the drive state to the init state.`

or something like that

thekenu:
Yeah I think the new wording will be less ambiguous

PR #637, opened 2020-05-27
Add voting algorithm for SoCs

thekenu:
### Summary

The SoC is stored in EEPROM which can get corrupted. As a workaround, the same SoC value is stored in 3 separate EEPROM locations. If one EEPROM value is corrupted (e.g. due to power failure), the other two may still be valid.

```
t = 0
SoC 1 = 1.0% 
SoC 2 = 1.0% 
SoC 3 = 1.0% 

t = 1 // Power failure prevents 1.5% to be written on SoC 2 and 3
SoC 1 = 1.5% 
SoC 2 = 1.0% 
SoC 3 = 1.0% 
```
There are many other ways that data can get corrupted. To be as safe as possible, we look for the two values out of the three that are the closest.
 
### Testing Done

Wrote unit tests

### Checklist

### DIFF
--- a/boards/BMS/Inc/App/App_Soc.h
+++ b/boards/BMS/Inc/App/App_Soc.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include "App_SharedExitCode.h"
+
+/**
+ * Given three state-of-charge (SoCs), check whether the difference between any
+ * two SoCs is less than or equal to the specified threshold. If any two SoCs
+ * pass the check, return the average value of those two SoCs.
+ * @param max_difference The maximum tolerable difference between any two given
+ *                       state-of-charge values
+ * @param soc_1 The first SoC
+ * @param soc_2 The second SoC
+ * @param soc_3 The third SoC
+ * @param result The average value of two SoCs whose difference is less than or
+ *               equal to max_value, else NaN. In the event of a tiebreak (i.e.
+ *               All three SoCs pass the check), th first two SoCs are used to
+ *               calculate the average value.
+ * @return EXIT_CODE_INVALID_ARGS if the max difference or any of the SoCs is
+ *                                less than 0 or greater than 100
+ */
+ExitCode App_Soc_Vote(
+    float  max_difference,
+    float  soc_1,
+    float  soc_2,
+    float  soc_3,
+    float *result);
--- a/boards/BMS/Inc/App/App_Soc.h
+++ b/boards/BMS/Inc/App/App_Soc.h
@@ -3,23 +3,43 @@
 #include "App_SharedExitCode.h"
 
 /**
- * Given three state-of-charge (SoCs), check whether the difference between any
- * two SoCs is less than or equal to the specified threshold. If any two SoCs
- * pass the check, return the average value of those two SoCs.
- * @param max_difference The maximum tolerable difference between any two given
- *                       state-of-charge values
- * @param soc_1 The first SoC
- * @param soc_2 The second SoC
- * @param soc_3 The third SoC
- * @param result The average value of two SoCs whose difference is less than or
- *               equal to max_value, else NaN. In the event of a tiebreak (i.e.
- *               All three SoCs pass the check), th first two SoCs are used to
- *               calculate the average value.
- * @return EXIT_CODE_INVALID_ARGS if the max difference or any of the SoCs is
- *                                less than 0 or greater than 100
+ * Given three state-of-charge (SoCs), check whether the absolute difference
+ * between any two SoCs is less than or equal to the specified maximum absolute
+ * difference. If any two SoCs pass the check, return the average value of those
+ * two SoCs.
+ * @param max_abs_difference The maximum allowable absolute difference between
+ *                           any two given SoCs, must be between 0 and 100
+ *                           inclusive
+ * @param soc_1 The first SoC, must be between 0 and 100 inclusive
+ * @param soc_2 The second SoC, must be between 0 and 100 inclusive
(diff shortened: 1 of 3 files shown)


boards/BMS/Inc/App/App_Soc.h
```diff
@@ -0,0 +1,26 @@
...
+ *                       state-of-charge values
+ * @param soc_1 The first SoC
+ * @param soc_2 The second SoC
+ * @param soc_3 The third SoC
+ * @param result The average value of two SoCs whose difference is less than or
+ *               equal to max_value, else NaN. In the event of a tiebreak (i.e.
+ *               All three SoCs pass the check), th first two SoCs are used to
+ *               calculate the average value.
+ * @return EXIT_CODE_INVALID_ARGS if the max difference or any of the SoCs is
+ *                                less than 0 or greater than 100
```
garethellis0, line 19: should `100` here be replaced with `max_difference`? Also I'm not sure that returning `NaN` to indicate no value is the best option. Perhaps we could just set the `result` pointer to `NULL`?

boards/BMS/Src/App/App_Soc.c
```diff
@@ -0,0 +1,48 @@
...
+        return EXIT_CODE_INVALID_ARGS;
+
+    if (soc_1 < 0.0f || soc_1 > 100.0f)
+        return EXIT_CODE_INVALID_ARGS;
+
+    if (soc_2 < 0.0f || soc_2 > 100.0f)
+        return EXIT_CODE_INVALID_ARGS;
+
+    if (soc_3 < 0.0f || soc_3 > 100.0f)
+        return EXIT_CODE_INVALID_ARGS;
```
garethellis0, line 21: In general, I think the `if` without brackets is a tad dangerous because it's easy to accidentally add lines and not realize they're not under the `if`. Could we add brackets just to be super safe?

garethellis0: As another thought, if we're concerned that takes up too many lines, we could consider different formattings for it via `clang-format`.

boards/BMS/Inc/App/App_Soc.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "App_SharedExitCode.h"
+
+/**
+ * Given three state-of-charge (SoCs), check whether the difference between any
+ * two SoCs is less than or equal to the specified threshold. If any two SoCs
+ * pass the check, return the average value of those two SoCs.
+ * @param max_difference The maximum tolerable difference between any two given
```
kirons96, line 9: i'm not sure if thinking about this in terms of a max difference makes sense

if there's a single/double bit error in the EEPROM, or if a write is corrupted during power down, chances are the resulting float will be significantly different upon retrieval, and not within some range

the max difference doesn't hurt, but the way i did it in the past was just comparing the numbers, which should be equal. 

- if 3/3 numbers are equal, use that number
- if 2/3 numbers are equal, use that number and write over the third number
- if none are equal, we still need to run the car. we should map the voltage to an SoC by a LUT generated by the discharge data we gathered a while ago. it's rudimentary but a simple solution to an extremely unlikely problem

thekenu: I didn't want the rules to be so strict, and at worst we can just define the `max_abs_difference` to be `0`. 

As for the part about rational current limits, I agree but that's not the responsibility of this function. The caller would be using this function to see if none are equal, and handle it accordingly as you suggested.

kirons96: this is fine with me then, should i create an issue for doing the LUT?

boards/BMS/Src/App/App_Soc.c
thekenu: No problem, added the brackets

boards/BMS/Inc/App/App_Soc.h
thekenu: It should be `100`. I rewrote most of the comment so hopefully it's all more clear now. As for the `NULL` suggestion, the user is providing the storage so setting it to NULL would be the same as returning 0.0f I think?

boards/BMS/Inc/App/App_Soc.h
kirons96: or actually do you think that what i recommended should be in the spec?

boards/BMS/Inc/App/App_Soc.h
garethellis0: ah yeah, it would be. My bad, this looks fine.

boards/BMS/Inc/App/App_Soc.h
thekenu: I think you should just add it to the spec if possible

boards/BMS/Inc/App/App_Soc.h
```diff
@@ -0,0 +1,46 @@
...
+ *         max_difference = 5
+ *         soc_1 = 0
+ *         soc_2 = 50
+ *         soc_3 = 100
+ *
+ *       In this scenario, the difference between soc_1 and soc_2 is the same as
+ *       the difference between soc_2 and soc_3, but we prioritize soc_1 / soc_2
+ *       ovr soc_2 / soc_3. Thus, the average value to calculate,
+ *
+ *       (soc_1 + soc_2) / 2 = 25
```
thekenu, line 37:
```suggestion
 *       over soc_2 / soc_3. Thus, result will be set to  (soc_1 + soc_2) / 2 = 25    
```

PR #641, opened 2020-06-01
Add App and Io code for board status LEDs

thekenu:
### Summary

Implement DIM-2. Note that DIM-2 requires an error table so an error table was added to DIM world as well.

### Testing Done

- Integration test in Test_StateMachine.cpp
### Checklist


boards/DIM/Inc/App/App_BoardStatusLeds.h
```diff
@@ -0,0 +1,98 @@
...
+    void (*turn_on_dcm_blue)(void),
+    void (*turn_on_fsm_red)(void),
+    void (*turn_on_fsm_green)(void),
+    void (*turn_on_fsm_blue)(void),
+    void (*turn_on_bms_red)(void),
+    void (*turn_on_bms_green)(void),
+    void (*turn_on_bms_blue)(void),
+    void (*turn_on_pdm_red)(void),
+    void (*turn_on_pdm_green)(void),
+    void (*turn_on_pdm_blue)(void));
```
garethellis0, line 56: Hmm, doesn't seem like this allows any way to turn off led's? Perhaps we have something like:
```
enum LED_STATE {ON, OFF};

void setDimRedLedState(LED_STATE state);
```
?

garethellis0, line 56: Hmm, doesn't seem like this allows any way to turn off led's? Perhaps we have something like:
```
enum LED_STATE {ON, OFF};

void setDimRedLedState(LED_STATE state);
```
?

garethellis0, line 56: At another extreme, we could have:
```
struct SharedLed *App_SharedLed_Create(
    void (*enable)(void),
    void (*disable)(void));

struct BoardStatusLedGroup* App_BoardStatusLedGroup_Create(
     struct SharedLed* green, struct SharedLed* red, struct SharedLed* blue);

struct BoardStatusDisplay* App_BoardStatusDisplay_Create(
     struct BoardStatusLedGroup fsm_leds,
     ....
     );
```

And there are definitely halfway options between this suggestion and what's currently in code. 

I think it's OK as-is, but just wanted to present these options to see what you think. Personally I'd lean towards what I have above, but again, happy to approve it as-is.

thekenu: So That's actually why I had initially but then I realized the spec never requires it to be in the OFF state. It seems like we only need RED, GREEN, or BLUE, but never OFF. That was the reasoning behind it. Does it seem reasonable to you?

thekenu: That's fine with me, let me spend some time refactoring it

thekenu: On a second though, let's just go with the suggestion here: https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/641#discussion_r433611469

thekenu: I think I came up with a even cleaner solution. Take a look at the new changes and let me know what you think!

boards/shared/Inc/App/App_SharedRgbLed.h
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Allocate and initialize a RGB LED
+ * @param turn_red Turn the RGB LED red
+ * @param turn_green Turn the RGB LED green
+ * @param turn_blue Turn the RGB LED blue
+ * @param turn_off Turn off the RGB LED
+ * @return The created RGB LED, whose ownership is transferred to the caller
+ */
+struct RgbLed *App_SharedRgbLed_Create(
```
garethellis0, line 11: do we need to pre-declare this struct?

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -175,6 +176,42 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
         CANMSGS_DIM_SWITCHES_START_SWITCH_ON_CHOICE,
         CANMSGS_DIM_SWITCHES_START_SWITCH_OFF_CHOICE);
 
+    struct RgbLed *board_status_leds[NUM_BOARDS] = {
+        [BMS] = App_DimWorld_GetBmsStatusLed(world),
+        [DCM] = App_DimWorld_GetDcmStatusLed(world),
+        [DIM] = App_DimWorld_GetDimStatusLed(world),
+        [FSM] = App_DimWorld_GetFsmStatusLed(world),
+        [PDM] = App_DimWorld_GetPdmStatusLed(world),
```
garethellis0, line 184: This is interesting syntax that I've never seen before. Does `[BMS]` just effectively transform `BMS` into a number and insert into the array at that index?

thekenu: Yup that's it! TIL it doesn't seem supported in C++ so that's why it may seem odd to you!? 🤔  

https://eli.thegreenplace.net/2011/02/15/array-initialization-with-enum-indices-in-c-but-not-c

boards/shared/Inc/App/App_SharedRgbLed.h
thekenu: Oh yeah, not sure how this still compiles tbh 🙃

boards/shared/Inc/App/App_SharedRgbLed.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
```
thekenu, line 2:
```suggestion
#pragma once

struct RgbLed;

```
```suggestion
#pragma once

struct RgbLed;

```

PR #642, opened 2020-06-01
Add test for IO error table process function

thekenu:
### Checklist


boards/shared/Src/Io/Io_SharedErrorTable.c
```diff
@@ -71,38 +71,37 @@ static void Io_ProcessFsmNonCriticalErrorMsg(
 {
     SET_ERROR(
         error_table, FSM_NON_CRITICAL_PAPPS_OUT_OF_RANGE,
-        data->papps_out_of_range == 0);
```
thekenu, line 74: All the `== 0` are wrong. Not sure how I didn't notice that before ... Oh  well that's why we have tests.

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -754,3 +758,450 @@ TEST_F(
         ASSERT_TRUE(App_SharedError_IsBoardInList(&board_list, board));
     }
 }
+
+TEST_F(SharedErrorTableTest, process_bms_non_critical_errors)
+{
+    // A list of BMS non-critical error IDs. This must be maintained manually as
+    // errors are added to/removed from the DBC!
+    std::vector<enum ErrorId> bms_non_critical_error_ids = {
```
thekenu, line 766: If there's a way to _not_ have the user manually maintain this list, that would be helpful. But as of now I can't think of anything.

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -754,3 +758,450 @@ TEST_F(
...
+        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1KHZ,
+        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANRX,
+        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANTX,
+        BMS_NON_CRITICAL_WATCHDOG_TIMEOUT
+    };
+
+    // Each struct member that has a non-zero value indicates an error that is
+    // set. For this test, we want every error to be set. Instead of setting
+    // each struct member manually, we're going to "cheat" using memset().
+    struct CanMsgs_bms_non_critical_errors_t data;
```
thekenu, line 777: Initially, I wrote assigning each struct member manually but that was incredibly cumbersome and error-prone (For example, we might forget to assign a struct member). I think using `memset` here is justified, especially with the explanation in the comments.

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -86,6 +88,48 @@ class SharedErrorTableTest : public testing::Test
...
+    void TestRoutineForSetErrorsFromCanMsg(
+        enum Board                board,
+        std::vector<enum ErrorId> error_ids,
+        uint32_t                  can_stdid,
+        uint32_t                  can_dlc,
+        int (*pack_can_msg)(uint8_t *dst_p, const T *src_p, size_t size),
+        void (*get_boards)(const struct ErrorTable *, struct ErrorBoardList *),
+        void (*get_errors)(struct ErrorTable *, struct ErrorList *))
+    {
+        // Each struct member that has a non-zero value indicates an error that
```
thekenu, line 101: Initially, I was assigning each struct member manually but that was incredibly cumbersome and error-prone (For example, we might forget to assign a struct member). Also it didn't allow me to write this function to be shared across different boards.

I think using memset here as a "hack" is justified, especially with the explanation in the comments.

boards/shared/Test/Src/Test_ErrorTable.cpp
thekenu: This applies to all the boards

thekenu: One option could be to generate 
```
#define BMS_NON_CRITICAL_ERROR_IDS \
BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1HZ, \
        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1KHZ, \
        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANRX, \
        BMS_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANTX, \
        BMS_NON_CRITICAL_WATCHDOG_TIMEOUT. \
```

And stick it into `enum ErrorId`?

```
#include <SomeAutoGenHeader.h>
enum ErrorId
{
    BMS_NON_CRITICAL_ERROR_IDS,
    DCM_NON_CRITICAL_ERROR_IDS,
    ...
    BMS_CRITICAL_ERROR_IDS,
    DCM_CRITICAL_ERROR_IDS,
    ....
    NUM_ERROR_IDS,
}
```

And this test case would be rewritten as, where the user doesn't need to deal with the mental overhead of modifying this error ID list every time a BMS non-critical error is added to the DBC
```
TEST_F(SharedErrorTableTest, process_bms_non_critical_errors)
{
    std::vector<enum ErrorId> bms_non_critical_error_ids = {
       BMS_NON_CRITICAL_ERROR_IDS
    };

    TestRoutineForSetErrorsFromCanMsg(
        BMS, bms_non_critical_error_ids,
        CANMSGS_BMS_NON_CRITICAL_ERRORS_FRAME_ID,
        CANMSGS_BMS_NON_CRITICAL_ERRORS_LENGTH,
        App_CanMsgs_bms_non_critical_errors_pack,
        App_SharedErrorTable_GetBoardsWithNonCriticalErrors,
        App_SharedErrorTable_GetAllNonCriticalErrors);
}
````

Comment (garethellis0) [reply boards/shared/Test/Src/Test_ErrorTable.cpp:812]: Honestly I'd say generating the enum is fine, and I don't think we need to check the code in?

File: boards/shared/Test/Src/Test_ErrorTable.cpp [hunk 5]
```diff
@@ -86,6 +88,52 @@ class SharedErrorTableTest : public testing::Test
...
+        // Check that we can retrieve the correct errors from the error table
+        get_errors_from_error_table(error_table, &error_list);
+        ASSERT_EQ(error_ids.size(), error_list.num_errors);
+        for (auto &error_id : error_ids)
+        {
+            ASSERT_EQ(
+                EXIT_CODE_OK, App_SharedErrorTable_IsErrorSet(
+                                  error_table, error_id, &is_set));
+            ASSERT_TRUE(is_set);
+        }
```
garethellis0, line 134: So this test tells us if every single error is set, then all the errors in the table will return true. Really what we want to check is that setting a _single_ error in the CAN message sets that error in the table, no?

thekenu: So I was thinking what we're really trying to check here. I believe the main goal is that we have a `SET_ERROR()` in `Io_Process{Board}{Critical|NonCritical}ErrorMsg` for every error. If we set every single error at once, we'll be able to catch any missing `SET_ERROR`. Writing one test per error just seemed a bit cumbersome and it'll be really easy to forget to write a test or two.

boards/shared/Test/Src/Test_ErrorTable.cpp
thekenu: Sounds good let me go add in a script for this !

boards/shared/Test/Src/Test_ErrorTable.cpp
garethellis0: Ok, I can live with this.

boards/shared/Test/Src/Test_ErrorTable.cpp
garethellis0: Just to double check, this PR is still blocked on this script?

thekenu: Yup still working on it!

thekenu: Finally got around to fixing this!

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -86,7 +86,7 @@ BO_ 206 DCM_TORQUE_REQUEST : 4 DCM
 SG_ Torque_Request : 0|32@1+ (1,0) [-3.4E038|3.4E038] "" LOGGER
 
 BO_ 207 DCM_CRITICAL_ERRORS: 8 DCM
-SG_ DUMMY_CRITICAL_ERROR : 0|1@1+ (1,0) [0|1] "" DEBUG
+SG_ DUMMY : 0|1@1+ (1,0) [0|1] "" DEBUG
```
thekenu, line 89: Had to update the names after using codegen so the executable compiles without errors

scripts/codegen/ErrorId/generate_error_ids.py
```diff
@@ -0,0 +1,106 @@
...
+import argparse
+import os
+
+from scripts.utilities.supported_boards import get_board_names
+
+
+ERRORID_ENUM_TEMPLATE = '''\
+#pragma once
+
+#define BMS_NON_CRITICAL_ERRORS \\
```
thekenu, line 11: Errors are grouped using macros we so can use them in Test_ErrorTables.cpp

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -754,3 +804,159 @@ TEST_F(
...
+
+        App_SharedErrorTable_GetAllCriticalErrors);
+}
+
+TEST_F(SharedErrorTableTest, process_pdm_non_critical_errors)
+{
+    // A list of PDM non-critical error IDs. This must be maintained manually as
+    // errors are added to/removed from the DBC!
+    std::vector<enum ErrorId> pdm_non_critical_error_ids = {
+        PDM_NON_CRITICAL_ERRORS
```
thekenu, line 938: Instead of hard-coding the error enums by-hand, we now use the auto-generated macro (which expands to enums). This way, when we add/remove errors from the DBC we don't have to manually update this test case as well. The same applies to other boards.

boards/shared/Src/App/App_SharedErrorTable.c
```diff
@@ -54,6 +55,8 @@ struct ErrorTable *App_SharedErrorTable_Create(void)
     INIT_ERROR(FSM_NON_CRITICAL_RIGHT_WHEEL_SPEED_OUT_OF_RANGE, FSM, false);
     INIT_ERROR(FSM_NON_CRITICAL_PRIMARY_FLOW_RATE_OUT_OF_RANGE, FSM, false);
     INIT_ERROR(FSM_NON_CRITICAL_SECONDARY_FLOW_RATE_OUT_OF_RANGE, FSM, false);
+    INIT_ERROR(FSM_NON_CRITICAL_STEERING_ANGLE_OUT_OF_RANGE, FSM, false);
```
thekenu, line 58: With the changes mentioned [here](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/642/files#r436227848), I was able to identify errors I forgot to add (1 from DIM and 2 from FSM). This is the kind of human-error I wanted to avoid with the help of the codegen.

boards/shared/Src/Io/Io_SharedErrorTable.c
```diff
@@ -52,6 +52,9 @@ static void Io_ProcessDimNonCriticalErrorMsg(
     SET_ERROR(
         error_table, DIM_NON_CRITICAL_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1HZ,
         data->stack_watermark_above_threshold_task1_hz);
+    SET_ERROR(
```
thekenu, line 55: Similar error to the one mentioned [here](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/642/files#r436229383), except this time it's for `Io` layer

scripts/codegen/CAN/App_CanMsgs.dbc
garethellis0: Wait, why? I'm confused on how this fixed anything?

thekenu: So previously we had 

```
SG_ DUMMY_CRITICAL_ERROR ...
```
and the corresponding enum should have been

```
BOARD_CRITICAL_ERROR_DUMMY_CRITICAL_ERROR
```
but instead I (manually) wrote
```
BOARD_CRITICAL_ERROR_DUMMY
```

Now I could either update all instances of `BOARD_CRITICAL_ERROR_DUMMY` to `BOARD_CRITICAL_ERROR_DUMMY_CRITICAL_ERROR` or change `SG_ DUMMY_CRITICAL_ERROR` to `SG_ DUMMY` as a workaround, I chose the latter option. 

TL;DR I hand wrote the enum poorly previously, which conflicts with what the codegen outputs.

PR #644, opened 2020-06-03
Cycle through errors using 7-segment displays on DIM

thekenu:
### Summary

Slightly modify DIM-9 and implement the part about displaying errors on 7-segment displays.

### Testing Done

- Wrote tests
- Manually test using DIM hardware

### Checklist


boards/DIM/Inc/App/App_DimWorld.h
```diff
@@ -49,6 +49,32 @@ struct DimWorld *App_DimWorld_Create(
...
 
+/**
+ * Get the current second for the given world
+ * @note The second is rounded down:
+ *          - If the current time is 999ms, this function returns 0 second.
+ *          - If the current time is 1001ms, this function returns 1 second.
+ * @param world The world to get current second for
+ * @return The current second for the given world
+ */
+uint32_t App_DimWorld_GetCurrentSecond(struct DimWorld *world);
```
thekenu, line 60: To facilitate cycling through errors at 1Hz like we discussed, we need to have the concept of time. I thought it would be appropriate to add time to `World` with these APIs.

Note that `App_DimWorld_SetCurrentMillisecond` is called in `Task1kHz` for ARM target but it's called manually in the tests.

thekenu: Whatever solution we end up going with, we should eventually implement it on other boards that also require time.

thekenu: Actually it was kinda late and I meant to write a separate shared time module and pass it into world. 

I.e.

App_DimWorld_GetTime()
App_SharedTime_SetCurrentMs()
App_SharedTime_GetCurrentMs()
App_SharedTime_GetCurrentSec()

garethellis0: Yeah, this seems like an entirely reasonable solution to me.

boards/DIM/Src/App/App_DimWorld.c
```diff
@@ -6,6 +6,7 @@
 
 struct DimWorld
 {
+    uint32_t                  current_ms;
```
garethellis0, line 9: Perhaps `current_time_ms` just to be a bit more clear?

boards/DIM/Src/App/App_DimWorld.c
```diff
@@ -76,6 +78,23 @@ void App_DimWorld_Destroy(struct DimWorld *world)
     free(world);
 }
 
+uint32_t App_DimWorld_GetCurrentSecond(struct DimWorld *world)
+{
+    return world->current_ms / 1000;
```
garethellis0, line 83: Not sure how this will get truncated, might want to write it like:
``` suggestion
    return (uint32_t)((float)(world->current_ms) / 1000.0);
```
just to be safe

boards/DIM/Src/App/App_DimWorld.c
```diff
@@ -76,6 +78,23 @@ void App_DimWorld_Destroy(struct DimWorld *world)
...
+}
+
+uint32_t App_DimWorld_GetCurrentMillisecond(struct DimWorld *world)
+{
+    return world->current_ms;
+}
+
+void App_DimWorld_SetCurrentMillisecond(
```
garethellis0, line 91: Similar thought with naming these functions, we might want to say `SetCurrentTimeMilliseconds` just to be super clear (and similarly for the other ones).

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -212,6 +207,38 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+            seven_seg_displays,
+            (uint32_t)App_CanRx_BMS_STATE_OF_CHARGE_GetSignal_STATE_OF_CHARGE(
+                can_rx));
+    }
+    else
+    {
+        // We want to display each error for one second, so we assume that the
+        // error set will not change much and just associate each second to
+        // a particular error. Errors that are set for less than one second may
+        // not be displayed.
```
garethellis0, line 225: I think this should be `Errors that are set for less the N seconds, where N is the maximum possible number of errors, may not be displayed`?

boards/DIM/Src/App/App_DimWorld.c
thekenu: That's fine with me, i think i will use `floor` for this and write some test to make sure what's specified in the jdocs is true (i.e. 999ms rounds down to 0, and 1001ms rounds down to 1s).

Another thought is that `getCurrentTimeInSeconds` won't be perfect when it overflows because the last second will be shorter. But I guess for our use-case we will never leave the MCU on for that long anyway so no need to cover for that use-case?

thekenu: I can just document in the jdocs that this method is imperfect during an overflow?

boards/DIM/Src/App/states/App_DriveState.c
thekenu: Ah I see what you were saying. The type of error I was thinking of that if we're displaying error A for 1 second and another error B sneaks in during that but disappears before we stop displaying error A, then error B would be not be displayed. What you're saying is also correct though.

I went with what you said since it (sort-of) encompasses the error I described.

boards/DIM/Src/App/App_DimWorld.c
thekenu: Fixed!

boards/DIM/Src/App/App_DimWorld.c
thekenu: Fixed! Went with `SetCurrentTimeInMilliseconds` (Same as yours, except with an extra `In`)

boards/DIM/Src/App/App_DimWorld.c
thekenu: Another update, turns out using floating point will cause some rounding errors which I noticed while writing the test `check_current_time_overflow`! I could go into more detail but basically the root cause is that floating point can express all integer values exactly.

I stuck with using unsigned integers and wrote tests to make sure the behaviour is correct

garethellis0: I'm cool with it as long as we have carefully thought out test coverage.

boards/DIM/Src/App/states/App_DriveState.c
garethellis0: Did you push this change? I don't see it.

garethellis0:
> @garethellis0 I know you've approved this but can you give it one last quick pass before I merge, since I rewrote time as a separate Clock library in SharedClock.c?

Yep, no problem. This looks good to me, I like the separate clock.

boards/DIM/Src/App/states/App_DriveState.c
thekenu: Oops yup somehow forgot to push this small change!

PR #645, opened 2020-06-06
Add shared signal library

thekenu:
### Summary

Our specs necessitates the need to do something after a condition (e.g. Encoder fault) has persisted for some time. `SharedSignal` is the answer to that.

### Checklist


boards/FSM/Src/App/App_FsmWorld.c
```diff
@@ -121,3 +133,20 @@ struct RgbLedSequence *
...
 }
+
+void App_FsmWorld_RegisterSignal(struct FsmWorld *world, struct Signal *signal)
+{
+    assert(world->num_signals < MAX_NUM_SIGNALS);
+    world->signals[world->num_signals] = signal;
+    world->num_signals++;
+}
+
+void App_FsmWorld_UpdateSignals(
```
thekenu, line 144: Not quite sure where this function should be called. If it's put inside `RunTask1kHz`, then the signals won't be updated in `Test_StateMachine` so I thought it'd be more appropriate to call it in a state machine. If I go with that approach, I run into  two problems:

- I will have to update the state machine library to support 1kHz because it only supports 1Hz and 100Hz at the moment (Not really a "problem", but just something we need to add)
- Inside the state machine, there is no "time" so I'm not sure what value to pass in for `current_time_ms`. A potential solution would be to add `App_FsmWorld_GetTimeMs` by passing in `get_time_ms` to `App_FsmWorld_Create()`

garethellis0: I think we need to call it separately. It might be worth having a test util function called `letTimePass(unsigned int time_ms, StateMachine* state_machine)` or something that manually calls both this and the 1hz and 1khz functions? (hopefully that makes sense, I can write out more of an example if not)

boards/shared/Inc/App/App_SharedSignal.h
```diff
@@ -0,0 +1,53 @@
...
+ * @param signal The signal to get the time from
+ * @retur The last time the given signal was low, in milliseconds
+ */
+uint32_t App_SharedSignal_GetLastTimeLowMs(const struct Signal *signal);
+
+/**
+ * Get the last time the given signal was high, in milliseconds
+ * @param signal The signal to get the time from
+ * @retur The last time the given signal was high, in milliseconds
+ */
```
garethellis0, line 43: In general, I'd suggest changing `was high` and `was low` to `was observed to be high` and `was observed to be low` just to be super clear.

boards/shared/Inc/App/App_SharedSignal.h
```diff
@@ -0,0 +1,53 @@
...
+uint32_t App_SharedSignal_GetLastTimeHighMs(const struct Signal *signal);
+
+/**
+ * Update the internal state of the given signal. If the signal has been
+ * continuously high for a period equal to or greater than the configured
+ * duration (See: `duration_high_ms`), the callback function will be triggered.
+ * @param signal The signal to update
+ * @param current_time_ms The current time, in milliseconds
```
garethellis0, line 51: Might want to note here that this has the side effect of actually calling the function that checks if the signal is high or not.

boards/shared/Src/App/App_SharedSignal.c
```diff
@@ -0,0 +1,75 @@
...
+#include <stdlib.h>
+#include "App_SharedSignal.h"
+
+struct Signal
+{
+    // The last time this signal was low, in milliseconds
+    uint32_t last_time_low_ms;
+
+    // The last time this signal was high, in milliseconds
+    uint32_t last_time_high_ms;
```
garethellis0, line 11: similar comment here about noting `was observed to be` instead of just `was`

boards/shared/Src/App/App_SharedSignal.c
```diff
@@ -0,0 +1,75 @@
...
+
+    // The function to call to check if this signal is high
+    bool (*is_high)(void);
+
+    // How long this signal must be continuously high for, in milliseconds,
+    // before the callback function is triggered
+    uint32_t duration_high_ms;
```
garethellis0, line 18:
``` suggestion
    uint32_t duration_high_ms_before_callback_triggered
```
And similarly for function args.

boards/FSM/Src/main.c
```diff
@@ -148,6 +149,16 @@ static void CanTxQueueOverflowCallBack(size_t overflow_count)
...
 
+static bool ReturnTrue(void)
+{
+    return true;
+}
+
+static void DemoCallBack(void)
+{
+    printf("Test\n");
+}
```
garethellis0, line 160: I'm assuming we want to remove the changes to this file before we merge?

boards/FSM/Src/App/App_FsmWorld.c
thekenu: Can you write an example how `UpdateSignal` would get used in embedded and test targets?

boards/FSM/Src/main.c
thekenu: Yup!

boards/shared/Inc/App/App_SharedSignal.h
thekenu: Fixed!

boards/shared/Inc/App/App_SharedSignal.h
thekenu: Added a `@note`

boards/shared/Src/App/App_SharedSignal.c
thekenu: Fixed

boards/FSM/Src/main.c
thekenu: I'm hoping to write a real-example before this gets merged in, but for now I will just stick with these dummy functions so I can get things to compile

boards/shared/Src/App/App_SharedSignal.c
thekenu: That seemed a bit verbose but I see how the current name is unclear. I will fall back to using the `struct SignalCallback` that you suggested in the beginning.

i.e.,
```
struct SignalCallback {
  uint32_t high_duration_ms;
  void (*function)(void)
};
```

boards/FSM/Src/App/App_FsmWorld.c
```diff
@@ -121,3 +130,26 @@ struct RgbLedSequence *
...
 }
+
+void App_FsmWorld_RegisterSignal(struct FsmWorld *world, struct Signal *signal)
+{
+    struct SignalNode *item = malloc(sizeof(struct SignalNode));
+    assert(item != NULL);
+    item->signal = signal;
+    item->next   = NULL;
+
+    SL_APPEND(world->signals_head, item);
```
thekenu, line 141: Instead of relying on hard-coded values like `MAX_NUM_SIGNALS`, I thought it's time that we move towards using data structure from libraries to make the game more elegant.

I replaced `MAX_NUM_SIGNAL` with a header-only linked list C library: https://github.com/nbulischeck/list.h. Let me know what you think about this decision.

boards/FSM/Src/App/App_FsmWorld.c
garethellis0:
```
void letTimePass(struct SharedStateMachine* state_machine, unsigned int time_ms){
    for (unsigned int i = 0; i < time_ms; i++){
        if (i % 1000 == 0) {
            app_SharedStateMachine_tick1Hz(state_machine);
        }
        app_SharedStateMachine_tick1kHz(state_machine);
    }
    // So we'd probably have to do something where this is a virtual function
    // that is implemented in each of the board-specific test fixtures. I've
    // put an example down below for erferece
    updateSignals(state_machine);
}

// This would have to be board-specific. We can enforce that people implement
// it if we have a shared test fixture that all the board-specific test 
// fixtures extend
void updateSignals(struct SharedStateMachine* state_machine){
    // This specific example is for the DCM
    struct DcmWorld* world = app_SharedStateMachine_getWorld(state_machine);
    app_DcmWorld_updateSignals(world);
}
```

boards/FSM/Src/App/App_FsmWorld.c
garethellis0: I'm fine with it. Obviously we just need to be a bit more careful about properly de-allocating things in the destructor to avoid memory leaks.

boards/FSM/Src/App/App_FsmWorld.c
thekenu: This is more or less what I had in mind, looks good!

thekenu: Also slightly off topic, but later we would probably call `SetCurrentTimeInMilliseconds` from #644 in `letTimePass` as well right?

boards/FSM/Inc/App/App_AcceleratorPedal.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
thekenu, line 1: Later on, accelerator pedal will have things like `GetPedalPosition` and what not. Keeping the interface with just enough methods to work with `Signal`.

boards/FSM/Test/Src/Test_StateMachine.cpp
```diff
@@ -43,12 +45,16 @@ FAKE_VALUE_FUNC(float, get_right_wheel_speed);
 FAKE_VALUE_FUNC(float, get_steering_angle);
 FAKE_VALUE_FUNC(float, get_brake_pressure);
 FAKE_VALUE_FUNC(bool, is_brake_actuated);
+FAKE_VALUE_FUNC(bool, is_papps_encoder_alarm_active);
+FAKE_VALUE_FUNC(bool, is_sapps_encoder_alarm_active);
 
-class FsmStateMachineTest : public testing::Test
+class FsmStateMachineTest : public BaseStateMachineTest
```
thekenu, line 51: Only FSM is using `BaseStateMachineTest` for now, I will fix this up for other boards later on.

boards/FSM/Src/App/App_Signals.c
```diff
@@ -0,0 +1,52 @@
...
+
+static void SappsAlarmCallback(void *context)
+{
+    struct FsmWorld *         world  = (struct FsmWorld *)context;
+    struct FsmCanTxInterface *can_tx = App_FsmWorld_GetCanTx(world);
+
+    App_CanTx_SetPeriodicSignal_SAPPS_MAPPED_PEDAL_PERCENTAGE(can_tx, 0.0f);
+}
+
+void App_Signals_Init(struct FsmWorld *world)
```
thekenu, line 33: This file is needed because I need to make sure the same `IsActive` and `Callback` are used for the embedded and test targets.

boards/FSM/Src/main.c
```diff
@@ -819,8 +834,12 @@ void RunTask1kHz(void const *argument)
 
     for (;;)
     {
-        Io_CanTx_EnqueuePeriodicMsgs(
-            can_tx, osKernelSysTick() * portTICK_PERIOD_MS);
+        const uint32_t current_time_ms = osKernelSysTick() * portTICK_PERIOD_MS;
+
+        App_SharedClock_SetCurrentTimeInMilliseconds(clock, current_time_ms);
+        App_FsmWorld_UpdateSignals(world, current_time_ms);
```
thekenu, line 840: These two lines are what `LetTimePass` is doing, but for the embedded target.

boards/FSM/Src/App/App_Signals.c
```diff
@@ -0,0 +1,52 @@
...
+    _papps = App_FsmWorld_GetPapps(world);
+    _sapps = App_FsmWorld_GetSapps(world);
+
+    struct SignalCallback callback;
+    struct Signal *       signal;
+
+    callback.function         = PappsAlarmCallback;
+    callback.high_duration_ms = 10;
+    callback.context          = world;
+    signal = App_SharedSignal_Create(0, IsPappsAlarmActive, callback);
```
garethellis0, line 44: Somewhat of a nit, but I think it's a bit cleaner to init stuff in one statement, ie:
```
    struct SignalCallback callback  = {
        .function         = PappsAlarmCallback,
        .high_duration_ms = 10,
        .context          = world
    }
    struct Signal* signal = App_SharedSignal_Create(0, IsPappsAlarmActive, callback);
```
similarly for down below.

boards/FSM/Src/App/App_Signals.c
```diff
@@ -0,0 +1,52 @@
...
+
+    App_CanTx_SetPeriodicSignal_PAPPS_MAPPED_PEDAL_PERCENTAGE(can_tx, 0.0f);
+}
+
+static bool IsSappsAlarmActive(void)
+{
+    return App_AcceleratorPedal_IsEncoderAlarmActive(_papps);
+}
+
+static void SappsAlarmCallback(void *context)
```
garethellis0, line 25: I think I'd prefer we just pass in either `World` or `SharedStateMachine` here, just because it's more clear.

boards/FSM/Src/App/App_Signals.c
```diff
@@ -0,0 +1,52 @@
...
+{
+    struct FsmWorld *         world  = (struct FsmWorld *)context;
+    struct FsmCanTxInterface *can_tx = App_FsmWorld_GetCanTx(world);
+
+    App_CanTx_SetPeriodicSignal_PAPPS_MAPPED_PEDAL_PERCENTAGE(can_tx, 0.0f);
+}
+
+static bool IsSappsAlarmActive(void)
+{
+    return App_AcceleratorPedal_IsEncoderAlarmActive(_papps);
```
garethellis0, line 22:
``` suggestion
    return App_AcceleratorPedal_IsEncoderAlarmActive(_sapps);
```
not sure how extensively we're testing this version of the shared library or the FSM specific stuff here, but might want to add at least one or two cases that would catch this

boards/FSM/Src/App/App_Signals.c
garethellis0: If these signals are an implicit part of `FsmWorld`, why not set them up in the `FsmWorld_Create` function?

boards/FSM/Src/App/App_Signals.c
thekenu: Fine with me, fixed!

boards/FSM/Src/App/App_Signals.c
thekenu: Fixed!

boards/FSM/Src/App/App_Signals.c
thekenu: That's a good point actually, let me try it out and  see how it looks

thekenu: I tried this out but ran into some problems with this function:
```
static bool IsPappsAlarmActive(void)                                                                                 
{                                                                                                                    
     return App_AcceleratorPedal_IsEncoderAlarmActive(_papps);                                                        
}
```

The problem is it's unclear where `_papps` should exist and when it should be initialized. I could place it in `App_FsmWorld.cpp` but it feels kind of odd.

```
// App_FsmWorld.c
static AcceleratorPedal* _papps;

static bool IsPappsAlarmActive(void)                                                                                 
{                                                                                                                    
     return App_AcceleratorPedal_IsEncoderAlarmActive(_papps);                                                        
}

FsmWorld* App_FsmWorld_Create(
  ...,
  papps)
{
  ...
  _papps = papps;
  struct Signal *papps_signal =                                                                                    
    App_SharedSignal_Create(0, IsPappsAlarmActive, papps_callback);                                              
  App_FsmWorld_RegisterSignal(world, papps_signal);  
   ...
}
```
This should work but it it feels odd because it's hardcoding the same `_papps` for every call to `App_FsmWorld_Create`. What do you think?

boards/FSM/Src/App/App_Signals.c
thekenu: Turns out the problem lies in `App_AcceleratorPedal.c`

I wrote 
```
return accelerator_pedal->is_encoder_alarm_active;
```
instead of 
```
return return accelerator_pedal->is_encoder_alarm_active();
```

Kind of a silly bug. Turns out I just forgot to write `Test_AcceleratorPedal` so this wasn't caught :sweat_smile:

boards/FSM/Src/App/App_Signals.c
thekenu: Turns out the problem lies in `App_AcceleratorPedal.c`

I wrote 
```
// This _always_ returns  true!
return accelerator_pedal->is_encoder_alarm_active;
```
instead of 
```
return return accelerator_pedal->is_encoder_alarm_active(); 
```

Kind of a silly bug. Turns out I just forgot to write `Test_AcceleratorPedal` so this wasn't caught :sweat_smile:

boards/FSM/Src/App/App_Signals.c
thekenu: This function signature has to conform to `struct SignalCallback` so it must be a void pointer

```
    8   struct SignalCallback                                                                                                                                                                                                                             
    1 {                                                                                                                                                                                                                                                 
    2     // How long the signal must be continuously high for, in milliseconds,                                                                                                                                                                        
    3     // before the callback function is triggered                                                                                                                                                                                                  
    4     uint32_t high_duration_ms;                                                                                                                                                                                                                    
    5                                                                                                                                                                                                                                                   
    6     // The callback function                                                                                                                                                                                                                      
    7     void (*function)(void *);                                                                                                                                                                                                                     
    8                                                                                                                                                                                                                                                   
    9     // The context to pass into the callback function. Set as NULL if unused.                                                                                                                                                                     
   10     void *context;                                                                                                                                                                                                                                
   11 };                                                                                                                                                                                                                                                
   12    
```
I can't think of an elegant way to do this otherwise but let me if there's a clever solution I'm missing

garethellis0: Why not have function just take a `World` instead?

boards/FSM/Src/App/App_Signals.c
garethellis0: Why not add an argument to `App_FsmWorld_Create`: `bool (*is_papps_alarm_active)(void)`? Then we can just pass the function in.

garethellis0: Another option is to chnage `function` to take a `World`, then we can just write the function as:
```
bool IsPappsAlarmActive(FsmWorld* world) {
    return return App_AcceleratorPedal_IsEncoderAlarmActive(world->_papps);
}
```
and pass it in when we create the signal

boards/FSM/Src/App/App_Signals.c
thekenu: I guess as of now `Signal` is meant to be used in conjunction with `World*` only so this would make sense. Changed!

boards/PDM/Inc/App/configs/App_SharedSignalConfig.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+struct PdmWorld;
```
thekenu, line 3: Had to add this for all boards as required by `SharedSignal.h` to compile successfully (Even if a board isn't using `Signal`)

boards/shared/Inc/App/App_SharedSignal.h
```diff
@@ -0,0 +1,67 @@
...
+ * @param initial_time_ms The initial time, in milliseconds, used to initialize
+ *                        the internal state of the signal
+ * @param is_high A function that can be called to check if the signal is high
+ * @param callback The signal callback for the signal
+ * @return The created signal, whose ownership is given to the caller
+ */
+struct Signal *App_SharedSignal_Create(
+    uint32_t initial_time_ms,
+    bool (*is_high)(struct World *),
+    struct World *        world,
```
thekenu, line 34: Since `world` is used by the `callback` _and_ `is_high`, I thought it would make sense to put it in `struct Signal` as opposed to `struct SignalCallback`.

boards/FSM/Src/App/App_Signals.c
thekenu: Went with the second option so we don't have to worry about initializing static variables `_papps` and `_sapps`. I think this new approach is quite a bit cleaner!

thekenu: Also since `RegisterSignal` is a public function, I thought it made sense to call it _outside` of `FsmWorld_Create`. Now the code calls `RegisterSignal` after `FsmWorld_Create`, i.e.

```
    world = App_FsmWorld_Create(... papps, sapps);

    struct SignalCallback papps_callback = {
        .high_duration_ms = 10,
        .function         = App_AcceleratorPedalSignals_PappsAlarmCallback,
    };
    papps_signal = App_SharedSignal_Create(
        0, App_AcceleratorPedalSignals_IsPappsAlarmActive, world,
        papps_callback);
    App_FsmWorld_RegisterSignal(world, papps_signal);

    struct SignalCallback sapps_callback = {
        .high_duration_ms = 10,
        .function         = App_AcceleratorPedalSignals_SappsAlarmCallback,
    };
    sapps_signal = App_SharedSignal_Create(
        0, App_AcceleratorPedalSignals_IsSappsAlarmActive, world,
        sapps_callback);
    App_FsmWorld_RegisterSignal(world, sapps_signal);

garethellis0: Hmm, so I think I still prefer this in the `Create` function, because since we would presumably pass in the callback functions to the world function constructor, this forces any users of the world to properly setup the callback. I suppose I'm trying to minimize any duplication that might be required between here and the tests.

garethellis0:
Just one comment, otherwise LGTM. This has turned out pretty nicely!

boards/FSM/Src/App/App_Signals.c
thekenu: That's fine with me too. I will make `RegisterSignal` private and also create another private function `UnregisterSignal` to clean things up in `World_Destroy`.

PR #646, opened 2020-06-06
Add brake pressure sensor OC/SC detection 

pasitlaothamatas:
### Summary

Added a function to detect if the brake pressure sensor is open circuited or short circuited

### Testing Done

- created unit test in Test_StateMachine to check if the OC/SC status is broadcasted in all states

### Checklist


boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
 #pragma once
 
+#include <stdbool.h>
+
 /**
  * Check if the pressure from MSP-300-2K5-P-3-N-1 exceeds the brake pressure
  * threshold. If the pressure exceeds threshold, the brake is said to be
- * actuated.
+ * actuated
```
thekenu, line 8:
```suggestion
 * actuated.
```

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
...
  * @return True if the brake is actuated, false if the brake is not actuated
  */
 bool Io_MSP3002K5P3N1_IsBrakeActuated(void);
 
 /**
- * Get the pressure in psi from the MSP-300-2K5-P-3-N-1 brake pressure sensor
+ * Check if the MSP-300-2K5-P-3-N-1 is in an open circuit or short circuit
+ * state. If the output voltage exceeds the upper threshold or goes below the
+ * minimum threshold, the brake pressure sensor is said to be open/short
+ * circuited
```
thekenu, line 17: I don't think this is necessary. Previously it wasn't clear what brake is "actuated" means. But here open-circuit and short-circuit are well-defined and unambiguous.
```suggestion
 * state.
```

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
...
+ * @return True if the brake is open/short circuited, false if the brake is ok
+ */
+bool Io_MSP3002K5P3N1_IsBrakeOpenOrShortCircuited(void);
+
+/**
+ * Get the pressure in psi from the MSP-300-2K5-P-3-N-1 brake pressure
+ * sensor
```
thekenu, line 24:
```suggestion
 * Get the pressure in psi from the MSP-300-2K5-P-3-N-1 brake pressure sensor
```

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
...
  */
 bool Io_MSP3002K5P3N1_IsBrakeActuated(void);
 
 /**
- * Get the pressure in psi from the MSP-300-2K5-P-3-N-1 brake pressure sensor
+ * Check if the MSP-300-2K5-P-3-N-1 is in an open circuit or short circuit
+ * state. If the output voltage exceeds the upper threshold or goes below the
+ * minimum threshold, the brake pressure sensor is said to be open/short
+ * circuited
+ * @return True if the brake is open/short circuited, false if the brake is ok
```
thekenu, line 18:
```suggestion
 * @return True if the sensor is open or short circuited, false if the sensor is not
```

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
...
+ */
+bool Io_MSP3002K5P3N1_IsBrakeOpenOrShortCircuited(void);
```
thekenu, line 20:
```suggestion
bool Io_MSP3002K5P3N1_IsSensorOpenOrShortCircuited(void);
```

boards/FSM/Src/App/App_FsmWorld.c
```diff
@@ -16,6 +16,7 @@ struct FsmWorld
     struct InRangeCheck *     steering_angle_in_range_check;
     struct InRangeCheck *     brake_pressure_in_range_check;
     struct BinaryStatus *     brake_actuation_status;
+    struct BinaryStatus *     brake_oc_sc_status;
```
thekenu, line 19:
```suggestion
    struct BinaryStatus *     brake_sensor_open_or_short_circuit_status;
```

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -341,6 +345,7 @@ VAL_ 300 STEERING_ANGLE_OUT_OF_RANGE 0 "OK" 1 "UNDERFLOW" 2 "OVERFLOW";
 VAL_ 300 BRAKE_PRESSURE_OUT_OF_RANGE 0 "OK" 1 "UNDERFLOW" 2 "OVERFLOW";
 VAL_ 305 State 0 "AIR_OPEN" 1 "AIR_CLOSED";
 VAL_ 308 Brake_Is_Actuated 0 "FALSE" 1 "TRUE";
+VAL_ 313 Brake_Is_OC_SC 0 "FALSE" 1 "TRUE";
```
thekenu, line 348:
```suggestion
VAL_ 313 Brake_Sensor_Is_Open_Or_Short_Circuit 0 "FALSE" 1 "TRUE";
```

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -141,6 +141,9 @@ SG_ Steering_Angle : 0|32@1+ (1,0) [-110|110] "deg" DCM
 BO_ 312 FSM_BRAKE_PRESSURE_SENSOR: 4 FSM
 SG_ Brake_Pressure: 0|32@1+ (1,0) [0|2500] "psi" DEBUG 
 
+BO_ 313 FSM_BRAKE_OC_SC: 1 FSM
+SG_ Brake_Is_OC_SC: 0|1@1+ (1,0) [0|1] "" DEBUG
```
thekenu, line 145: I would probably put this in `BO_ 312` instead of making a new message

boards/FSM/Inc/App/App_SetPeriodicCanSignals.h
```diff
@@ -10,3 +10,5 @@ void App_SetPeriodicSignals_SteeringAngleInRangeCheck(
 void App_SetPeriodicSignals_BrakePressureInRangeCheck(
     const struct FsmWorld *world);
 void App_SetPeriodicSignals_BrakeActuationStatus(const struct FsmWorld *world);
+void App_SetPeriodicSignals_BrakeOpenOrShortCircuitedStatus(
```
thekenu, line 13:
```suggestion
void App_SetPeriodicSignals_BrakeSensorOpenOrShortCircuitStatus(
```

boards/FSM/Src/App/App_SetPeriodicCanSignals.c
```diff
@@ -104,3 +104,17 @@ void App_SetPeriodicSignals_BrakeActuationStatus(const struct FsmWorld *world)
         CANMSGS_FSM_BRAKE_BRAKE_IS_ACTUATED_TRUE_CHOICE,
         CANMSGS_FSM_BRAKE_BRAKE_IS_ACTUATED_FALSE_CHOICE);
 }
+
+void App_SetPeriodicSignals_BrakeOpenOrShortCircuitedStatus(
```
thekenu, line 108:
```suggestion
void App_SetPeriodicSignals_BrakeSensorOpenOrShortCircuitStatus(
```

boards/FSM/Inc/App/App_FsmWorld.h
```diff
@@ -29,6 +29,7 @@ struct FsmWorld *App_FsmWorld_Create(
     struct InRangeCheck *     steering_angle_in_range_check,
     struct InRangeCheck *     brake_pressure_in_range_check,
     struct BinaryStatus *     brake_actuation_status,
+    struct BinaryStatus *     brake_oc_sc_status,
```
garethellis0, line 32: could we rename this to `brake_open_circuit_or_short_circuit`? Just to be a bit more clear.

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
```diff
@@ -1,15 +1,27 @@
...
+#include <stdbool.h>
+
 /**
  * Check if the pressure from MSP-300-2K5-P-3-N-1 exceeds the brake pressure
  * threshold. If the pressure exceeds threshold, the brake is said to be
- * actuated.
+ * actuated
  * @return True if the brake is actuated, false if the brake is not actuated
  */
 bool Io_MSP3002K5P3N1_IsBrakeActuated(void);
```
garethellis0, line 11: As a general thought for these functions, is there anything to `MSP3002K5P3N1` sensor that is actually related to braking if we're not looking at it in the context of our car? The point I'm trying to make here is that this should solely be an abstraction around the `MSP3002K5P3N1`. An abstraction around the brakes might have an instance of this as part of it. 

Hope that's clear. Not super critical for this PR, but just a thought.

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: Okay this should be good

boards/FSM/Inc/App/App_FsmWorld.h
pasitlaothamatas: Ken addressed this earlier. I renamed it to brake_open_or_short_circuit.

boards/FSM/Inc/Io/Io_MSP3002K5P3N1.h
pasitlaothamatas: I'm still a bit confused. In this case are you suggesting that functions like `IsBrakeActuated` and `IsOpenOrShortCircuit` should fall under the abstraction around the brakes instead of the `MSP3002K5P3N1` sensor?

thekenu: I agree with Gareth. I think IsBrakeActuated should go elsewhere while IsOpenOrShortCircuit can stay since it's directly related to the sensor itself

thekenu: Not sure what the exact kind of abstraction @garethellis0 had in mind

garethellis0: I was just making a comment that generally we should only put stuff under `Io_MSP3002K5P3N1` that is _specific_ to the `MSP3002K5P3N1`. So if you think of the `MSP3002K5P3N1`, does it have the concept of "isBrakeActuated" or `isOpenOrShortCircuit"? Or are those things that we are associating with it because of our specific use-case of it? If they are not specific to this sensor, then we should have an `App_Brake` that wraps around it.

Sorry if this is a bit confusing, it's a little hard for me to convey in text. Let me know if you want to hope in a call to clarify it quickly!

pasitlaothamatas: Oh ok! thx for clearing this up

boards/FSM/Src/App/App_Brake.c
```diff
@@ -0,0 +1,70 @@
+#include <stdlib.h>
```
pasitlaothamatas, line 1: @garethellis0 @thekenu Can you guys check to see if this is along the lines of what you were looking for b4 I continue? I added moved the `IsBrakeActuated` to `Io_Brake`, and used `App_Brake` to wrap around the Io fxns

boards/FSM/Src/App/App_Brake.c
```diff
@@ -0,0 +1,70 @@
...
+{
+    struct InRangeCheck *brake_pressure_in_range_check;
+    struct BinaryStatus *is_brake_open_or_short_circuit_binary_status,
+        *is_brake_actuated_binary_status;
+
+    float min_brake_pressure_psi;
+    float max_brake_pressure_psi;
+
+    float (*get_pressure_psi)(void);
+    bool (*is_open_or_short_circuit)(void);
```
garethellis0, line 16: bit of a nit, but I think `pressure_sensor_has_open_short_circuit` just to be _super_ clear.

garethellis0, line 16: bit of a nit, but I think `pressure_sensor_has_open_short_circuit` just to be _super_ clear.

boards/FSM/Src/App/App_FsmWorld.c
```diff
@@ -30,6 +32,8 @@ struct FsmWorld *App_FsmWorld_Create(
     struct InRangeCheck *const      steering_angle_in_range_check,
     struct InRangeCheck *const      brake_pressure_in_range_check,
     struct BinaryStatus *const      brake_actuation_status,
+    struct BinaryStatus *const      brake_sensor_open_or_short_circuit_status,
+    struct Brake *const             brake,
```
garethellis0, line 36: It seems a bit odd to me that we're passing in this status alongside the `Brake`, when you can really get both just from the `Brake`?

boards/FSM/Test/Src/Test_StateMachine.cpp
```diff
@@ -329,15 +338,26 @@ TEST_F(FsmStateMachineTest, check_brake_pressure_can_signals_in_all_states)
...
+    CheckBinaryStatusCanSignalInAllStates(
         is_brake_actuated_fake.return_val,
         App_CanTx_GetPeriodicSignal_BRAKE_IS_ACTUATED,
         CANMSGS_FSM_BRAKE_BRAKE_IS_ACTUATED_TRUE_CHOICE,
         CANMSGS_FSM_BRAKE_BRAKE_IS_ACTUATED_FALSE_CHOICE);
 }
 
+TEST_F(
+    FsmStateMachineTest,
+    check_brake_open_or_short_circuited_can_signal_in_all_states)
```
garethellis0, line 352: Is there a spec we can associate with this test?

boards/FSM/Src/App/App_FsmWorld.c
pasitlaothamatas: The other files aren't up to date yet. Just wanted either you or ken to check `App_Brake` before I start changing stuff.

boards/FSM/Src/App/App_Brake.c
thekenu: Yeah I think that would be nice, since the concept of brake doesn't short circuit. It's the underlying sensor that does.

On a semi-related note, I'd probably rename all instances of `brake_pressure` to `pressure`  (or the other way around) for consistency

boards/FSM/Inc/Io/Io_Brake.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <stdbool.h>
+
+/**
+ * Check if the pressure from MSP-300-2K5-P-3-N-1 exceeds the brake pressure
+ * threshold. If the pressure exceeds threshold, the brake is said to be
+ * actuated.
+ * @return True if the brake is actuated, false if the brake is not actuated
```
thekenu, line 9: Since we moved this to `Io_Brake`, the jdocs should be independent of the sensor because that's an implementation detail
 ```suggestion
/**
 * Check if the brake is actuated
 * @return True if the brake is actuated, false if the brake is not actuated
```

File: boards/FSM/Src/App/App_Brake.c [hunk 18]
```diff
@@ -0,0 +1,70 @@
...
+
+    return brake;
+}
+
+void App_Brake_Destroy(struct Brake *brake)
+{
+    free(brake);
+}
+
+void App_Brake_GetBrakePressureInRangeCheck(
```
thekenu, line 52: This function doesn't seem to make sense, I can't tell if you're  trying to return the struct InRangeCheck (based on the function name) or enum InRangeCheck_Status. In either case, currently this returns void which just doesn't make sense to me.

boards/FSM/Src/App/App_Brake.c
```diff
@@ -0,0 +1,70 @@
...
+{
+    struct Brake *brake = malloc(sizeof(struct Brake));
+    assert(brake != NULL);
+
+    brake->brake_pressure_in_range_check = App_InRangeCheck_Create(
+        get_pressure_psi, min_brake_pressure_psi, max_brake_pressure_psi);
+    brake->is_brake_open_or_short_circuit_binary_status =
+        App_SharedBinaryStatus_Create(is_open_or_short_circuit);
+    brake->is_brake_actuated_binary_status =
+        App_SharedBinaryStatus_Create(is_brake_actuated);
```
thekenu, line 35: These need to be freed in the destroy function

boards/FSM/Test/Src/Test_StateMachine.cpp
pasitlaothamatas: This seems to be used for debugging. No specs associated to the test

garethellis0: Ah, sounds good. Could we leave a quick comment indicating as such?

boards/FSM/Src/Io/Io_MSP3002K5P3N1.c
```diff
@@ -1,6 +1,6 @@
-#include <stdbool.h>
 #include "Io_Adc.h"
 #include "main.h"
+#include "Io_MSP3002K5P3N1.h"
 
 bool Io_MSP3002K5P3N1_IsBrakeActuated(void)
```
garethellis0, line 5: Hmm, it seems like this function is duplicated between here and `Io_Brake`? Could we remove one of them?

garethellis0: Also I'm not entirely clear on what "actuated" means here?

pasitlaothamatas: Would using IsBrakeApplied be clearer?

pasitlaothamatas: The rule book uses actuated. I'm fine with either.

actuated

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+struct BinaryStatus *
+    App_Brake_GetIsPressureSensorOpenOrShortCircuitBinaryStatus(
+        const struct Brake *brake);
+
+/**
+ * Deallocate the memory used by the given brake
+ * @param brake The brake to deallocate
+ */
+void App_Brake_Destroy(struct Brake *brake);
```
thekenu, line 58: nit; Would be nice to preserver the order where `Destroy` is right below `Create`

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,58 @@
...
+    float min_pressure_psi,
+    float max_pressure_psi);
+
+/**
+ * Get the pressure in-range check for the given brake
+ * @param brake The brake to get the pressure in-range check for
+ * @return The pressure in-range check for the given brake
+ */
+struct InRangeCheck *
+    App_Brake_GetPressureInRangeCheck(const struct Brake *const brake);
```
thekenu, line 32:
```suggestion
    App_Brake_GetPressureInRangeCheck(const struct Brake * brake);
```

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,58 @@
...
+ * Get the pressure sensor open or short circuit binary status for the given
+ * brake
+ * @param brake The brake to get the pressure sensor open or short circuit
+ * binary status for
+ * @return The pressure sensor open or short circuit binary status for the given
+ * brake
+ */
+struct BinaryStatus *
+    App_Brake_GetIsPressureSensorOpenOrShortCircuitBinaryStatus(
+        const struct Brake *brake);
```
thekenu, line 52: Now that we wrapped the idea of `IsPressureSensorOpenOrShortCircuit` and `IsBrakeActuated` inside `App_Brake`, I don't think there's really any point to using `BinaryStatus` any more. When I made `BinaryStatus` it was more for standalone objects.

I think this interface would be easier to use and cleaner if it's just `App_Brake_IsBrakeActuated` and `App_Brake_IsPressureSensorOpenOrShortCircuit`. Plus you don't have to do all the extra memory allocation/deallocation in `Create` and `Destroy`

boards/FSM/Inc/App/App_SetPeriodicCanSignals.h
```diff
@@ -10,3 +10,6 @@ void App_SetPeriodicSignals_SteeringAngleInRangeCheck(
 void App_SetPeriodicSignals_BrakePressureInRangeCheck(
     const struct FsmWorld *world);
 void App_SetPeriodicSignals_BrakeActuationStatus(const struct FsmWorld *world);
+void App_SetPeriodicSignals_BrakeSensorOpenOrShortCircuitStatus(
+    const struct FsmWorld *world);
```
thekenu, line 14:
```suggestion
```
I don't think any of these exist anymore?

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -138,8 +138,9 @@ SG_ DUMMY : 0|1@1+ (1,0) [0|1] "" DEBUG
 BO_ 311 FSM_STEERING_ANGLE_SENSOR: 4 FSM
 SG_ Steering_Angle : 0|32@1+ (1,0) [-110|110] "deg" DCM
 
-BO_ 312 FSM_BRAKE_PRESSURE_SENSOR: 4 FSM
+BO_ 312 FSM_BRAKE_PRESSURE_SENSOR: 5 FSM
```
thekenu, line 141: I think with the `App_Brake`, we can just combine CAN message 308 and 312 to just `BO_ 308 FSM_BRAKE:`

i.e.
```
BO_ 308 FSM_BRAKE: 8 FSM
SG_ Is_Brake_Actuated ...
SG_ Is_Pressure_Sensor_Open_Or_Short_Circuit ...
SG_ Brake_Pressure ...
```

boards/FSM/Inc/App/App_Brake.h
pasitlaothamatas: I keep making this mistake F

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: ok this is done

pasitlaothamatas:
@garethellis0 @thekenu OK should be close to being done now

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+struct InRangeCheck *
+    App_Brake_GetPressureInRangeCheck(const struct Brake *brake);
+
+/**
+ * Check if the given brake is actuated
+ * @param brake The brake to check if it is actuated
+ * @return True if the brake is actuated, false if it is not
+ */
+bool App_Brake_GetIsBrakeActuated(const struct Brake *brake);
```
thekenu, line 45:
```suggestion
bool App_Brake_IsBrakeActuated(const struct Brake *brake);
```

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Check if the brake pressure sensor is in an open or short circuit state
+ * @param brake The brake to check if the corresponding pressure sensor is
+ * in an open or short circuit state
+ * @return True if the brake is in an open or short circuit state, false if it
+ * is not
+ */
+bool App_Brake_GetIsPressureSensorOpenOrShortCircuit(const struct Brake *brake);
```
thekenu, line 54:
```suggestion
bool App_Brake_IsPressureSensorOpenOrShortCircuit(const struct Brake *brake);
```

pasitlaothamatas: both should be good now

boards/FSM/Inc/App/App_Brake.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+bool App_Brake_IsBrakeActuated(const struct Brake *brake);
+
+/**
+ * Check if the brake pressure sensor is in an open or short circuit state
+ * @param brake The brake to check if the corresponding pressure sensor is
+ * in an open or short circuit state
+ * @return True if the brake is in an open or short circuit state, false if it
+ * is not
+ */
```
thekenu, line 53:
```suggestion
/**
 * Check if the brake pressure sensor is in an open or short circuit state
 * @param brake The brake to check if the corresponding pressure sensor is
 * in an open or short circuit state
 * @return True if the brake pressure sensor is in an open or short circuit state, false if it
 * is not
 */
```

PR #648, opened 2020-06-11
Write App and Io code for 18650s

thekenu:
### Checklist


boards/PDM/Inc/App/App_LowVoltageBatteries.h
```diff
@@ -0,0 +1,60 @@
...
+ * @param has_charge_fault A function that can be called to check if the cell
+ *                         balanacing IC for the low voltage batteries has a
+ *                         fault
+ * @param has_boost_fault A function that can be called to check if the boost
+ *                        controller IC for the low voltage batterise has a
+ *                        fault
+ * @return The created low voltage batteries, whose ownership is given to the
+ *         caller
+ */
+struct LowVoltageBatteries *App_LowVoltageBatteries_Create(
```
thekenu, line 24: `LowVoltageBatteries` represent the two 18650 cells on the car

garethellis0: Just as a general thought, should these be plural if we're effectively looking at them as a single battery?

boards/PDM/Src/App/states/App_AirOpenState.c
```diff
@@ -20,6 +20,8 @@ static void AirOpenStateRunOnTick1Hz(struct StateMachine *const state_machine)
 
 static void AirOpenStateRunOnTick100Hz(struct StateMachine *const state_machine)
 {
+    App_AllStatesRunOnTick100Hz(state_machine);
```
garethellis0, line 23: Side-note; if this "AllStates" pattern is super common, it might be worth modifying `SharedStateMachine` to explicitly support it, ie. having `AllStatesRunOnTick100Hz`, `AllStatesRunOnTick1Hz etc. that are called during the respective "tick" functions for the statemachine.

boards/PDM/Inc/App/App_LowVoltageBatteries.h
thekenu: I was considering this too, let's change it!

boards/PDM/Src/App/states/App_AirOpenState.c
thekenu: Made a ticket for it! 

https://github.com/UBCFormulaElectric/Consolidated-Firmware/issues/649

PR #652, opened 2020-06-13
Split up stateless and all-states specifications

thekenu:
### Summary

In `specs.md`, most of what's considered "stateless" are supposed to run in every state. A better name would be "all states". Furthermore,  some other considered "stateless" _really_ is stateless. That is, they should run outside of the state machine (e.g. the boot startup message).

### Checklist

### DIFF
--- a/specs.md
+++ b/specs.md
@@ -22,26 +22,32 @@ There are two types of faults:
 
 ## Table Of Contents
 - [FSM (Front Sensor Module)](#FSM)
+    - [FSM Stateless](#FSM_STATELESS)
+    - [FSM All States](#FSM_ALL_STATES)
     - [FSM AIR-Open State](#FSM_AIR_OPEN)
     - [FSM AIR-Closed State](#FSM_AIR_CLOSED)
 - [DCM (Drive Control Module)](#DCM)
     - [DCM Stateless](#DCM_STATELESS)
+    - [DCM All States](#DCM_ALL_STATES)
     - [DCM Init State](#DCM_INIT)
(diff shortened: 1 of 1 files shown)


thekenu:
All states means something that should be done in every state. Whereas stateless would be something done outside of the state machine. For example, sending the startup CAN message happens before the state machine starts.

garethellis0:
Right, but could we just everything that's not in a particular state be "stateless"? Just seems like it will be a little tricky to sort specs in the future, and ideally we keep this a simple as possible so that people actually use/update it.

thekenu:
Do you mean that things that run in every state should also be stateless?

My issue with `Stateless` right now is that it represents two types of specs:
- Those that should run in every state
- Those that should run once outside of the state machine

And it just seems misleading to group them together

garethellis0:
I just can't think of many things that are in the first group but not the second group. They're really just specs that should be enforced at all times, instead of in a specific state.

thekenu:
I guess that's a valid way to interpret what Stateless means as well. I was hoping that every `AllStates` specs would directly go in the `RunAllStatesTick` functions we have been adding, while `Stateless` specs wouldn't. 

In any case, I'm not super attached to having this change merged in, so it you think having `Stateless` and `AllStates` would make the specs harder to maintain I'll be happy to close this PR too! :taco:

garethellis0:
Yeah, I really don't think they're worth it, really want to keep the specs as easy to maintain as possible, even at the cost of some organisation.

PR #653, opened 2020-06-14
Export JSON compilation database

thekenu:
### Summary

Export `compile_commands.json` for third party tools like `clangd`. To be frank, I'm doing this largely for my own vim plugin (`CoC`) and I could get away by only modifying my `CMakeLists.txt` locally. But this change won't really affect build time for anyone so I think tracking this in version control is fine.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,7 @@ Pipfile.lock
 CMakeCache.txt
 cmake_install.cmake
 CMakeFiles/
-compile_commands.json
+compile_commands*.json
 Makefile
 auto_generated/
 __lib__/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -77,7 +77,7 @@ Pipfile.lock
 CMakeCache.txt
 cmake_install.cmake
 CMakeFiles/
-compile_commands.json
+compile_commands*.json
```
thekenu, line 80: This is intentional as I ending creating multiple symbolic links with `compile_commands1.json` and `compile_commands2.json` to link with the `compile_commands.json` for `embedded` and `gtest`. Again, this is mostly for my personal use but I think it doesn't obfuscate anything and should be fine.

PR #656, opened 2020-06-15
Update tests to use LetTimePass()

thekenu:
### Summary

Update tests to use LetTimePass(). This way we get to test all tick functions together, which is much more representative of what would happen on the car.

### Checklist


boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -231,7 +233,7 @@ TEST_F(
     {
         SetInitialState(state);
         get_pwm_frequency_fake.return_val = fake_frequency;
-        App_SharedStateMachine_Tick100Hz(state_machine);
+        LetTimePass(state_machine, first_state_machine_tick ? 1 : 10);
```
thekenu, line 236: So this `first_state_machine_tick ? 1 : 10` shows up many times in this PR and let me explain what issue it is addressing.

We have a lot of tests that check for a condition in all states using let's say, the 100hz tick function.

The old code looks something like:
```
all_states = state_A, state_B;

for (state in all_states) {
    SetInitialState(state);
    Tick100Hz();
    ASSERT_EQ();
}
```

If we were to change this to LetTimePass and flatten the for-loop, it would look like
```
SetInitialState(state_A);
LetTimePass(1);
ASSERT_EQ();

SetInitialState(state_B);
LetTimePass(10);
ASSERT_EQ();
```

Long story short, the tricky part is that the 100Hz tick function executes at t = 0, t = 10, t = 20, etc. Due to the zero-indexing nature of this, the first `LetTimePass()` will always be different from the rest (i.e. `LetTimePass(1)`).

The most elegant solution I could think of was  to update `LetTimePass` to set a boolean flag `first_state_machine_tick`. This is a bit hard to explain over text so let me know if it's unclear.

But in any case, I would be very much open to a cleaner solution.

garethellis0: I'm not super clear on how these are equivalent? In the second one it seems like we're ticking more then once, whereas in the first one we only seem to tick once?

thekenu: The first one is in a for-loop, which runs twice?

garethellis0: Oh I think I understand now. Why not just do effectively this inside the `LetTimePass` function?

thekenu: Just to make sure we're on the same page, I imagine the new `LetTimePass` would look like

```
void LetTimePass(u32 time_ms) {
  static bool is_first_tick  = true;

  if (is_first_tick) {
    time_ms = 1; // The first call to this function 
                 // isn't allowed to tick more than once!
    is_first_tick = false;  
  }
  
  for (uint32_t ms = 0; ms < time_ms; ms++) {
    // Same as before
  }
}
```
And when we have consecutive `LetTimePass` calls:
```
LetTimePass(10) // Request 10ms to pass, but since it's the first call we only tick 1 time
LetTimePass(10) // Request 10ms to pass, tick 10 times
LetTimePass(10) // Request 10ms to pass, tick 10 times
```

The changes to `LetTimePass` makes the function name a bit awkward imo, because the first call requests 10ms to pass but only 1 ms ended up passing. Maybe this could be solved with a better function name? What do you think?

garethellis0: Hmm, I'm still a bit confused about why we're ticking like this in the first place. Would it not be sufficient to say that behavior is not defined until we've ticked at least once, and then just always tick for 10ms? 

This might be worth a quick call to hash out.

thekenu:
@garethellis0 Implemented the fix we discussed!

PR #660, opened 2020-06-17
Split critical errors into AIR shutdown and motor shutdown errors

thekenu:
### Summary

The current error table categorizes the errors as either non-critical or critical. However, to better support our specs we should further divide the critical errors as air shutdown and motor shutdown errors. 

### Changelist 

Added the following methods:
```
bool App_SharedError_IsCritical(const struct Error *error);
bool App_SharedError_IsNonCritical(const struct Error *error);
bool App_SharedErrorTable_HasAnyAirShutdownErrorSet(const struct ErrorTable *error_table);
bool App_SharedErrorTable_HasAnyMotorShutdownErrorSet(const struct ErrorTable *error_table);
```
I could've added some other methods such as `GetAllSetAirShutdownErrros` but I don't really think there will ever be a use-case for that so might as well keep the changes to a minimum.

### Checklist


kirons96:
HasAny instead of GetAll, is a good call, i see no use case for GetAll. we should be checking CAN if we have more than 1 critical error

LGTM, aka:
- looks good to me
- let's get this merged
- let's get this money

thekenu:
Let's get this :taco:

PR #662, opened 2020-06-20
Update PDM Ioc

markomarkusi:
### Summary
Updated the PDM's Ioc file to match the new 2020 PDM
image
[Current state of board schematic](https://github.com/UBCFormulaElectric/Consolidated-Hardware/blob/joe_pdm_changes/PDM/Altium/Project%20Outputs/Schematics/PDM.PDF)

### Checklist


markomarkusi:
> * Could we update the Altium net names to match the `.ioc`? I prefer having `FSB_AUX1_AUX2` over `FSB1`. Makes it a bit easier to review the pin mapping in this PR if the net names match with the `.ioc` too.
**Ya I agree, I had just copied the naming scheme of the original PDM but they weren't named the best. Changed**
> * It looks like the MCU I previously picked will be sufficient for the new revision? Just making sure it's clear that we're _not_ switching the MCU
**Yep it's good thx**
> * The CI is failing so that needs some fixing
**Ooops looks like I got rid of the FreeRtos setting by accident. Remade the Ioc should pass now**
> * Not related to firmware but PA12 on the new schematic is connected to CAN_RX, but I think it should be CAN_TX?
**Ya good catch tyty**

thekenu:
I'm just matching the schematic with the screenshot you linked, here are some (possible) mistakes I found:

- PA3 is `PIN_AUX2` on schematic but `PIN_AUX1` in `.ioc`
- PB13 is `SCLK` on schematic but `SCK in `.ioc`. Either name is fine but it'd be nice to just stick with one.

markomarkusi:
> I'm just matching the schematic with the screenshot you linked, here are some (possible) mistakes I found:
> 
> * PA3 is `PIN_AUX2` on schematic but `PIN_AUX1` in `.ioc`
**AUX1 was accidentally also labelled AUX2. Fixed**
> * PB13 is `SCLK` on schematic but `SCK in `.ioc`. Either name is fine but it'd be nice to just stick with one.
> -**Changed to SCK in schematic, since default naming in cube is sck**
> Also is the `CLOCK` pin from the efuse deliberately left disconnected, unlike the last revision?
-**No use case for use since none of our outputs are modulated by PWM**

markomarkusi:
@thekenu should be good now

thekenu:
The pin mapping looks good now, but looks like CI is still failing? Also, since the OV pin is removed, can we update the specs.md as well?

markomarkusi:
> The pin mapping looks good now, but looks like CI is still failing? Also, since the OV pin is removed, can we update the specs.md as well?

Passing now, I assumed it would pass last time so I forgot to check after it was done. Updated the Specs as well

PR #670, opened 2020-09-01
Add a Shared Wait Signal

pasitlaothamatas:
### Summary

- Added a shared wait signal to deal with situations like DCM-12
- Add unit tests to check that the buzzer is turned on for 2 seconds after entering the drive state to check that the wait signal is operating correctly

### Changelist 

- Added a flag for the buzzer struct to indicate when the buzzer has been turned on/off

### Testing Done

 - Create unit tests for the wait signal in `Test_SharedWaitSignal.cpp`
- Turn the buzzer on in `DriveStateRunOnEntry` in `App_DriveState`
- Test that the wait signal is working properly by checking that the buzzer is turned on for two seconds after entering the drive state

### Checklist


boards/DCM/Src/App/App_DcmWorld.c
```diff
@@ -13,6 +14,7 @@ struct DcmWorld
     struct BrakeLight *       brake_light;
     struct Buzzer *           buzzer;
     struct ErrorTable *       error_table;
+    struct WaitSignal *       buzzer_wait_signal;
```
pasitlaothamatas, line 17: will add a LL library if required. This should suffice for now

boards/shared/Test/Src/Test_SharedWaitSignal.cpp
```diff
@@ -0,0 +1,250 @@
+#include "Test_Shared.h"
+
+extern "C"
+{
+#include "App_SharedWaitSignal.h"
+}
+
+// TODO: Refactor fake function names after 668 is merged
```
markomarkusi, line 8: :eyes: :eyes: :eyes: :eyes: :eyes: :eyes:

boards/DCM/Inc/App/App_DcmWorld.h
```diff
@@ -28,7 +30,10 @@ struct DcmWorld *App_DcmWorld_Create(
     struct BrakeLight *       brake_light,
     struct Buzzer *           buzzer,
     struct ErrorTable *       error_table,
-    struct Clock *            clock);
+    struct Clock *            clock,
+
+    bool (*is_buzzer_on)(struct DcmWorld *),
+    void (*buzzer_callback)(struct DcmWorld *));
```
markomarkusi, line 36: LGTM did you just want to change the order here as a NIT?

markomarkusi (approved):

boards/shared/Test/Src/Test_SharedWaitSignal.cpp
pasitlaothamatas: oh whoops forgot to delete this

boards/DCM/Inc/App/App_DcmWorld.h
pasitlaothamatas: Do we currently have a convention for this?

markomarkusi: No i don't think so it just looks nicer to group the buzzer stuff together than have part of it after the clock structure

boards/DCM/Src/App/App_DcmWorld.c
```diff
@@ -24,9 +26,12 @@ struct DcmWorld *App_DcmWorld_Create(
...
     struct Buzzer *const            buzzer,
     struct ErrorTable *const        error_table,
-    struct Clock *const             clock)
+    struct Clock *const             clock,
+
+    bool (*is_buzzer_on)(struct DcmWorld *),
+    void (*buzzer_complete_callback)(struct DcmWorld *))
 {
-    struct DcmWorld *world = (struct DcmWorld *)malloc(sizeof(struct DcmWorld));
+    struct DcmWorld *world = malloc(sizeof(struct DcmWorld));
```
pasitlaothamatas, line 34: Got rid of casting here https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

boards/DCM/Inc/App/App_DcmWorld.h
pasitlaothamatas: I prefer this cuz it splits the members in world from the callback related signals

pasitlaothamatas:
This was approved but should prob review a couple of things before I merge. Ready for another round.

boards/shared/Test/Src/Test_SharedWaitSignal.cpp
pasitlaothamatas: This should be deleted

Issue #434, opened 2019-09-07
Create A Shared SPI Library


garethellis0:
### Summary
We need a shared library for talking to external chips over SPI. One example of this is the BMS talking to the Efuse chip (a `NXP 22XS4200`)

### Useful Resources
- example of SPI communication between two boards with STM32F3 microcontrollers (we use the F3 in all our boards currently):https://github.com/eleciawhite/STM32Cube/tree/master/STM32Cube_FW_F3_V1.3.0/Projects/STM32F3-Discovery/Examples/SPI/SPI_FullDuplex_ComDMA 

### Deliverables
- [ ] A new folder, `boards/shared/SPI` which should contain `SharedSPI.h` and `SharedSPI.c` (see `boards/shared/GPIO` for an example of the format of the files). These files should contain functions that allow us to initialize/configure and read/write to/from SPI.
- [ ] Test that your code works by reading/writing values to/from the an efuse chip (a `NXP 22XS4200`) on the BMS

PR #671, opened 2020-09-27
Add a shared spi driver

pasitlaothamatas:
### Summary

- Added a shared spi driver (blocking mode).

### Testing Done

- Tested all of the functions in the driver with LTC6813 chip.

### Resolved Issues

#434 

### Checklist

### DIFF
--- a/boards/shared/Inc/Io/Io_SharedSpi.h
+++ b/boards/shared/Inc/Io/Io_SharedSpi.h
@@ -10,8 +10,8 @@ struct SharedSpi;
  * @param spi_handle The handle of the given SPI interface.
  * @param nss_port Pointer to the GPIO port of the given SPI interface.
  * @param nss_pin The GPIO pin of the given SPI interface.
- * @param delay_ms The delay in milliseconds during which a blocking process
- * should wait until an error is returned.
+ * @param delay_ms The delay in milliseconds during a SPI transaction which a
+ * blocking process should wait until an error is returned.
  * @return Pointer to the allocated and initialized SPI interface
  */
 struct SharedSpi *Io_SharedSpi_Create(
@@ -40,7 +40,7 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *spi);
  * interface.
  * @param tx_size The size of the data transmitted to the device connected to
  * the SPI interface.
- * @param rx_data The data received from the device connected ot the SPI
+ * @param rx_data The data received from the device connected to the SPI
  * interface.
  * @param rx_size The size of the data received from the device connected to the
  * SPI interface.
--- a/boards/shared/Inc/Io/Io_SharedSpi.h
+++ b/boards/shared/Inc/Io/Io_SharedSpi.h
@@ -1,0 +1,115 @@
+#pragma once
+
+#include <stm32f3xx_hal.h>
+#include <stm32f3xx_hal_spi.h>
+
+struct SharedSpi;
+
+/**
+ * Allocate and initialize a SPI interface.
+ * @param spi_handle The handle of the given SPI interface.
+ * @param nss_port Pointer to the GPIO port of the given SPI interface.
+ * @param nss_pin The GPIO pin of the given SPI interface.
+ * @param delay_ms The delay in milliseconds during which a blocking process
+ * should wait until an error is returned.
+ * @return Pointer to the allocated and initialized SPI interface
+ */
+struct SharedSpi *Io_SharedSpi_Create(
+    SPI_HandleTypeDef *spi_handle,
+    GPIO_TypeDef *     nss_port,
+    uint16_t           nss_pin,
+    uint32_t           delay_ms);
+
+/**
+ * Set the NSS pin low for the given SPI interface.
+ * @param spi The given SPI interface.
+ */
+void Io_SharedSpi_SetNssLow(const struct SharedSpi *spi);
+
+/**
+ * Set the NSS pin high for the given SPI interface.
+ * @param spi The given SPI interface.
+ */
+void Io_SharedSpi_SetNssHigh(const struct SharedSpi *spi);
+
+/**
+ * Transmit data to and receive data from the device connected to the given SPI
+ * interface.
+ * @param spi The given SPI interface.
+ * @param tx_data The data transmitted to the device connected to the SPI
+ * interface.
+ * @param tx_size The size of the data transmitted to the device connected to
+ * the SPI interface.
+ * @param rx_data The data received from the device connected ot the SPI
+ * interface.
+ * @param rx_size The size of the data received from the device connected to the
+ * SPI interface.
+ * @return The HAL status of the data transmission and reception.
+ */
+HAL_StatusTypeDef Io_SharedSpi_TransmitAndReceive(
+    const struct SharedSpi *spi,
+    uint8_t *               tx_data,
+    uint16_t                tx_size,
+    uint8_t *               rx_data,
+    uint16_t                rx_size);
+
+/**
+ * Transmit data to the device connected to the given SPI interface.
+ * @param spi The given SPI interface.
+ * @param tx_data The data transmitted to the device connected to the SPI
+ * interface.
+ * @param tx_size The size of the data transmitted to the device connected to
+ * the SPI interface.
+ * @return The HAL status of the data transmission.
+ */
+HAL_StatusTypeDef Io_SharedSpi_Transmit(
+    const struct SharedSpi *spi,
+    uint8_t *               tx_data,
+    uint16_t                tx_size);
+
+/**
+ * Receive data from the device connected to the given SPI interface.
+ * @param spi The given SPI interface.
+ * @param rx_data The data received from the device connected to the SPI
+ * interface.
+ * @param rx_size The size of the data received from the device connected to the
+ * SPI interface.
+ * @return The HAL status of the data reception.
+ */
+HAL_StatusTypeDef Io_SharedSpi_Receive(
+    const struct SharedSpi *spi,
+    uint8_t *               rx_data,
+    uint16_t                rx_size);
+
+/**
+ * Transmit multiple copies of the a data packet to the device connected to the
+ * given SPI interface without toggling the NSS pin.
+ * @param spi The given SPI interface.
+ * @param tx_data The data transmitted to the device connected to the SPI
+ * interface.
+ * @param tx_size The size of the data transmitted to the device connected to
+ * the SPI interface.
+ * @param num_tx_data_copies The number of copies of tx_data transmitted to the
+ * device connected to the SPI interface.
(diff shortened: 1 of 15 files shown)


boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -0,0 +1,114 @@
...
+{
+    Io_SharedSpi_SetNssLow(spi);
+    HAL_StatusTypeDef status =
+        HAL_SPI_Receive(spi->hspi, rx_data, rx_size, 100U);
+    Io_SharedSpi_SetNssHigh(spi);
+
+    return status;
+}
+
+HAL_StatusTypeDef Io_SharedSpi_MultipleTransmitWithoutNssToggle(
```
pasitlaothamatas, line 92: Functions that end with `WithoutNssToggle` can be used to create a more customized approach to communicating with devices connected to the SPI interface.

pasitlaothamatas: There are cases for particular ICs where the 'standard' functions for SPI communication provided above won't work

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -0,0 +1,114 @@
...
+HAL_StatusTypeDef Io_SharedSpi_TransmitAndReceive(
+    const struct SharedSpi *const spi,
+    uint8_t *                     tx_data,
+    uint16_t                      tx_size,
+    uint8_t *                     rx_data,
+    uint16_t                      rx_size)
+
+{
+    Io_SharedSpi_SetNssLow(spi);
+    RETURN_IF_STATUS_NOT_OK(
```
pasitlaothamatas, line 55: Felt I was repeating code here when I'm checking for the HAL status returned after each SPI transmission/reception so I decided to use a macro here

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -0,0 +1,114 @@
...
+{
+    for (size_t i = 0U; i < num_tx_data_copies; i++)
+    {
+        RETURN_IF_STATUS_NOT_OK(
+            HAL_SPI_Transmit(spi->hspi, tx_data, tx_size, 100U), spi->nss_port,
+            spi->nss_pin)
+    }
+
+    return HAL_OK;
+}
```
markomarkusi, line 106: even though you are not toggling between each sequential spi data frame, shouldn't the nss pin be brought low before transmitting the first frame and then brought high after the last?

markomarkusi: also this for loop would not work you are not making use of the counter variable in you HAL_SPI_Transmit() command

```
for (size_t i = 0U; i < num_tx_data_copies; i++)
    {
        RETURN_IF_STATUS_NOT_OK(
            HAL_SPI_Transmit(spi->hspi, tx_data + i, tx_size, 100U), spi->nss_port,
            spi->nss_pin)
    }

```
or simply you can remove the for loop altogether:
  HAL_SPI_Transmit(spi->hspi, tx_data, num_tx_data_copies, 100U), spi->nss_port,
            spi->nss_pin)

pasitlaothamatas: `tx_size` is the size of the `tx_data` buffer you are trying to send. This function is used to send the same `tx_data` multiple times. This ends up sending the `tx_buffer[tx_size]` `num_tx_data_copies` times

pasitlaothamatas: I specified specifically in the name that we won't be toggling the NSS pin. Here's an example of where this function would be used. 

https://github.com/pasitlaothamatas/Consolidated-Firmware-1/blob/bb58d43bf417f6836aef7603af261d82d800c515/boards/BMS/Src/Io/Io_LTC6813.c#L271

pasitlaothamatas: Cases like this are IC specific. You probably won't see data being sent like this to other SPI chips. Since we're writing a shared SPI driver I thought it was a better idea to include code that the user can use to create custom functions for sending out packets to a SPI chip.

image

pasitlaothamatas:
@markomarkusi addressed comments 🐧🐧🐧🐧🐧

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -0,0 +1,117 @@
...
+HAL_StatusTypeDef Io_SharedSpi_MultipleTransmitWithoutNssToggle(
+    const struct SharedSpi *const spi,
+    uint8_t *                     tx_data,
+    uint16_t                      tx_size,
+    size_t                        num_tx_data_copies)
+{
+    HAL_StatusTypeDef status;
+    for (size_t i = 0U; i < num_tx_data_copies; i++)
+    {
+        status = HAL_SPI_Transmit(spi->hspi, tx_data, tx_size, 100U);
```
garethellis0, line 101: what is `100U` here? Is that a timeout? I would try to avoid doing operations like this synchronously, as if this started timing out, even just a few calls to this function could run over the time limit for an RTOS task.

pasitlaothamatas: Okay that makes sense. Were you thinking of moving away from blocking functions in general? Or were you thinking of just leaving the timeout duration as 0? @garethellis0

pasitlaothamatas: I was testing HAL_SPI_Transmit_DMA. For the number of packages that I was dealing this it didn't help much in terms of performance. Also ran into this [link](https://community.st.com/s/question/0D50X00009sW5Ld/halspitransmitit-blocks) cuz I ran into a similar problem with HAL_SPI_DMA.

pasitlaothamatas:
@garethellis0 Addressed some comments 🐧🐧🐧

scripts/travis_ci/series_ci.sh
```diff
@@ -37,4 +37,4 @@ curl \
   --header "Authorization: Token 27d62d1f-71c6-4075-95b5-f7aec56d0204" \
   --header "Content-Type: application/json" \
   --data "{\"values\":[{\"value\":\"$ram\",\"line\":\"RAM\"},{\"value\":\"$flash\",\"line\":\"FLASH\"}],\"sha\":\"$SHA\"}" \
-  https://seriesci.com/api/UBCFormulaElectric/Consolidated-Firmware/$SERIES_NAME/many
+  -k https://seriesci.com/api/UBCFormulaElectric/Consolidated-Firmware/$SERIES_NAME/many
```
pasitlaothamatas, line 40: I was getting this error message while building in travis ci: 
`curl: (60) SSL certificate problem: certificate has expired .sh`

[This](https://stackoverflow.com/questions/18964175/how-to-fix-curl-60-ssl-certificate-invalid-certificate-chain) seemed to fix it

markomarkusi: LGTM lets get this 🍞

markomarkusi: * Lets get this 🍆

boards/shared/Src/Io/Io_SharedSpi.c
markomarkusi: I'm fine with having this in shared

scripts/travis_ci/series_ci.sh
pasitlaothamatas: PLEASE APPROVE PLEASE APPROVE

boards/shared/Inc/Io/Io_SharedSpi.h
```diff
@@ -0,0 +1,115 @@
...
+void Io_SharedSpi_SetNssHigh(const struct SharedSpi *spi);
+
+/**
+ * Transmit data to and receive data from the device connected to the given SPI
+ * interface.
+ * @param spi The given SPI interface.
+ * @param tx_data The data transmitted to the device connected to the SPI
+ * interface.
+ * @param tx_size The size of the data transmitted to the device connected to
+ * the SPI interface.
```
markomarkusi, line 42: So this is actually poorly named in the HAL_SPI library, it does not represent the size of the data but instead the amount of SPI messages/packets/frames to transmit/receive. Size implies bytes. Passing a _1U_  to this function would mean send/receive one SPI data frame which is of the size that the SPI peripheral is configured in (could be either 8bits or 16bits or less)

scripts/travis_ci/series_ci.sh
markomarkusi: Added a comment actually

markomarkusi: NIT NIT

boards/shared/Inc/Io/Io_SharedSpi.h
pasitlaothamatas: Oh okay good point. I was just going by what was in the HAL library but sometimes that may not be the best idea lol

markomarkusi: Ya I found the HAL naming very confusing and so do a lot of people online, I had to go through the driver to figure out what this was actually doing and what is the correct value to pass to that parameter

pasitlaothamatas: On a second though I guess it is referring to the size of the tx data or the rx buffer?

tx_data[tx_size]
rx_data[rx_size]

markomarkusi: hmmm I see should we re-word it to say that tx_size is the number of elements in the tx_data array ?

pasitlaothamatas: Yup I can do that

boards/shared/Src/Io/Io_SharedSpi.c
pasitlaothamatas: Hmm even if other IC's don't use it? I know the efuse chip doesn't exactly follow this pattern when its communicating with other devices right

boards/shared/Inc/Io/Io_SharedSpi.h
```diff
@@ -106,5 +106,5 @@ HAL_StatusTypeDef Io_SharedSpi_MultipleTransmitWithoutNssToggle(
  */
 HAL_StatusTypeDef Io_SharedSpi_TransmitWithoutNssToggle(
     const struct SharedSpi *spi,
-    uint8_t *               tx_data,
+    uint8_t *               tx_buffer,
```
pasitlaothamatas, line 109: Should I be calling tx_data tx_buffer? @markomarkusi

markomarkusi: ya buffer sounds good!

markomarkusi (approved):

PR #672, opened 2020-09-28
Add Io code for LTC6813 Cell Voltage Measurement

pasitlaothamatas:
### Summary

- Add Io code for the LTC6813 chip for cell voltage measurement. 

### Testing Done

- Tested code with the DC2350A-B boards
- Tested that code works with the boards in a daisy chained configuration

### Checklist


pasitlaothamatas:
Merging will be blocked #671 is merged. Feel free to review rn though.

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,368 @@
...
+    {
+        if (Io_SharedSpi_TransmitAndReceive(
+                ltc_6813.spi, tx_cmd, NUM_OF_CMD_BYTES, &rx_data, 1U) != HAL_OK)
+        {
+            return EXIT_CODE_UNIMPLEMENTED;
+        }
+
+        ++adc_conversion_timeout_counter;
+
+        // Timeout counter threshold of 10 was chosen arbitrarily.
```
pasitlaothamatas, line 216: Tested this function on a daisy chain containing two LTC6813 ICs. Usually takes one cycle in here to finish. The function does not return EXIT_CODE_TIMEOUT unless the power supply has been disconnected from the DC2350A-B board.

pasitlaothamatas: I'll set an appropriate timeout counter threshold when I test this code with the segment. There are six LTC6813's on the daisy chain so the ADC conversion might take a little longer to complete.

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,368 @@
...
+void Io_LTC6813_Init(
+    SPI_HandleTypeDef *hspi,
+    GPIO_TypeDef *     nss_port,
+    uint16_t           nss_pin)
+{
+    assert(hspi != NULL);
+
+    ltc_6813.spi                 = Io_SharedSpi_Create(hspi, nss_port, nss_pin);
+    ltc_6813.pec15_error_counter = 0U;
+    memset(
```
pasitlaothamatas, line 274: I remember Ken telling me its always good practice to initialize the members of the struct. Since the `ltc_6813` struct is static, all values in cell_voltages array should already be set to zero. Is this necessary?

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+    size_t   current_ic,
+    size_t   current_register_group,
+    uint8_t *rx_cell_voltages)
+{
+    size_t cell_voltage_index = current_ic * NUM_OF_CELL_VOLTAGE_RX_BYTES;
+
+    for (size_t current_cell = 0U;
+         current_cell < NUM_OF_CELLS_PER_LTC6813_REGISTER_GROUP; current_cell++)
+    {
+        assert(
```
pasitlaothamatas, line 239: @garethellis0  Not sure if this is the best way to do it but I wanted to prevent access to unitialized memory locations. If there's a better way of doing this please let me know

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+        }
+    }
+
+    return EXIT_CODE_OK;
+}
+
+static void Io_LTC6813_ParseCellsAndPerformPec15Check(
+    size_t   current_ic,
+    size_t   current_register_group,
+    uint8_t *rx_cell_voltages)
```
pasitlaothamatas, line 232: @garethellis0 should I also do `uint8_t rx_cell_voltages[EXPECTED_SIZE]` just to specify the size of `rx_cell_voltages` we're dealing with here? Just wondering what the best practice would be here..

boards/BMS/Inc/Io/Io_LTC6813.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <stm32f3xx_hal.h>
+#include "App_SharedExitCode.h"
+
+/**
+ * Initialize the LTC6813 daisy chain.
+ * @param hspi The SPI handle for the LTC6813 daisy chain.
```
garethellis0, line 8: Could we call this `spi_handle` instead? Seems a bit more descriptive.

boards/BMS/Inc/Io/Io_LTC6813.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Read all cell register groups for each chip on the daisy chain.
+ * @return EXIT_CODE_OK if no PEC15 mismatches occur. Else, return
+ * EXIT_CODE_ERROR.
+ */
+ExitCode Io_LTC6813_ReadAllCellRegisterGroups(void);
+
+/**
+ * Get the pointer to the double array containing converted cell voltages for
```
garethellis0, line 31: this returns a pointer to a `uint16_t` array, not to a `double`? Or is this like a 2 by `n` array? Either way, we probably need to update this comment (and below).

boards/BMS/Inc/Io/Io_LTC6813.h
```diff
@@ -0,0 +1,36 @@
...
+ * each chip on the daisy chain.
+ * @return The pointer to the double array containing converted chip cell
+ * voltages.
```
garethellis0, line 34: How long is this array? Is the caller responsible for de-allocating it?

garethellis0: We probably need to return a number of cells as well if it can change. Otherwise we should specify a constant.

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <stdbool.h>
+#include "Io_LTC6813.h"
+#include "Io_SharedSpi.h"
+#include "configs/Io_LTC6813Configs.h"
+
+enum CellVoltageRegisterGroups
+{
+    CELL_VOLTAGE_REGISTER_GROUP_A = 0,
```
garethellis0, line 11: I think this would be `0` anyhow. Any particular reason we specified a value for only the first entry in the enum?

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+    CELL_VOLTAGE_REGISTER_GROUP_E,
+    CELL_VOLTAGE_REGISTER_GROUP_F,
+    NUM_OF_CELL_VOLTAGE_REGISTER_GROUPS
+};
+
+struct LookupTables
+{
+    // Commands used to read cell voltage register groups.
+    uint16_t cell_voltage_register_group_commands
+        [NUM_OF_CELL_VOLTAGE_REGISTER_GROUPS];
```
garethellis0, line 24: These are `uint16_t`'s, how should they be interpreted as commands?

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * Check if all LTC6813 chips in the daisy chain have all completed converting
+ * analogue cell voltages to digital voltages.
+ * @return EXIT_CODE_OK if all LTC6813 chips on the daisy chain have completed
+ * ADC conversions. EXIT_CODE_TIMEOUT if ADC conversions could not be completed
+ * before timing out. EXIT_CODE_UNIMPLEMENTED if the command sent and received
+ * to check the status of ADC conversions was not transmitted or received
+ * successfully.
+ */
+static ExitCode Io_LTC6813_PollAdcConversion(void);
```
garethellis0, line 141: Why are all these functions `static`? I don't think it should matter?

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+        if (Io_SharedSpi_TransmitAndReceive(
+                ltc_6813.spi, tx_cmd, NUM_OF_CMD_BYTES, &rx_data, 1U) != HAL_OK)
+        {
+            return EXIT_CODE_UNIMPLEMENTED;
+        }
+
+        ++adc_conversion_timeout_counter;
+
+        // Timeout counter threshold of 10 was chosen arbitrarily.
+        if (adc_conversion_timeout_counter >= 10U)
```
garethellis0, line 220: Should probably make `10U` here a constant.

boards/BMS/Src/Io/Io_LTC6813.c
garethellis0: That seems like a good idea.

boards/BMS/Src/Io/Io_LTC6813.c
garethellis0: I think this is fine.

boards/BMS/Src/Io/Io_LTC6813.c
garethellis0: I think it's good practice. This way it's a habit to get into and standard across the codebase.

boards/BMS/Src/Io/Io_LTC6813.c
```diff
@@ -0,0 +1,374 @@
...
+    uint8_t  rx_cell_voltages[NUM_OF_CELL_VOLTAGE_RX_BYTES * NUM_OF_LTC6813] = {
+        0
+    };
+
+    // Reset the value of the PEC15 error counter at the start of each cycle.
+    ltc_6813.pec15_error_counter = 0U;
+
+    RETURN_IF_EXIT_NOT_OK(Io_LTC6813_EnterReadyState())
+    RETURN_IF_EXIT_NOT_OK(Io_LTC6813_StartADCConversion())
+    RETURN_IF_EXIT_NOT_OK(Io_LTC6813_PollAdcConversion())
```
garethellis0, line 337: This is a neat little macro :)

garethellis0 (dismissed):
Looking good! This all makes sense and sounds like you've done the hardware testing.

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: I made them static just because the user isn't expected to be calling these functions in the app layer. `Io_LTC6813_ReadAllCellRegisterGroups` is the only function that would use the static functions (at least for now).

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: Sure that sounds like a good idea

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: Oh I think I based this off of Ken's code somewhere. Thought I was following some sort of convention for enums. I'll get rid of it.

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: This is used in `Io_LTC6813_ReadAllCellRegisterGroups`. Basically a command that is used to read from a cell voltage register group. `cell_register_group_cmd` is split into two bytes and transmitted to the IC.

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: Oh I should have been more clear with this. I meant 2D array. Will update the comment.

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: This should be referring to the `cell_voltages` 2D array in the LTC6813 struct.

pasitlaothamatas: When you say specify a constant do you mean something like [this](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/672/files/b757ab004a8f6b3a243ff1a06a49d04c5c5e09cd#r497274956) in the javadoc?

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: Ok sure I can do that

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: Shout out to @markomarkusi

pasitlaothamatas: At this point we might have to call him Macro Markusi

boards/BMS/Src/Io/Io_LTC6813.c
markomarkusi: might as well add static as well 👀 https://hamberg.no/erlend/posts/2013-02-18-static-array-indices.html

pasitlaothamatas: Oh nice I didn't know this was a thing

markomarkusi: me neither lol :neckbeard:

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: This should be renamed now

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: This should be fixed now

boards/BMS/Src/Io/Io_LTC6813.c
pasitlaothamatas: Turned this to a constant

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: @garethellis0 There's a constant that I used to define the size of the array. Did you want me to add the constant to the javadoc?

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: Updated javadoc!

pasitlaothamatas:
@markomarkusi @garethellis0 OK. Went though and addressed the comments. Should be ready for another round of reviews.

boards/BMS/Inc/Io/Io_LTC6813.h
pasitlaothamatas: I added it to the javadoc anyway since I think this would be good information for a user glancing through the .h file

boards/BMS/Src/Io/Io_LTC6813.c
garethellis0: It's probably worth having a comment saying basically this here so that people in the future can understand.

boards/shared/Inc/App/App_SharedExitCode.h
```diff
@@ -5,8 +5,14 @@ typedef enum
...
+    EXIT_CODE_ERROR,
     EXIT_CODE_UNIMPLEMENTED,
     NUM_EXIT_CODES,
 } ExitCode;
 
 #define EXIT_OK(code) (EXIT_CODE_OK == (code))
+
+#define RETURN_IF_EXIT_NOT_OK(code) \
+    if ((code) != EXIT_CODE_OK)     \
+        return (code);
```
markomarkusi, line 18: ❤️

scripts/travis_ci/series_ci.sh
```diff
@@ -37,4 +37,4 @@ curl \
   --header "Authorization: Token 27d62d1f-71c6-4075-95b5-f7aec56d0204" \
   --header "Content-Type: application/json" \
   --data "{\"values\":[{\"value\":\"$ram\",\"line\":\"RAM\"},{\"value\":\"$flash\",\"line\":\"FLASH\"}],\"sha\":\"$SHA\"}" \
-  -k https://seriesci.com/api/UBCFormulaElectric/Consolidated-Firmware/$SERIES_NAME/many
+  https://seriesci.com/api/UBCFormulaElectric/Consolidated-Firmware/$SERIES_NAME/many
```
markomarkusi, line 40: :neckbeard:

markomarkusi (approved):

PR #676, opened 2020-10-12
Add App code for cell voltage monitoring

pasitlaothamatas:
### Summary

- Added app code for cell voltage monitoring (waiting for #672 to get closed, so ignore the Io code in this file for now)
- Split cell voltage monitoring into `App_CellVoltages` and `App_CellMonitoring` 
- `App_CellVoltages` consists of functions used to calculate segment, pack, average cell, min, and max voltages.
- `App_CellMonitoring` consists of in-range checks to see if the cell voltages have overflowed or underflowed

### Testing Done

- Validated in hardware that if a single cell voltage drops below 3.0V a transition into fault state occurs. I cannot test the cell voltage overflow case as the power supply I'm currently working with can only source ~60V max. 

### Checklist


scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -63,6 +65,36 @@ SG_ AIR_NEGATIVE : 1|1@1+ (1,0) [0|1] "" FSM,PDM
...
+
+BO_ 117 BMS_SEGMENT_2_AND_3: 8 BMS
+SG_ SEGMENT_2_VOLTAGE : 0|32@1+ (1,0) [48.0|67.2] "" DEBUG
+SG_ SEGMENT_3_VOLTAGE : 32|32@1+ (1,0) [48.0|67.2] "" DEBUG
+
+BO_ 118 BMS_SEGMENT_4_AND_5: 8 BMS
+SG_ SEGMENT_4_VOLTAGE : 0|32@1+ (1,0) [48.0|67.2] "" DEBUG
+SG_ SEGMENT_5_VOLTAGE : 32|32@1+ (1,0) [48.0|67.2] "" DEBUG
+
+BO_ 119 BMS_CELL_VOLTAGE_ERRORS: 8 BMS
```
pasitlaothamatas, line 88: I wanted to add these signals since I thought they would be helpful for debugging. Not sure what to name these errors though.

pasitlaothamatas: Since the min and max cell voltages is already being used to set AIR_SHUTDOWN errors. I guess these could be labeled as non critical errors? I think it may slightly misleading since having a segment, pack or average cell that is under/overvoltage is a critical error. Open to ideas here.

pasitlaothamatas:
@markomarkusi this is ready for review

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: @kirons96

kirons96: under/over voltage has to be critical faults, even if we misread it, because of the fire danger

how would you distinguish between misreading and an  actual over/undervoltage event?

maybe by saving the past cell voltages at 1Hz and if you see over/undervoltage, make sure it's not by an excessive amount (eg. 1V difference)?

pasitlaothamatas: Oh good point. Right now if the voltage is detected to be under 3.0V or above 4.2V then we transition into the fault state but it currently depends on the value of the min and max cell voltage read at that time. This would happen every second though. 

Since we are measuring values every 10ms I'm thinking we can use a moving average here for the max/min cell voltage values?

pasitlaothamatas: Original question was more about asking what to name these signals though (CAN signals that aren't min/max cell voltages)

kirons96: there's already an RC filter in hardware, so we should be careful about not filtering too aggressively in firmware

if we have a firmware filter, then we probably don't need cell voltage error detection + this CAN message

pasitlaothamatas:
I'm gonna put this on hold for now. Lets get the die temp PR merged in before doing anything with the app code since I've refactored the Io layer quite a bit. @markomarkusi

scripts/codegen/CAN/App_CanMsgs.dbc
kirons96: i mean it's a legit solution to the problem of a bad reading, but those are highly improbable. the hardware filter won't fix a single/double bit error right, assuming it either gets past the CRC (lol) or the error occurs at the MCU (?)

not sure if this problem is even worth considering - take some time to think about how you'd actually get a bad reading and what the probabilities are

pasitlaothamatas:
Refactored IO code so this is ready for review again

boards/BMS/Src/App/App_CellVoltages.c
```diff
@@ -0,0 +1,129 @@
+#include <stddef.h>
+#include "App_CellVoltages.h"
+
+#include "configs/App_CellConfigs.h"
+
+struct CellVoltages
+{
+    uint16_t *measured_cells;
+    uint32_t  total_num_of_cells;
+    uint32_t  num_of_cells_per_segment;
```
markomarkusi, line 10: I've notices in some places you use size_t some places you use uint32_t, why not here as well then?

boards/BMS/Src/App/App_CellVoltages.c
```diff
@@ -0,0 +1,129 @@
...
+
+static float App_SumOfArrayElements(uint16_t *array, uint32_t size)
+{
+    uint32_t array_sum = 0U;
+    for (size_t i = 0U; i < size; i++)
+    {
+        array_sum += array[i];
+    }
+    return (float)array_sum;
+}
```
markomarkusi, line 31:
```suggestion
/**
 * A function used to compute the sum of array elements.
 * @param array A pointer to the given array.
 * @param length The length of the given array.
 * @return The sum of elements for the given array.
 */
static float App_SumOfArrayElements(uint16_t *array, size_t length);

static float App_SumOfArrayElements(uint16_t *array, size_t length)
{
    uint32_t array_sum = 0U;
    for (size_t i = 0U; i < length; i++)
    {
        array_sum += array[i];
    }
    return (float)array_sum;
}
```

boards/BMS/Src/App/App_CellVoltages.c
```diff
@@ -0,0 +1,129 @@
...
+    {
+        uint16_t current_cell_voltage =
+            cell_voltages.measured_cells[current_cell];
+        if (min_cell_voltage > current_cell_voltage)
+        {
+            min_cell_voltage = current_cell_voltage;
+        }
+    }
+
+    return min_cell_voltage / 10000.0f;
```
markomarkusi, line 59: what is 10000 👀 ?

boards/BMS/Src/App/App_CellVoltages.c
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    return (float)max_cell_voltage / 10000.0f;
+}
+
+float App_CellVoltages_GetPackVoltage(void)
+{
+    return (float)App_SumOfArrayElements(
+               cell_voltages.measured_cells, cell_voltages.total_num_of_cells) /
```
markomarkusi, line 81: does this cast do anything since App_SumOfArrayElements already returns a float:
` return (float)array_sum;`

boards/BMS/Src/stm32f3xx_hal_timebase_tim.c
```diff
@@ -1,21 +1,4 @@
...
- * <h2><center>&copy; Copyright (c) 2020 STMicroelectronics.
- * All rights reserved.</center></h2>
- *
- * This software component is licensed by ST under Ultimate Liberty license
- * SLA0044, the "License"; You may not use this file except in compliance with
- * the License. You may obtain a copy of the License at:
- *                             www.st.com/SLA0044
- *
- ******************************************************************************
- */
```
markomarkusi, line 18: interesting interesting

boards/BMS/Src/App/App_CellVoltages.c
pasitlaothamatas: This should be fixed now

boards/BMS/Src/App/App_CellVoltages.c
pasitlaothamatas: Should be converting raw voltages read back from the chip to V. The raw voltages are read back in units of 100uV so we would have to divide by 10000.0 to convert it into voltages. I can add a comment in the .h file for this.

boards/BMS/Src/App/App_CellVoltages.c
pasitlaothamatas: Changed `App_SumOfArrayElements` to return a uint32_t so this cast would be necessary.

boards/BMS/Src/App/App_CellVoltages.c
pasitlaothamatas: Added a comment in the .h file

boards/BMS/Src/stm32f3xx_hal_timebase_tim.c
kirons96: * <h2><center>&copy; Copyright (c) 2020 Pasit Laothamatas.
 * All rights reserved.</center></h2>

boards/BMS/Inc/App/App_CellVoltages.h
```diff
@@ -0,0 +1,76 @@
...
+#include <stdint.h>
+
+/**
+ * Initialize a set of functions used to calculate voltages.
+ * @param get_cell_voltages A pointer to a 2D array containing raw cell voltages
+ * measured by a daisy chain of cell monitoring chips.
+ * @param num_of_cells_per_segment The number of cell voltage measurements per
+ * segment.
+ * @note The raw cell voltages are read back in 100µV. Divide the raw voltages
+ * by 10000 to compute the voltage in V.
```
markomarkusi, line 12: 👀

markomarkusi: Noice do you think a #define would be nice as well for the conversion factor or nah?

boards/BMS/Src/App/App_CellVoltages.c
markomarkusi: icic 👀

boards/BMS/Src/App/App_CellVoltages.c
```diff
@@ -0,0 +1,132 @@
+#include <stddef.h>
+#include "App_CellVoltages.h"
+
+#include "configs/App_CellConfigs.h"
+
+struct CellVoltages
+{
+    uint16_t *raw_cell_voltages;
+    size_t    total_num_of_cells;
+    size_t    num_of_cells_per_segment;
```
markomarkusi, line 10: 👀

boards/BMS/Inc/App/App_CellVoltages.h
pasitlaothamatas: Hmm yeah good idea

pasitlaothamatas: I added a define here `V_PER_100V`

boards/BMS/Inc/App/configs/App_AccumulatorConfigs.h
```diff
@@ -1,7 +1,5 @@
 #pragma once
 
-#include "assert.h"
```
pasitlaothamatas, line 3: Forgot to remove this from an old commit

boards/BMS/Src/App/App_AccumulatorVoltages.c
```diff
@@ -92,21 +95,21 @@ float App_CellVoltages_GetAverageCellVoltage(void)
...
 {
     return (float)App_SumOfArrayElements(
-               &cell_voltages.measured_cells
+               &cell_voltages.raw_cell_voltages
                     [CELL_MONITOR_IC_1 *
                      cell_voltages.num_of_cells_per_segment],
-               cell_voltages.num_of_cells_per_segment) /
-           10000.0f;
+               cell_voltages.num_of_cells_per_segment) *
+           V_PER_100UV;
```
markomarkusi, line 112: ❤️

markomarkusi (approved):
LGTM

PR #677, opened 2020-10-12
Fix BMS specs

pasitlaothamatas:
### Summary

Fix specs for AIRs and SoC.

### Checklist
*Please change `[ ]` to `[x]` when you are ready.*
- [ ] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [ ] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).

PR #683, opened 2020-11-17
Fix AIR related specs that were missed in #677

pasitlaothamatas:
### Summary

- Noticed there were some AIR related specs that still weren't fixed in #677 

### Checklist

### DIFF
--- a/specs.md
+++ b/specs.md
@@ -68,13 +68,13 @@ FSM-18 | Brake actuation reporting | - The FSM must report the brake actuation O
 ID | Title | Description | Associated Competition Rule(s)
 --- | --- | --- | ---
 FSM-17 | Default state | The FSM state machine must begin in the AIR-Open state by default.
-FSM-12 | Exiting the AIR-Open state | The FSM must enter the AIR-Closed state when the BMS closes the AIR+ and AIR-.
+FSM-12 | Exiting the AIR-Open state | The FSM must enter the AIR-Closed state when the BMS closes the AIR+. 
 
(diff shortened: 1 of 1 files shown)


specs.md
```diff
@@ -68,13 +68,13 @@ FSM-18 | Brake actuation reporting | - The FSM must report the brake actuation O
 ID | Title | Description | Associated Competition Rule(s)
 --- | --- | --- | ---
 FSM-17 | Default state | The FSM state machine must begin in the AIR-Open state by default.
-FSM-12 | Exiting the AIR-Open state | The FSM must enter the AIR-Closed state when the BMS closes the AIR+ and AIR-.
+FSM-12 | Exiting the AIR-Open state | The FSM must enter the AIR-Closed state when the BMS closes the AIR+.
```
pasitlaothamatas, line 71: @kirons96 Should I also add "or when the BMS detects AIR+ and AIR- are closed" or does this suffice

kirons96: AIR+ alone should be fine as AIR- is less meaningful (i.e. during precharge)

kirons96 (approved):

specs.md
pasitlaothamatas: @kirons96 Actually before I merge this in could you explain why AIR- is less meaningful

kirons96: AIR- closes during precharge
AIR+ is open during that time
we don't want to be in the "Air-Closed state" during precharge, only after it

PR #686, opened 2020-11-22
Add Flowmeter Callback Functions

pasitlaothamatas:
### Summary

- Add callback functions for the primary and secondary flow meters (FSM-14) used to set/clear can tx motor shutdown signals

### Testing Done

- Added unit tests to make sure that the motor shutdown can tx signal is set only after the flow rate is below 1.0L/min for one second
- Added unit tests to check that the motor shutdown can tx signal can be cleared after the flow rate is greater than 1.0L/min for one second

### Checklist


boards/FSM/Inc/App/App_FlowMeterSignals.h
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Check if the secondary flow rate is in range
+ * @param world The world to get the secondary flow rate in-range check
+ * @return true if the secondary flow rate (L/min) is in range
+ */
+bool App_FlowMetersSignals_IsSecondaryFlowRateInRange(struct FsmWorld *world);
+
+/**
+ * The callback function for the primary flow rate signal in the given world
```
alxshev, line 38: Since these are called in the case of underflow, I think they should be named something like PrimaryFlowRateBelowThresholdCallback

pasitlaothamatas: 👍👍

boards/FSM/Inc/App/configs/App_SignalCallbackDurations.h
```diff
@@ -8,3 +8,5 @@
 #define APPS_EXIT_HIGH_MS 1000
 #define APPS_AND_BRAKE_ENTRY_HIGH_MS 1
 #define APPS_AND_BRAKE_EXIT_HIGH_MS 1
+#define FLOW_METER_ENTRY_HIGH_MS 1000
+#define FLOW_METER_EXIT_HIGH_MS 1000
```
kirons96, line 12: these defines should be unsigned (i.e. 1000U) so that you're not comparing unsigned and signed ints (assuming that the time variables you will compare/store these with are unsigned)

also, every define you make should have brackets around every number and variable, i.e. `#define yar (1000U)` - https://stackoverflow.com/a/9081708

could be worth doing a dedicated PR for that instead of doing it here

pasitlaothamatas: image

alxshev (approved):
lgtm :moneybag:

boards/FSM/Inc/App/configs/App_SignalCallbackDurations.h
pasitlaothamatas: This should be addressed now, king

kirons96: in all .h files? or just this one?

pasitlaothamatas: Just this one

PR #688, opened 2020-12-03
Add App and Io Code for enabling Pre-charge

pasitlaothamatas:
### Summary

- Add app and io code to enable precharging

### Checklist

### DIFF
--- a/boards/BMS/Src/Io/Io_PreCharge.c
+++ b/boards/BMS/Src/Io/Io_PreCharge.c
@@ -1,0 +1,13 @@
+#include "Io_PreCharge.h"
+#include "main.h"
+
+void Io_PreCharge_Enable(void)
+{
+    HAL_GPIO_WritePin(PRE_CHARGE_EN_GPIO_Port, PRE_CHARGE_EN_Pin, GPIO_PIN_SET);
+}
+
+void Io_PreCharge_Disable(void)
+{
+    HAL_GPIO_WritePin(
+        PRE_CHARGE_EN_GPIO_Port, PRE_CHARGE_EN_Pin, GPIO_PIN_RESET);
+}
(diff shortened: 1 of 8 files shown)


boards/BMS/Src/Io/Io_PreCharge.c
```diff
@@ -0,0 +1,13 @@
+#include "Io_PreCharge.h"
+#include "main.h"
+
+void Io_PreCharge_Enable(void)
+{
+    HAL_GPIO_WritePin(PRE_CHARGE_EN_GPIO_Port, PRE_CHARGE_EN_Pin, GPIO_PIN_SET);
```
alxshev, line 6: I noticed some of the other Io methods that enable/disable pins return ExitCodes (like in Io_OkStatuses), whereas these return void. Is there reason to use one over the other?

alxshev: Aside from this question everything looks good

pasitlaothamatas: For the fxns in Io_OkStatuses I think Ken wanted to differentiate pins that we could actually use to enable OkStatuses from ones that couldn't.

pasitlaothamatas: In our case since we're only working with the PRE_CHARGE_EN pin (which also happens to be a pin we can control atm) using a void fxn makes the most sense

alxshev: ok

alxshev: ok

alxshev (approved):
Let's Get This Money

alxshev:
> Let's Get This Money

alxshev:
Oops wrong button

pasitlaothamatas:
💸💸💸

PR #689, opened 2020-12-03
Fix Io_VoltageSense

pasitlaothamatas:
### Summary

- Fix Io_VoltageSense

### Changelist 

- Return floating point value instead of an ExitCode. This is done so that we can generate fake TS+ values when testing BMS-12

### Checklist


pasitlaothamatas:
Actually this might not be necessary

alxshev (approved):
Looks good

PR #695, opened 2020-12-19
Add previous ms to Clock

pasitlaothamatas:
### Summary

- Add previous ms to the current Clock. This will be helpful for timekeeping when I write code for BMS-12.

### Changelist 

- Fixed const correctness for App_Clock

### Testing Done

- Added tests to check that prev_ms can be set and acquired from the clock object correctly 

### Checklist

### DIFF
--- a/boards/shared/Inc/App/App_SharedClock.h
+++ b/boards/shared/Inc/App/App_SharedClock.h
@@ -26,6 +26,15 @@ void App_SharedClock_SetCurrentTimeInMilliseconds(
     uint32_t      current_time_ms);
 
 /**
+ * Set the previous time for the given clock, in milliseconds
+ * @param clock The clock to set the previous time for
+ * @param current_time_ms The previous time to set, in milliseconds
+ */
+void App_SharedClock_SetPreviousTimeInMilliseconds(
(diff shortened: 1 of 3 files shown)


boards/shared/Inc/App/App_SharedClock.h
```diff
@@ -42,3 +51,11 @@ uint32_t
...
  */
 uint32_t App_SharedClock_GetCurrentTimeInSeconds(const struct Clock *clock);
+
+/**
+ * Get the previous time for the given clock, in milliseconds
+ * @param clock The clock to get the previous time for
+ * @return The current time for the given clock, in milliseconds
+ */
+uint32_t
+    App_SharedClock_GetPreviousTimeInMilliseconds(const struct Clock *clock);
```
markomarkusi, line 61: Maybe add a previous time in seconds as well? Otherwise LGTM

markomarkusi (approved):

boards/shared/Inc/App/App_SharedClock.h
pasitlaothamatas: Hmm yeah shore I can add prev time in seconds

PR #709, opened 2021-07-24
Replaced all occurrences of isnanf with isnan to support newer compilers

alxshev:
### Summary
Replaced all occurrences of `isnanf` with `isnan` to support newer compilers. This was done because the `isnanf` function is considered obsolete. Instead, the `isnan(float x)` function was declared as a replacement in C99 (see [here](https://linux.die.net/man/3/isnanf) under **Notes**).

### Changelist 
- Renamed all occurrences of `isnanf` with the `isnan(float x)` macro in the source code
- Modified CAN generation Python script to use `isnan` also

### Checklist
- [ x] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).

### DIFF
--- a/scripts/codegen/CAN/cantx_codegen.py
+++ b/scripts/codegen/CAN/cantx_codegen.py
@@ -136,7 +136,7 @@ class AppCanTxFileGenerator(CanFileGenerator):
                         function_prefix, signal.snake_name.upper(), self._sender.capitalize(), signal.type_name),
                         '',
                         '''\
-    if (isnanf(value))
+    if (isnan(value))
     {{
         can_tx_interface->periodic_can_tx_table.{msg_snakecase_name}.{signal_snakecase_name} = value;
     }}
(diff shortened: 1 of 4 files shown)


scripts/codegen/CAN/cantx_codegen.py
```diff
@@ -136,7 +136,7 @@ def __init_functions(self, function_prefix):
                         function_prefix, signal.snake_name.upper(), self._sender.capitalize(), signal.type_name),
                         '',
                         '''\
-    if (isnanf(value))
+    if (isnan(value))
```
pasitlaothamatas, line 139: I think we should keep this as isnanf since this code is auto-generated, and compiled for our arm processor

pasitlaothamatas:
isnan takes in a double as an input parameter, while isnanf takes in a float. May be better here to include cmath.h and use std::isnan if you are running into issues

pasitlaothamatas:
https://en.cppreference.com/w/cpp/numeric/math/isnan

pasitlaothamatas:
Yeah this is probably cool

pasitlaothamatas (approved):

PR #713, opened 2021-07-25
[BMS] Add ErrorTable Integration to CAN Broadcasting System

alxshev:
Add ErrorTable Integration to CAN Broadcasting System

### Summary
- Modified `App_SetPeriodicCanSignals_InRangeCheck(...)` to return an `InRangeCheck_Status` indicating whether the `InRangeCheck` values are out of range. This was done so that a board's outgoing `InRangeCheck` CAN signals can be used to set that board's error table
- Added code to set BMS error table values based on received CAN messages. This was done in the CanRx task.

### DIFF
--- a/boards/BMS/Src/App/App_Accumulator.c
+++ b/boards/BMS/Src/App/App_Accumulator.c
@@ -1,5 +1,6 @@
 #include <stdlib.h>
 #include <assert.h>
+#include <stdbool.h>
 #include "App_Accumulator.h"
 
 struct Accumulator
@@ -156,4 +157,22 @@ struct InRangeCheck *App_Accumulator_GetSegment5VoltageInRangeCheck(
     const struct Accumulator *const accumulator)
 {
     return accumulator->segment_5_voltage_in_range_check;
+}
+
+bool App_Accumulator_IsMinCellVoltageOutOfRange(
+    const struct Accumulator *const accumulator)
+{
+    float min_cell_voltage;
+    return App_InRangeCheck_GetValue(
+               accumulator->min_cell_voltage_in_range_check,
+               &min_cell_voltage) != VALUE_IN_RANGE;
+}
+
+bool App_Accumulator_IsMaxCellVoltageOutOfRange(
+    const struct Accumulator *const accumulator)
+{
+    float max_cell_voltage;
+    return App_InRangeCheck_GetValue(
+               accumulator->max_cell_voltage_in_range_check,
+               &max_cell_voltage) != VALUE_IN_RANGE;
 }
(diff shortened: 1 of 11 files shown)


boards/BMS/Src/main.c
```diff
@@ -972,7 +974,7 @@ void RunTaskCanTx(void const *argument)
 
     for (;;)
     {
-        Io_SharedCan_TransmitEnqueuedCanTxMessagesFromTask();
+        Io_SharedCan_TransmitEnqueuedCanTxMessagesFromTask(error_table);
```
pasitlaothamatas, line 977: I think it is better to set BMS errors through the state machine when we detect it. `Io_SharedErrorTable_SetErrorsFromCanMsg` is better suited for setting the error table from received CAN messages

alxshev: I know, this method calls `Io_SharedErrorTable_SetErrorsFromCanMsg` if you look down below.  I haven't called it in the `RunTask___` method like I did for CanRx because the CAN message needs to get popped off the tx queue first, and this is done in `Io_SharedCan_TransmitEnqueuedCanTxMessagesFromTask`

alxshev:
BMS stays in fault state like it's supposed to now

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -157,3 +158,21 @@ struct InRangeCheck *App_Accumulator_GetSegment5VoltageInRangeCheck(
 {
     return accumulator->segment_5_voltage_in_range_check;
 }
+
+bool App_Accumulator_IsMinCellVoltageOutOfRange(
```
pasitlaothamatas, line 162: Within `App_SetPeriodicSignals_AccumulatorInRangeChecks` there are calls to `App_SetPeriodicCanSignals_InRangeCheck`. If we get this function to return its `InRangeCheck_Status` we will not have to check with the CAN struct whether something is out of range or not

pasitlaothamatas: Would also avoid the need to add two more functions here to check if the mix and max cell voltages are out of range

pasitlaothamatas: For relevant min and max voltages, we can do a simple check as well as set the error table in `App_SetPeriodicSignals_AccumulatorInRangeChecks` 

```
if(App_SetPeriodicCanSignals_InRangeCheck(...) != VALUE_IN_RANGE)
{
    set_error_table();
}
```

pasitlaothamatas (commented):
Nice job. Left small comments on how setting the error on the BMS can be improved.

Could you update the scope of the PR to include information on changes that you plan on making to the InRangeChecks, as well as how you plan to set the error triggered by the BMS and errors received via CAN

alxshev:
> Nice job. Left small comments on how setting the error on the BMS can be improved.
> 
> Could you update the scope of the PR to include information on changes that you plan on making to the InRangeChecks, as well as how you plan to set the error triggered by the BMS and errors received via CAN

Sure, amended PR description

pasitlaothamatas (approved):
$$$$

pasitlaothamatas:
Lets get this moonies

PR #716, opened 2021-09-26
Added gtest compatibility for Windows systems

alxshev:
### Summary
Currently, the gtest suite does not run on Windows. This is because the semaphore/mutex mechanism in `App_SharedStateMachine` is only compatible with UNIX or ARM. This PR adds a Windows equivalent so that gtests can run

### DIFF
--- a/boards/CMakeLists.txt
+++ b/boards/CMakeLists.txt
@@ -284,7 +284,12 @@ function(compile_googletest_executable
         ARM_BINARY_X86_COMPATIBLE_SRCS
         ARM_BINARY_INCLUDE_DIRS
         )
-    set(TEST_EXECUTABLE_NAME "${BOARD_NAME}_test")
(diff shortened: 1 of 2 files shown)


boards/CMakeLists.txt
```diff
@@ -284,7 +284,12 @@ function(compile_googletest_executable
         ARM_BINARY_X86_COMPATIBLE_SRCS
         ARM_BINARY_INCLUDE_DIRS
         )
-    set(TEST_EXECUTABLE_NAME "${BOARD_NAME}_test")
+    if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+        set(TEST_EXECUTABLE_NAME "${BOARD_NAME}_test.exe")
```
alxshev, line 288: This is needed because windows generates an .exe at the end of the board test executables

markomarkusi (approved):
Awesome, works on my end too 💯

PR #718, opened 2021-11-06
Implement state transition conditions from init to drive

akrantz01:
### Summary
Implements the conditions for transitioning between the init and drive states as required by DCM-5:

_The DCM must also meet the following conditions in sequential order before entering the drive state:_
_1. The start switch must be switched in an upwards direction. If the start switch was already in the upwards position, it must be re-switched into the upwards position._
_2. The brakes must be actuated._

### Changelist 
- Added `StartSwitch` to `DcmWorld` to keep track of the position of the start switch
- Added pre-transition checks for
  -  no critical errors on any board
  - positive and negative AIRs closed
  - break actuated
  - switch in the up position

### Testing Done
Added two unit tests to ensure the transitions work correctly for when the start switch is initially in either the [up](https://github.com/akrantz01/Consolidated-Firmware/blob/9449aad755313f5caeba3e8814d9695f2a3f6867/boards/DCM/Test/Src/Test_StateMachine.cpp#L193-L218) or [down](https://github.com/akrantz01/Consolidated-Firmware/blob/9449aad755313f5caeba3e8814d9695f2a3f6867/boards/DCM/Test/Src/Test_StateMachine.cpp#L167-L190) position.

### Resolved Issues

[DCM-5](https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/master/specs.md#dcm-init-state-)

### Checklist


boards/DCM/Inc/App/App_StartSwitch.h
```diff
@@ -0,0 +1,49 @@
...
+void App_StartSwitch_SetInitialSwitchPosition(
+    struct StartSwitch *start_switch,
+    bool                position);
+
+/**
+ * Mark the start switch as having been toggled based on whether it was in the
+ * down position
+ * @param start_switch The given init state conditions to set whether the switch
+ *                     toggled on
+ * @param state The current state of the switch
```
timbitss, line 28: Comments could be more specific. E.g. Set start switch has having been toggled OFF based on present and initial state of switch.

boards/DCM/Src/App/App_StartSwitch.c
```diff
@@ -0,0 +1,53 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+struct StartSwitch
+{
+    bool initial_switch_position;
+    bool switch_toggled;
```
timbitss, line 8: I would rename `switch_toggled` to `switch_toggled_off` to be more specific.

boards/DCM/Src/App/App_StartSwitch.c
```diff
@@ -0,0 +1,53 @@
...
+    bool                position)
+{
+    start_switch->initial_switch_position = position;
+}
+
+void App_StartSwitch_SetSwitchToggled(
+    struct StartSwitch *start_switch,
+    bool                state)
+{
+    if (!state)
```
timbitss, line 32: To better represent the purpose of the function, I would check `if(start_switch->initial_switch_position == ON && state == OFF)`

timbitss (commented):
Good job. I don't think we've checked whether precharge has occurred in the code yet. We will also need to confirm if the "the shutdown circuit" refers to the AIRs.

boards/DCM/Inc/App/App_StartSwitch.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+void App_StartSwitch_SetSwitchToggled(
+    struct StartSwitch *start_switch,
+    bool                state);
+
+/**
+ * Check whether the switch conditions have passed to transition from the init
+ * state
+ * @param start_switch The given init state conditions to check against
+ * @param state The current state of the switch
```
alxshev, line 38: Since we're already making a start switch struct we should be storing the current state of the switch inside the struct, not passing it as a parameter

boards/DCM/Src/App/states/App_InitState.c
```diff
@@ -23,12 +29,35 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+    bool bms_positive_air_closed =
+        App_CanRx_BMS_AIR_STATES_GetSignal_AIR_POSITIVE(can_rx_interface) == 1;
+    bool bms_negative_air_closed =
+        App_CanRx_BMS_AIR_STATES_GetSignal_AIR_NEGATIVE(can_rx_interface) == 1;
+    uint8_t start_switch_position =
+        App_CanRx_DIM_SWITCHES_GetSignal_START_SWITCH(can_rx_interface);
+    uint8_t break_actuated =
+        App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx_interface);
+
+    App_StartSwitch_SetSwitchToggled(start_switch, start_switch_position);
```
alxshev, line 50: It would be better to store the current start switch position inside the start struct and update that value by calling SetSwitchPosition here instead.

boards/DCM/Src/App/App_StartSwitch.c
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void App_StartSwitch_SetInitialSwitchPosition(
+    struct StartSwitch *start_switch,
+    bool                position)
+{
+    start_switch->initial_switch_position = position;
+}
+
+void App_StartSwitch_SetSwitchToggled(
```
alxshev, line 28: I think we change this function to be SetSwitchPosition, and record the current state of the start switch inside the struct. The toggle logic can still be handled here by comparing initial state with current state

timbitss:
inactive

PR #719, opened 2021-11-06
DCM20 - Transition to fault state from drive state if motor shutdown error detected. 

timbitss:
### Changelist 

- Modified CAN_Rx task within main.c to update the error table based on the latest CAN message received.
- Added a test within Test_StateMachine.cpp to check for state transition. 
- Updated App_DriveState.c to transition to fault state from drive state if motor shutdown error detected.
- Set receiver of all motor shutdown errors to DCM from DEBUG in App_CanMsgs.dbc. 

### Testing Done
- Added a test within Test_StateMachine.cpp to check for state transition from DRIVE to FAULT when a motor shutdown error is set within the error table.
 
### Checklist


akrantz01 (approved):
LGTM!

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -27,13 +28,18 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
     App_AllStatesRunOnTick100Hz(state_machine);
 
     struct DcmWorld *world = App_SharedStateMachine_GetWorld(state_machine);
-    struct DcmCanRxInterface *can_rx = App_DcmWorld_GetCanRx(world);
+    struct DcmCanRxInterface *can_rx      = App_DcmWorld_GetCanRx(world);
+    struct ErrorTable *const  error_table = App_DcmWorld_GetErrorTable(world);
```
alxshev, line 32:
```suggestion
    struct ErrorTable *error_table = App_DcmWorld_GetErrorTable(world);
```
Just for consistency with `error_table` code

boards/DCM/Test/Src/Test_StateMachine.cpp
```diff
@@ -580,4 +580,24 @@ TEST_F(
         0.0f, App_CanTx_GetPeriodicSignal_TORQUE_REQUEST(can_tx_interface));
 }
 
+// DCM-20
+TEST_F(
+    DcmStateMachineTest,
+    transition_to_fault_state_from_drive_state_if_motor_shutdown_error_detected)
```
alxshev, line 586: This is good. Can we also include tests where the error is introduced into the CanRx table instead of the error table (using `CanRx_SetSignal`)? So one test for receiving the BMS motor shutdown error over CAN, one for receiving FSM motor shutdown error over CAN, etc.

timbitss: In the CAN RX task, `Io_SharedErrorTable_SetErrorsFromCanMsg()` is called below `Io_CanRx_UpdateRxTableWithMessage()` which updates the error table with the latest CAN RX message in the queue. Therefore, I don't think updating the CanRx table will update the error table. I can add more to the test to check if motor shutdown errors from other boards result in a state transition.

PR #726, opened 2022-01-24
Add minor fix to shared spi driver

pasitlaothamatas:
### Summary

- Simplifying shared spi logic
- Change return type to bool as checks for HAL statuses are cluttering Io code that uses SPI
- Working on a branch that will address the impact of changing shared spi return type

### Checklist

### DIFF
--- a/boards/shared/Src/Io/Io_SharedSpi.c
+++ b/boards/shared/Src/Io/Io_SharedSpi.c
@@ -41,7 +41,7 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *const spi_interface)
         spi_interface->nss_port, spi_interface->nss_pin, GPIO_PIN_SET);
 }
 
-HAL_StatusTypeDef Io_SharedSpi_TransmitAndReceive(
+bool Io_SharedSpi_TransmitAndReceive(
     const struct SharedSpi *const spi_interface,
     uint8_t *                     tx_buffer,
     uint16_t                      tx_buffer_size,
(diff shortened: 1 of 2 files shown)


pasitlaothamatas:
image

Happy new year lol. Think I will need to disable STM32CubeMX Code Generation for this PR

kirons96:
> image
> 
> Happy new year lol. Think I will need to disable STM32CubeMX Code Generation check for this PR?

sounds good to me. i think you should create a dedicated PR which updates all the auto-generated files' years

kirons96 (approved):

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -41,92 +41,90 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *const spi_interface)
...
+        HAL_SPI_Receive(
+            spi_interface->spi_handle, rx_buffer, rx_buffer_size,
+            spi_interface->timeout_ms) == HAL_OK;
     Io_SharedSpi_SetNssHigh(spi_interface);
 
     return status;
 }
 
-HAL_StatusTypeDef Io_SharedSpi_MultipleTransmitWithoutNssToggle(
+bool Io_SharedSpi_MultipleTransmitWithoutNssToggle(
```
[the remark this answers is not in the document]
pasitlaothamatas: I believe this is used once in `Io_LTC6813_ConfigureRegisterA`. Since its only used once i might remove it from `SharedSpi`. I can do this in my PR to refactor LTC6813 related stuff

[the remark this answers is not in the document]
kirons96: shore

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -41,92 +41,90 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *const spi_interface)
...
-        spi_interface->spi_handle, rx_buffer, rx_buffer_size,
-        spi_interface->timeout_ms);
-    if (status != HAL_OK)
+    if (HAL_SPI_Transmit(
+            spi_interface->spi_handle, tx_buffer, tx_buffer_size,
+            spi_interface->timeout_ms) == HAL_OK)
     {
-        Io_SharedSpi_SetNssHigh(spi_interface);
-        return (status);
+        status = HAL_SPI_Receive(
```
alxshev, line 60: This is cleaner, and don't need to declare status separately:
```
const bool status = HAL_SPI_Transmit(...) == HAL_OK &&  HAL_SPI_Receive(...) == HAL_OK;
 ```

File: boards/shared/Src/Io/Io_SharedSpi.c [hunk 3]
```diff
@@ -41,92 +41,90 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *const spi_interface)
...
-            spi_interface->spi_handle, tx_buffer, tx_buffer_size,
-            spi_interface->timeout_ms);
-        if (status != HAL_OK)
+        if (HAL_SPI_Transmit(
+                spi_interface->spi_handle, tx_buffer, tx_buffer_size,
+                spi_interface->timeout_ms) != HAL_OK)
         {
-            return status;
+            status = false;
+            break;
```
alxshev, line 115: There's no need for a status var, not like the status ever changes once it's set. Simpler to return false here and return true at the bottom of the method

alxshev (commented):
Left some code style feedback. Is the point of this PR to make the code which uses these methods cleaner by removing the constant `== HAL_OK` comparisons?

alxshev:
Also, what is the current approach for handling cases where SPI fails to transmit? Do we keep trying to send the message?

pasitlaothamatas:
> Left some code style feedback. Is the point of this PR to make the code which uses these methods cleaner by removing the constant `== HAL_OK` comparisons?

Basically. IMO our IO code just needs to know if something was sent successfully.

boards/shared/Src/Io/Io_SharedSpi.c
pasitlaothamatas: kinda wanted a single return statement here

pasitlaothamatas: ehhhh im breaking anyways. sure i can just return false. will fix this

boards/shared/Src/Io/Io_SharedSpi.c
```diff
@@ -41,92 +41,90 @@ void Io_SharedSpi_SetNssHigh(const struct SharedSpi *const spi_interface)
...
-    {
-        Io_SharedSpi_SetNssHigh(spi_interface);
-        return status;
-    }
 
-    status = HAL_SPI_Receive(
-        spi_interface->spi_handle, rx_buffer, rx_buffer_size,
-        spi_interface->timeout_ms);
-    if (status != HAL_OK)
+    if (HAL_SPI_Transmit(
```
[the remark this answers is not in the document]
pasitlaothamatas: yup basically https://github.com/ARMmbed/mbed-hal-st-stm32cubef4/blob/master/source/stm32f4xx_hal_spi.c

pasitlaothamatas:
> What is our current approach for handling cases where SPI fails to transmit? Do we keep trying to send the message regardless of the failure mode?

Depends on what you wanna do. Can try to send the msg again, or can drop the message

boards/shared/Src/Io/Io_SharedSpi.c
kirons96: i've seen bugs at work where people release a mutex after returning in a short function like this - big fan of one return per function to avoid stupid bugs like that

alxshev:
> > What is our current approach for handling cases where SPI fails to transmit? Do we keep trying to send the message regardless of the failure mode?
> 
> Depends on what you wanna do. Can try to send the msg again, can also drop the message and return an error.

I'm wondering whether code using this driver would ever need to know the specific kind of error message given by SPI and act on it, like do one thing if it's `HAL_TIMEOUT` and another if it's `HAL_ERROR`.
Just a thought, I'll approve once you push the other updates

timbitss (approved):
Looks good to me, straightforward and easy to read, nothing ground breaking

PR #729, opened 2022-02-03
Update Pipfile to point cantools package to forked version with no doubles

timbitss:
### Summary
Updated Pipfile to install the cantools package from FormulaE [fork](https://github.com/UBCFormulaElectric/cantools), removing the usage of doubles in `encode()` and `decode()` functions. V33.1.1 is still maintained.

Motivation: Some of the inverter CAN signals have an offset of 0.1 and thus require the encode() and decode() functions. The use of doubles in the autogenerated encode() and decode() functions are unnecessary since it increases the execution time (STM32F4's FPU supports floats only). Therefore, I updated cantools to generate code that uses floats instead of doubles.

Before App_CanMsgs.c:
image

After App_CanMsgs.c:
image

### Changelist 
- Updated Pipfile to install the cantools package from FormulaE [fork](https://github.com/UBCFormulaElectric/cantools). 
- Updated year of stm32f3xx_hal_conf.h file in each board folder from 2021 to 2022

### Testing Done

1. Entered  `pipenv uninstall -all` in the command line to uninstall all packages without modifying the Pipfile.  
2. Entered `pipenv install` to re-install all packages including cantools which is now located at the fork.
3. Built .elf file to see that doubles had changed to floats in App_CanMsgs.h and App_CanMsgs.c

### Checklist

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ name = "pypi"
 pylint = "*"
 
 [packages]
-cantools = "~=33.1.1"
+cantools = {ref = "v33.1.1-noDoubles", git = "git://github.com/UBCFormulaElectric/cantools"}
 
 [requires]
 python_version = "3"
--- a/boards/BMS/Inc/stm32f3xx_hal_conf.h
+++ b/boards/BMS/Inc/stm32f3xx_hal_conf.h
@@ -5,7 +5,7 @@
   ******************************************************************************
   * @attention
   *
-  * <h2><center>&copy; COPYRIGHT(c) 2021 STMicroelectronics</center></h2>
+  * <h2><center>&copy; COPYRIGHT(c) 2022 STMicroelectronics</center></h2>
   *
   * Redistribution and use in source and binary forms, with or without modification,
   * are permitted provided that the following conditions are met:
--- a/boards/DCM/Inc/stm32f3xx_hal_conf.h
+++ b/boards/DCM/Inc/stm32f3xx_hal_conf.h
@@ -5,7 +5,7 @@
   ******************************************************************************
   * @attention
   *
-  * <h2><center>&copy; COPYRIGHT(c) 2021 STMicroelectronics</center></h2>
+  * <h2><center>&copy; COPYRIGHT(c) 2022 STMicroelectronics</center></h2>
   *
   * Redistribution and use in source and binary forms, with or without modification,
   * are permitted provided that the following conditions are met:
--- a/boards/DIM/Inc/stm32f3xx_hal_conf.h
+++ b/boards/DIM/Inc/stm32f3xx_hal_conf.h
@@ -5,7 +5,7 @@
   ******************************************************************************
   * @attention
   *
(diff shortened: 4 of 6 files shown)


alxshev:
What's the rationale for doing this?

timbitss:
> Why is this being done?

Some of the inverter CAN signals have an offset of 0.1 and thus require the encode() and decode() functions. The use of doubles in the autogenerated encode() and decode() functions are unnecessary since it increases the execution time (STM32F4's FPU supports floats only). Therefore, I updated cantools to generate code that uses floats instead of doubles.

Counterpoint: we could go without using the encode() and decode() functions and do the conversions directly, but I think it's more elegant and less error-prone if we use the auto-gen code.

alxshev (approved):
lgtm, just add what you said in the comments to the PR description so people in the future know why this was done

akrantz01:
It looks like there is already an issue open for adding this to mainline cantools (cantools/cantools#393). I think it would be better to wait a bit for the feature request to get resolved and update to the latest version than maintain a custom fork of cantools. Since if we ever want to update the cantools version, we would need to manually update the fork every time.

If we wanted this sooner, it probably wouldn't be too difficult to adapt your change into a PR for the cantools repo to implement this feature.

Pipfile
```diff
@@ -7,7 +7,7 @@ name = "pypi"
 pylint = "*"
 
 [packages]
-cantools = "~=33.1.1"
+cantools = {git = "git://github.com/timbitss/cantools.git"}
```
pasitlaothamatas, line 10: if we decide to go with the forking approach, i think its a better idea to fork into UBCFormulaElectric org and update the following line to `{git = "git://github.com/UBCFormulaElectric/cantools.git"}`

pasitlaothamatas:
> It looks like there is already an issue open for adding this to mainline cantools ([cantools/cantools#393](https://github.com/cantools/cantools/issues/393)). I think it would be better to wait a bit for the feature request to get resolved and update to the latest version than maintain a custom fork of cantools. Since if we ever want to update the cantools version, we would need to manually update the fork every time.
> 
> If we wanted this sooner, it probably wouldn't be too difficult to adapt your change into a PR for the cantools repo to implement this feature.

Good point. Since this isnt blocking, i dont mind waiting a bit. In the case where we need to make a quick change, i dont mind using the forking approach either since we will probably stick to V33.1.1 (probably wont need to update cantools)

timbitss:
I'll test V37 with the DCM board tmmrw. If V37 doesn't break anything, then I think it's best we wait out for [issue](https://github.com/cantools/cantools/issues/393) to be resolved. Otherwise, I'll create a fork in the main repo with the changes.

pasitlaothamatas:
lgtm

pasitlaothamatas (approved):

timbitss:
I tried building with cantools V37.7 but got this error message. After an hour of trying to debug this issue, @pasitlaothamatas and I believed it wasn't worth the time. 
image

PR #730, opened 2022-02-22
Pre-charge. Overall system working correctly now with OC and SC fault detection.

Eddy13013:
### Summary
Precharge is now operating correctly. SC and OC scenarios will trigger a fault. An expected range of time for precharge to complete successfully can be modified and the trigger threshold can also be altered (max threshold will be (ACC_V - 15V)/ACC_V because of a 15mA current draw from BMS_HV and PCM). Voltage sense is working great (within about +/-1%). 

### Changelist 

Voltage readings are within +-1% now with an update to error correction.

### Testing Done

Overall System has been tested. OC and SC scenarios are tested. Tested an external shutdown scenario where something other than the BMS triggers and re-enables the shutdown circuit. Found that the precharge relay was briefly clicking on and off. This is a hardware-based response. This issue is unimportant too (it doesn't stay on for very long)

### Resolved Issues

**Differential Voltage sense using ADC.** First, configure ADC_1_IN_3 to differential. ADC_1_IN_4 is implicitly included as the negative side of this signal. Then offset the value by half of the max ADC reading because a differential signal is expected to be centered at 0V and expand to -3.3V and +3.3V. Without an offset, the ADC will read 1/2 at 0V. Instead of scaling the ADC reading by 3.3,  you need to scale by 6.6 for a differential (this is the total range of the input. if the input range was from -5V to +5V or 0V to 10V then you would need to scale by 10). 

### Checklist


boards/BMS/BMS.ioc
```diff
@@ -13,17 +13,18 @@ ADC1.EnableInjectedConversion=ENABLE
...
 ADC1.NbrOfConversion=1
-ADC1.Offset-0\#ChannelRegularConversion=0
-ADC1.OffsetNumber-0\#ChannelRegularConversion=ADC_OFFSET_NONE
+ADC1.Offset-0\#ChannelRegularConversion=1983
+ADC1.OffsetNumber-0\#ChannelRegularConversion=ADC_OFFSET_3
 ADC1.Overrun=ADC_OVR_DATA_OVERWRITTEN
 ADC1.Rank-0\#ChannelRegularConversion=1
 ADC1.Resolution=ADC_RESOLUTION_12B
-ADC1.SamplingTime-0\#ChannelRegularConversion=ADC_SAMPLETIME_61CYCLES_5
+ADC1.SamplingTime-0\#ChannelRegularConversion=ADC_SAMPLETIME_601CYCLES_5
```
pasitlaothamatas, line 25: Did we try playing with the sample time after changing the adc offset to `ADC_OFFSET_3`? Might be a good idea to play around with this value

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -1,14 +1,43 @@
 #include "states/App_AllStates.h"
+#include "states/App_PreChargeState.h"
+#include "App_AccumulatorVoltages.h"
+#include "Io_VoltageSense.h"
+#include "states/App_DriveState.h"
+#include "Io_PreCharge.h"
+#include "states/App_FaultState.h"
+#include <math.h>
```
pasitlaothamatas, line 8: what do we need from math.h here

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -1,14 +1,43 @@
...
+#include "Io_Airs.h"
+#include "states/App_InitState.h"
+#include "Io_Charger.h"
+#include "states/App_ChargeState.h"
 
 #include "App_SetPeriodicCanSignals.h"
 #include "App_SharedMacros.h"
 
+#define S_TO_MS (1000U)
```
pasitlaothamatas, line 17: this can probably go into `App_SharedConstants.h`

boards/BMS/Inc/Io/Io_VoltageSense.h
```diff
@@ -5,4 +5,4 @@
  * @param adc_voltage The ADC voltage to convert
  * @return The tractive system voltage in V
  */
-float Io_VoltageSense_GetTractiveSystemVoltage(float adc_voltage);
+float Io_VoltageSense_GetTractiveSystemVoltage();
```
pasitlaothamatas, line 8: nit; 
```suggestion
float Io_VoltageSense_GetTractiveSystemVoltage(void);
```

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -1,14 +1,43 @@
...
 static void PreChargeStateRunOnEntry(struct StateMachine *const state_machine)
 {
     struct BmsWorld *world = App_SharedStateMachine_GetWorld(state_machine);
     struct BmsCanTxInterface *can_tx_interface = App_BmsWorld_GetCanTx(world);
+    struct Clock *            clock            = App_BmsWorld_GetClock(world);
     App_CanTx_SetPeriodicSignal_STATE(
         can_tx_interface, CANMSGS_BMS_STATE_MACHINE_STATE_PRE_CHARGE_CHOICE);
+
+    // capture start time of precharge
+    pre_charge_start_ms = App_SharedClock_GetCurrentTimeInMilliseconds(clock);
```
pasitlaothamatas, line 37: We can use `App_SharedClock_SetPreviousTimeInMilliseconds` to update the clock and remove the `pre_charge_start_ms` entirely

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -20,11 +49,67 @@ static void
...
     App_AllStatesRunOnTick100Hz(state_machine);
+
+    struct BmsWorld *world = App_SharedStateMachine_GetWorld(state_machine);
+    struct Clock *   clock = App_BmsWorld_GetClock(world);
+    struct Airs *    airs  = App_BmsWorld_GetAirs(world);
+
+    float tractive_system_voltage = Io_VoltageSense_GetTractiveSystemVoltage();
+    float accumulator_voltage     = 355.0f; // TODO: Replace value with function to
+                                        // retrieve total pack voltage
+    uint32_t elapsed_time =
```
pasitlaothamatas, line 60: dont expect this to change so nicer if we make it a const

```suggestion
    const uint32_t elapsed_time =
```

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -20,11 +49,67 @@ static void
...
+        App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+    }
+
+    // if the voltage increases too quickly there may be an OC before capacitor
+    else if (
+        tractive_system_voltage > accumulator_voltage &&
+        elapsed_time <= PRECHARGE_EXPECTED_CHARGE_TIME_MS * 0.20)
+    {
+        App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+    }
```
pasitlaothamatas, line 77: can probably do an OR here since they both transition into the fault state

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -20,11 +49,67 @@ static void
     PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
 {
     App_AllStatesRunOnTick100Hz(state_machine);
+
```
pasitlaothamatas, line 52: might be nicer to read if we have a single state variable and updated the variable within our if-else statements like so:

```
state = precharge
if(voltage not rising OR voltage rising too quick)
   state = fault
else if (precharge ok)
   state = drive/charging

// so on...

App_SetNextState(state_macihne, state)

```

boards/BMS/Src/App/states/App_InitState.c
```diff
@@ -33,14 +38,14 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
     struct BmsWorld *world = App_SharedStateMachine_GetWorld(state_machine);
     struct Clock *   clock = App_BmsWorld_GetClock(world);
 
-    // After entering the init state for 5 seconds, enter the AIR open state
-    if (App_SharedClock_GetCurrentTimeInMilliseconds(clock) -
```
pasitlaothamatas, line 37: remove clock if its not being used

boards/BMS/Inc/Io/Io_Adc.h
```diff
@@ -19,6 +19,12 @@ uint16_t *Io_Adc_GetRawAdc2Values(void);
  */
 float Io_Adc_GetAdc1Channel3Voltage(void);
 
+/**
+ * Get the voltage measured at ADC1 channel 4
+ * @return The voltage measured at ADC1 channel 4, in volts
+ */
+float Io_Adc_GetAdc1Channel4Voltage(void);
```
pasitlaothamatas, line 26: we shouldnt need this for differential ADC correct?

boards/BMS/Src/stm32f3xx_hal_msp.c
```diff
@@ -108,11 +108,12 @@ void HAL_ADC_MspInit(ADC_HandleTypeDef *hadc)
...
         /**ADC1 GPIO Configuration
         PA2     ------> ADC1_IN3
+        PA3     ------> ADC1_IN4
         */
-        GPIO_InitStruct.Pin  = TS_VSENSE_P_Pin;
+        GPIO_InitStruct.Pin  = TS_VSENSE_P_Pin | TS_VSENSE_N_Pin;
         GPIO_InitStruct.Mode = GPIO_MODE_ANALOG;
         GPIO_InitStruct.Pull = GPIO_NOPULL;
-        HAL_GPIO_Init(TS_VSENSE_P_GPIO_Port, &GPIO_InitStruct);
+        HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
```
pasitlaothamatas, line 116: dont forget to name this to TS_VSENSE_P on cube

boards/BMS/BMS.ioc
```diff
@@ -13,17 +13,18 @@ ADC1.EnableInjectedConversion=ENABLE
...
+ADC1.IPParameters=ClockPrescaler,Resolution,DataAlign,ScanConvMode,ContinuousConvMode,DiscontinuousConvMode,DMAContinuousRequests,EOCSelection,Overrun,LowPowerAutoWait,EnableRegularConversion,NbrOfConversion,ExternalTrigConv,ExternalTrigConvEdge,EnableInjectedConversion,InjNumberOfConversion,EnableAnalogWatchDog1,EnableAnalogWatchDog2,EnableAnalogWatchDog3,Rank-0\#ChannelRegularConversion,Channel-0\#ChannelRegularConversion,SamplingTime-0\#ChannelRegularConversion,OffsetNumber-0\#ChannelRegularConversion,Offset-0\#ChannelRegularConversion,master,SingleDiff-0\#ChannelRegularConversion,SubFamily
 ADC1.InjNumberOfConversion=0
 ADC1.LowPowerAutoWait=DISABLE
 ADC1.NbrOfConversion=1
-ADC1.Offset-0\#ChannelRegularConversion=0
-ADC1.OffsetNumber-0\#ChannelRegularConversion=ADC_OFFSET_NONE
+ADC1.Offset-0\#ChannelRegularConversion=1983
```
pasitlaothamatas, line 20: if possible give this constant a descriptive name on cube as well, maybe something like `TS_VOLTAGE_OFFSET`

boards/BMS/Src/App/states/App_InitState.c
```diff
@@ -33,14 +38,14 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
-            App_SharedClock_GetPreviousTimeInMilliseconds(clock) >=
-        5000U)
+    // if AIR- closes and the TS voltage is below 5V then begin precharge
+    if (Io_Airs_IsAirNegativeClosed() &&
+        Io_VoltageSense_GetTractiveSystemVoltage() < 5)
```
pasitlaothamatas, line 43: how did we decide here on 5V? is this an arbitrarily chosen voltage?

boards/shared/Src/Io/Io_SharedAdc.c
```diff
@@ -7,6 +7,15 @@ float Io_SharedAdc_ConvertRawAdcValueToVoltage(
 {
     uint32_t full_scale;
 
+    float single_diff_scaling_ratio = 6.6f;
+    // TODO: Add switch statement that grabs adc single/diff setting and change
```
pasitlaothamatas, line 11: we can probably do this for the current PR

pasitlaothamatas:
@Eddy13013  Gave ur PR a quick review and left some comments. Can u update the scope of ur PR? I think most of the description is missing

boards/BMS/Src/App/states/App_PreChargeState.c
pasitlaothamatas: * shared macros

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -1,14 +1,40 @@
...
+#include "states/App_InitState.h"
 
 #include "App_SetPeriodicCanSignals.h"
 #include "App_SharedMacros.h"
 
+#define S_TO_MS (1000U)
+
+//expected charge time (to 98%) = 4RC * 1000ms/s
+#define PRECHARGE_RESISTANCE (1000U)
+#define PRECHARGE_CAPACITANCE (0.68e-3f)
```
timbitss, line 21: Just curious, where did you get PRECHARGE_RESISTANCE AND PRECHARGE_CAPACITANCE from?

pasitlaothamatas: Precharge resistance is the precharge resistor 1Kohm. Don't think that we had measured the bus cap yet so this value may be a value from the previous car

boards/BMS/BMS.ioc
Eddy13013: We did not. Is it better to have lower or higher sample times from an MCU standpoint?

boards/BMS/Src/App/states/App_PreChargeState.c
Eddy13013: this changes every 100Hz

boards/BMS/Src/App/states/App_PreChargeState.c
Eddy13013: done

boards/BMS/Src/App/states/App_PreChargeState.c
Eddy13013: we dont want to be re-entering precharge over and over tho

boards/BMS/Src/App/states/App_InitState.c
Eddy13013: done

boards/BMS/Inc/Io/Io_Adc.h
Eddy13013: no we dont. removed it

boards/BMS/Src/App/states/App_InitState.c
Eddy13013: Yeah just figured 5V was low enough

boards/BMS/Src/App/states/App_PreChargeState.c
Eddy13013: this is the testing cap I have been using. The actual cap is 0.56e-3. I will change it once I verify the code

boards/BMS/Src/App/states/App_InitState.c
Eddy13013: 5V was low enough in my mind and isn't greatly affected by measurement error. Our voltage readings have pretty large error and low voltages

boards/BMS/Src/stm32f3xx_hal_msp.c
Eddy13013: not sure where this is. TS_VSENSE is no longer on a GPIO pin. Its on ADC1 now

boards/shared/Src/Io/Io_SharedAdc.c
Eddy13013: you will need to explain to me how to access variables from cube in our code

boards/BMS/Src/App/states/App_PreChargeState.c
pasitlaothamatas: ahhh factsss

boards/BMS/Src/stm32f3xx_hal_msp.c
pasitlaothamatas: This should be a user constant specified in cubeMX

boards/BMS/Src/App/states/App_PreChargeState.c
pasitlaothamatas: yup thats right but when this function is getting called its not changing so its nice to keep it as a const

boards/BMS/Src/App/states/App_PreChargeState.c
pasitlaothamatas: were dont have to be in precharge over and over, this is more or less initializing our state variable. @timbitss did something similar here

boards/BMS/Src/App/states/App_InitState.c
pasitlaothamatas: okie

boards/BMS/Src/App/states/App_PreChargeState.c
pasitlaothamatas: Check out how the state variable is updated on his current PR `DriveStateRunOnTick100Hz`

boards/BMS/BMS.ioc
pasitlaothamatas: hm ic, lets see toms if there is a way to do this

boards/BMS/BMS.ioc
pasitlaothamatas: image

This will affect our total conversion time, better to have a lower sample time

timbitss: We may introduce more inaccuracies if we lower the sample time all the way to 1.5 ADC clock cycles. I think it's best to leave it at around 7.5 clock cycles to be safe. See [Sec 4.2.6](https://www.st.com/resource/en/application_note/cd00211314-how-to-get-the-best-adc-accuracy-in-stm32-microcontrollers-stmicroelectronics.pdf)

timbitss:
inactive

PR #736, opened 2022-03-13
Add app code for cell voltage readings, remove cell temp and die temp temporarily

pasitlaothamatas:
### Summary

Wanted to get new app code integrated so i can test io cell temp and io die temp more easily

- Add app code for cell voltage measurement
- Remove cell temp and die temp measurements
- Remove can messages associated with removed cell temp and die temp measurements
- Binary status is pretty unnecessary imo, not used currently and should not be used in the future. Removed

- Create new directory for LTC related IO files
- Added documentation and small fixes for Io_CellVoltages

### Testing Done

- tested that cell voltages for a single segment can be read back
- tested that dcc bits can still be written to

### Checklist


boards/BMS/Inc/App/App_Accumulator.h
```diff
@@ -1,185 +1,48 @@
...
  */
 struct Accumulator *App_Accumulator_Create(
-    ExitCode (*configure_cell_monitors)(void),
-    ExitCode (*read_cell_voltages)(void),
-
+    bool (*configure_cell_monitors)(void),
+    bool (*start_voltage_conv)(void),
+    bool (*read_cell_voltages)(void),
+    void (*get_min_cell_location)(uint8_t *, uint8_t *),
+    void (*get_max_cell_location)(uint8_t *, uint8_t *),
```
timbitss, line 32: What are your thoughts on including the names of the function pointer arguments? It's not clear what the two uint8_t ptrs are supposed to represent

boards/BMS/Inc/App/App_SetPeriodicCanSignals.h
```diff
@@ -5,12 +5,3 @@
...
     struct Imd *              imd);
-
-void App_SetPeriodicSignals_AccumulatorInRangeChecks(
-    struct BmsCanTxInterface *can_tx,
-    const struct Accumulator *accumulator,
-    struct ErrorTable *       error_table);
-
-void App_SetPeriodicSignals_CellMonitorsInRangeChecks(
-    struct BmsCanTxInterface * can_tx,
-    const struct CellMonitors *cell_monitors);
```
timbitss, line 16: Why was this deleted?

timbitss (commented):
thoughts??

boards/BMS/Inc/App/App_SetPeriodicCanSignals.h
pasitlaothamatas: Bc we shouldnt use in range checks, i think theyre a little extra

boards/BMS/Inc/App/App_Accumulator.h
pasitlaothamatas: i think we can refer to the io function that we currently pass in right now

boards/BMS/Inc/App/App_SetPeriodicCanSignals.h
pasitlaothamatas: I also want to wrap these using functions from App_Accumulator, don't think we need an entirely new file containing functions to broadcast CAN messages

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
```diff
@@ -45,21 +45,21 @@ enum RegGroupByteFormat
  * @param tx_data The command or register group to pack PEC15 bytes
  * @param size The size of the command or register group to transmit
  */
```
timbitss, line 47: rename to `num_bytes` and update comment

maybe mention that it's the number of bytes to transmit without the PEC15 since from first glance I would think it's the size of the tx_data buffer

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
```diff
@@ -68,22 +68,22 @@ void Io_NewLTC6813_InitSpiHandle(SPI_HandleTypeDef *spi_handle);
  * PEC15 bytes from
  * @return The PEC15 code generated from the contents of data_buffer
  */
-uint16_t Io_NewLTC6813_CalculateRegGroupPec15(uint8_t *data_buffer);
+uint16_t Io_SharedLTC6813_CalculateRegGroupPec15(uint8_t *data_buffer);
```
timbitss, line 71: should make pointer const since we're not modifying the data buffer

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
```diff
@@ -45,21 +45,21 @@ enum RegGroupByteFormat
...
  */
-void Io_NewLTC6813_PrepareCmd(uint16_t *tx_cmd, uint16_t cmd);
+void Io_SharedLTC6813_PrepareCmd(uint16_t *tx_cmd, uint16_t cmd);
 
 /**
  * Initialize the SPI handle used to communicate with the LTC6813
  * @param spi_handle The given SPI handle for the LTC6813 daisy chain.
  */
-void Io_NewLTC6813_InitSpiHandle(SPI_HandleTypeDef *spi_handle);
+void Io_SharedLTC6813_InitSpiHandle(SPI_HandleTypeDef *spi_handle);
```
timbitss, line 62: I think we should put the init function at the top as an indication that it should called on startup

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -1,160 +1,89 @@
...
+    float (*get_max_cell_voltage)(void);
+    float (*get_segment_voltage)(AccumulatorSegments_E);
+    float (*get_pack_voltage)(void);
+    float (*get_avg_voltage)(void);
 };
 
 struct Accumulator *App_Accumulator_Create(
-    ExitCode (*configure_cell_monitors)(void),
-    ExitCode (*read_cell_voltages)(void),
+    bool (*configure_monitoring_chip)(void),
```
timbitss, line 28: parameter name `configure_monitoring_chip` should match declaration in header file, could be more explicit that it's the LTC chip in particular

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
```diff
@@ -182,7 +190,7 @@ static void Io_PrepareCfgRegBytes(
     // cell location
     uint8_t min_cell_segment = 0U;
     uint8_t min_cell_index   = 0U;
-    Io_NewCellVoltages_GetMinCellLocation(&min_cell_segment, &min_cell_index);
+    Io_CellVoltages_GetMinCellLocation(&min_cell_segment, &min_cell_index);
 
     // Write to the last segment on the daisy chain first
```
timbitss, line 195: Does the last segment on the daisy chain correspond to segment 0?

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
```diff
@@ -233,48 +242,48 @@ void Io_NewLTC6813_InitSpiHandle(SPI_HandleTypeDef *spi_handle)
...
 {
-    return Io_NewLTC6813_CalculatePec15(
+    return Io_SharedLTC6813_CalculatePec15(
         data_buffer, NUM_REG_GROUP_PAYLOAD_BYTES);
 }
 
-void Io_NewLTC6813_PackPec15(uint8_t *tx_data, uint8_t num_bytes)
+void Io_SharedLTC6813_PackPec15(uint8_t *tx_data, uint8_t num_bytes)
 {
     const uint16_t cfg_reg_a_pec15 =
```
timbitss, line 253: I think naming the var `cfg_reg_a_pec15` is a bit unclear, simply naming it `pec15` would make more sense

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -24,11 +25,12 @@ void App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
     struct OkStatus *         bms_ok      = App_BmsWorld_GetBmsOkStatus(world);
     struct OkStatus *         imd_ok      = App_BmsWorld_GetImdOkStatus(world);
     struct OkStatus *         bspd_ok     = App_BmsWorld_GetBspdOkStatus(world);
-    struct Accumulator *      accumulator = App_BmsWorld_GetAccumulator(world);
     struct Airs *             airs        = App_BmsWorld_GetAirs(world);
+    struct Accumulator *      accumulator = App_BmsWorld_GetAccumulator(world);
     struct ErrorTable *       error_table = App_BmsWorld_GetErrorTable(world);
 
     App_SetPeriodicCanSignals_Imd(can_tx, imd);
+    App_Accumulator_RunOnTick100Hz(accumulator, can_tx);
```
pasitlaothamatas, line 33: whoops need to move this

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
```diff
@@ -233,48 +242,48 @@ void Io_NewLTC6813_InitSpiHandle(SPI_HandleTypeDef *spi_handle)
...
-        Io_NewLTC6813_CalculatePec15(tx_data, num_bytes);
+        Io_SharedLTC6813_CalculatePec15(tx_data, num_bytes);
 
     // Pack the PEC15 byte into tx_data in big endian format
     *((uint16_t *)(tx_data + num_bytes)) =
         CHANGE_WORD_ENDIANNESS(cfg_reg_a_pec15);
 }
 
-void Io_NewLTC6813_PrepareCmd(uint16_t *tx_cmd, uint16_t cmd)
+void Io_SharedLTC6813_PrepareCmd(uint16_t *tx_cmd, uint16_t cmd)
```
timbitss, line 261: What do you think of renaming it to `Io_SharedLTC6813_PrepareCmdWithPEC15`

boards/BMS/Src/Io/Io_LTC6813/Io_CellVoltages.c
```diff
@@ -176,7 +190,7 @@ bool Io_NewCellVoltages_ReadVoltages(void)
         {
             // Prepare the command used to read data back from a register group
             uint16_t tx_cmd[NUM_OF_CMD_WORDS] = { 0U };
-            Io_NewLTC6813_PrepareCmd(tx_cmd, cv_read_cmds[curr_reg_group]);
+            Io_SharedLTC6813_PrepareCmd(tx_cmd, cv_read_cmds[curr_reg_group]);
 
             // Transmit the command and receive data stored in register group
             if (Io_SharedSpi_TransmitAndReceive(
```
timbitss, line 196: Could we possibly wrap `Io_SharedSpi_TransmitAndReceive()` into `Io_SharedLTC6813`? e.g. `Io_SharedLTC6813_SendReadCmd()`. That way, we won't need to expose the SPI handle outside of `Io_SharedLTC6813.c`

timbitss (commented):
Looks good. See comments below

boards/shared/Inc/App/App_SharedSetPeriodicCanSignals.h
```diff
@@ -35,21 +35,3 @@
...
-    {                                                                        \
-        if (App_SharedBinaryStatus_IsActive(binary_status))                  \
-        {                                                                    \
-            can_signal_setter(can_tx, on_choice);                            \
-        }                                                                    \
-        else                                                                 \
-        {                                                                    \
-            can_signal_setter(can_tx, off_choice);                           \
-        }                                                                    \
-    }
```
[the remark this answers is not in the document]
pasitlaothamatas: Sure i can do that. Theyre currently not used in the BMS project, and i dont plan on integrating them in the future

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -80,55 +80,6 @@ SG_ AIR_NEGATIVE : 1|1@1+ (1,0) [0|1] "" FSM,DCM,PDM
...
-
-BO_ 127 BMS_CELL_MONITOR_4: 4 BMS
-SG_ CELL_MONITOR_4_DIE_TEMPERATURE : 0|32@1+ (1,0) [0.0|120.0] "degC" DEBUG
-
-BO_ 128 BMS_CELL_MONITOR_5: 4 BMS
-SG_ CELL_MONITOR_5_DIE_TEMPERATURE : 0|32@1+ (1,0) [0.0|120.0] "degC" DEBUG
-
-BO_ 129 BMS_MAX_CELL_MONITOR: 4 BMS
-SG_ MAX_CELL_MONITOR_DIE_TEMPERATURE : 0|32@1+ (1,0) [0.0|120.0] "degC" DEBUG
-
```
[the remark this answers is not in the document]
pasitlaothamatas: die temp and cell temp code, as well as the app level wrapper needs to be updated in future PR. They are removed for the current PR so these can messages should also be removed. I can add this to the PR description

boards/BMS/Src/Io/Io_LTC6813/Io_CellVoltages.c
```diff
@@ -59,10 +62,21 @@ static uint16_t        cell_voltages[NUM_OF_ACCUMULATOR_SEGMENTS]
...
+ */
+static void Io_UpdateCellVoltages(void);
+
+/**
+ * Parse cell voltage from all segments to be stored in cell_voltages array
+ * @param curr_reg_group The current register group to read cell voltages from
+ * @param rx_buffer The buffer containing data read from the LTC6813 chips
+ * @return True if the data is read back successfully. Else, false
+ */
+static bool Io_ParseCellVoltageFromAllSegments(
```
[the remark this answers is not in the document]
pasitlaothamatas: sure
```suggestion
static bool Io_ParseCellVoltagesFromAllSegments(
```

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
pasitlaothamatas: ah yea, good catch. looks like some leftover var names i can fix this.

boards/BMS/Src/Io/Io_LTC6813/Io_CellVoltages.c
pasitlaothamatas: Yeah im thinking about this too. We can actually get rid of shared spi entirely and implement SPI commands within the IO_SharedLTC file since the BMS is the only board that uses it. Old PDM needed the SPI driver, but new PDM no longer needs a SPI interface. Lets leave this for another PR though. I think its out of scope rn 🅿️🅿️🅿️

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
pasitlaothamatas: ah yeah nice catch, looks like some comments i forgot to update

boards/BMS/Src/App/App_Accumulator.c
pasitlaothamatas: ah yeah need to change this in the struct as well. i feel like running misra checks will help us a lot with detecting these. we should look into maybe doing this at some point

pasitlaothamatas: i would prefer to keep sensor information abstracted from these fxn ptrs though, since the app layer shouldnt really have sensor specific information. Rest of the app code for the vehicle follows this convention so it think its also nice to keep things consistent

Taken from the FSM for example, specific sensor information is abstracted here.
image

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
pasitlaothamatas: yeah i think i see where your confusion is coming from

this is rlly stupid i think i have a better way to do this. stay tuned
image

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
pasitlaothamatas: this isnt just preparing with pec15 though. its changing the cmd endianness and then doing the pec15 check.

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
pasitlaothamatas: hm ya agree

boards/BMS/Src/Io/Io_LTC6813/Io_CellVoltages.c
timbitss: LOL

boards/BMS/Src/Io/Io_LTC6813/Io_SharedLTC6813.c
pasitlaothamatas: implementation should be much nicer now

boards/BMS/Inc/Io/Io_LTC6813/Io_SharedLTC6813.h
pasitlaothamatas: im assuming you meant pointer to a `const uint8_t` here, added

pasitlaothamatas:
@timbitss Went over the comments, should all be addressed. 

also did the following:
- Renamed the Io files
- removed some unnecessary fxns in the `Io_LTC6813Shared` files
- Cleaned up `#defines` and `enums` in `Io_LTC6813` dir
- Clean up some unnecessary CAN msgs in the dbc
- Reversed the endianness of the commands used to talk to the LTC chip. Better to store them in their reversed endianness rather than storing them similarly to how they are presented in the datasheet as we can avoid swapping the endianness in runtime imo

timbitss (approved):
Good work :)

pasitlaothamatas:
> Good work :)



> Good work :)

lfg

pasitlaothamatas:
> Good work :)



> Good work :)

lfg

PR #744, opened 2022-03-26
TS Voltage ADC configuration

pasitlaothamatas:
### Summary

Prepping ADC configuration for precharging
Using values experimentally determined by edwin for scaling + offset
We still need to investigate why the offset at 0v is 1983. Think this is due to the TS_VSENSE_N pin and GND not being at the same potential

### Changelist 

- Configure ADC to be differential mode
- Update unit test for Test_VoltageSense.cpp
- Tweak the resistor divider configurations and also add compensation for resistor errors (experimentally determined by edwin)

### Testing Done

Measured with 60V on power supply

### Checklist


boards/BMS/Test/Src/Test_VoltageSense.cpp
```diff
@@ -8,22 +8,17 @@ extern "C"
...
         adc_voltage, Io_VoltageSense_GetTractiveSystemVoltage(adc_voltage));
 
     // Nominal tractive system voltage (400V)
     constexpr float nominal_ts_voltage = 400.0f;
     adc_voltage =
         nominal_ts_voltage * 8.0f *
-        (1.024e+3f / (1.024e+3f + 499e+3f + 499e+3f + 499e+3f + 499e+3f));
-    ASSERT_EQ(
+        (1.24e+3f / (1.24e+3f + 499e+3f + 499e+3f + 499e+3f + 499e+3f)) / 1.05;
+    ASSERT_FLOAT_EQ(
```
gtaharaedmonds, line 21: would prefer this to be made constants since the same values are reused from Io_VoltageSense.c
Also its not really clear what these values represent, I'm assuming they're the series resistors?

gtaharaedmonds (approved):
voltage ratio constants could be bit clearer w more explanation what they refer to, but otherwise looks good to me

pasitlaothamatas:
> voltage ratio constants could be bit clearer w more explanation what they refer to, but otherwise looks good to me

good point ill do this

PR #750, opened 2022-04-17
BMS: Complete CHARGE state functionality

timbitss:
### Summary
This PR will hopefully ensure that the BRUSA NLG513 does start charging the battery pack when it enters the CHARGE state by asserting the `PON` pin. The `FLT` pin is also monitored in case of a fault during the charging process. Two CAN signals are added to indicate when the charging process has been completed and whether a charging fault has occurred.

The current charging procedure is that the BMS transitions from the PRE-CHARGE state to the CHARGE state if the charger is connected to an outlet. While the BMS is in the CHARGE state and both AIRs are closed, an output pin is asserted to enable the charging process. Once the maximum cell voltage reaches a defined threshold, the BMS enters the FAULT state to indicate that the charger should be disconnected and the BMS should be power-cycled.

### Changelist 
- Enable automatic charging within the CHARGE state
- Transition from the CHARGE to the FAULT state when charging is done or if a fault has occurred during charging
- Added CAN msgs for debugging
- Renamed PA11 GPIO label from "GPIO1" to "BRUSA_FLT" for clarity

### Testing Done
- Nothing yet :P

### Checklist


scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -315,6 +315,7 @@ SG_ STACK_WATERMARK_ABOVE_THRESHOLD_TASK1KHZ : 1|1@1+ (1,0) [0|1] "" DEBUG
 SG_ STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANRX : 2|1@1+ (1,0) [0|1] "" DEBUG
 SG_ STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANTX : 3|1@1+ (1,0) [0|1] "" DEBUG
 SG_ WATCHDOG_TIMEOUT : 4|1@1+ (1,0) [0|1] "" DEBUG
+SG_ CHARGING_DONE : 5|1@1+ (1,0) [0|1] "" DEBUG
```
pasitlaothamatas, line 318: shouldnt we see an indication that this is complete when we are no longer in the charge state? I dont think this is needed

pasitlaothamatas:
Has this been tested with the charger @timbitss ?

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -2,13 +2,22 @@
 #include "states/App_ChargeState.h"
 #include "states/App_FaultState.h"
 
-#include "App_SetPeriodicCanSignals.h"
 #include "App_SharedMacros.h"
 
+#define TOTAL_CHARGE_TIME_S (3600U) // 1 hr
```
pasitlaothamatas, line 7: how was this time defined?

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,18 +34,42 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+
+    // Track the time since accumulator charging started
+    const uint32_t elapsed_time_s =
+        App_SharedClock_GetCurrentTimeInSeconds(clock) -
+        App_SharedClock_GetPreviousTimeInSeconds(clock);
+
+    // Go to the fault state when charging has completed to indicate that
+    // the charger should be disconnected and the BMS-Core should be
+    // power-cycled
+    if (elapsed_time_s >= TOTAL_CHARGE_TIME_S)
```
pasitlaothamatas, line 60: is this how we are defining when we have reached 100% soc? I think this may not be the best way to determine that we are fully charged up

pasitlaothamatas: I think doing a check to see if the MAX voltage has reached 4.2V or something might be a better approach instead of using a time limit

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,18 +34,42 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+    {
+        App_CanTx_SetPeriodicSignal_CHARGING_DONE(can_tx, true);
+        App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
```
pasitlaothamatas, line 63: I think we are supposed to transition to the init state

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: We should see something like this if the charger is connected: init->precharge->charge->init

So if we monitor the state variable on CAN we will have this information while debugging

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,18 +34,42 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
             can_tx, true);
         App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
     }
+
+    if (App_Charger_HasFaulted(charger))
+    {
+        App_CanTx_SetPeriodicSignal_CHARGER_FAULT_DETECTED(can_tx, true);
+        App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+    }
+
```
pasitlaothamatas, line 51: Most of the if statements here direct us to the fault state. In some cases certain sections of code do not need to be executed either. See if we can make this nicer :)

boards/BMS/Src/Io/Io_Charger.c
```diff
@@ -17,3 +17,9 @@ bool Io_Charger_IsConnected(void)
     return HAL_GPIO_ReadPin(CHARGE_STATE_GPIO_Port, CHARGE_STATE_Pin) ==
            GPIO_PIN_SET;
 }
+
+bool Io_Charger_HasFaulted(void)
+{
+    return HAL_GPIO_ReadPin(BRUSA_FLT_GPIO_Port, BRUSA_FLT_Pin) ==
+           GPIO_PIN_RESET;
+}
```
pasitlaothamatas, line 25: dont forget to add a newline

boards/BMS/Src/App/states/App_ChargeState.c
pasitlaothamatas: Might be a good idea to allow transition to the init state when we have completed charging AND physically disconnect the charger to avoid the possibility of cycling back into the charge state

boards/BMS/Src/App/states/App_ChargeState.c
timbitss: arbitrary, we weren't too sure how to determine when charging is complete, but I think your suggestion of tracking the pack voltage is the most logical

boards/BMS/Src/App/states/App_ChargeState.c
timbitss: agreed

boards/BMS/Src/App/states/App_ChargeState.c
timbitss: > Has this been tested with the charger @timbitss ?

no we plan to asap once joe and edwin get the wiring resolved

timbitss: > Might be a good idea to allow the transition to the init state when we have completed charging AND physically disconnect the charger to avoid the possibility of cycling back into the charge state

Yes, I think that definitely makes the most sense. I, therefore, believe we need to keep the "charging complete" CAN signal to indicate to the user they should disconnect the charger. What do you think about adding another CAN signal with the status of the PON pin? I don't think it hurts IMO, helps for debugging

scripts/codegen/CAN/App_CanMsgs.dbc
timbitss: To connect the comments and suggestions below, I think it's best that we stay in the charge state once charging is complete when the pack voltage is ~400 V. Then, set the CAN signal "CHARGING_COMPLETE". The user will then disconnect the charger which will put the BMS from charge->init state. "CHARGING_COMPLETE" should be cleared as well

Eddy13013: I agree with tim here. We don't want to re-enter init state then go through precharge again and back to charge state in a loop.

boards/BMS/Src/App/states/App_ChargeState.c
Eddy13013: determining a completed charging should be done through future SOC firmware right?

boards/BMS/Src/App/states/App_ChargeState.c
pasitlaothamatas: Sure yeah having PON would be good for debugging.

boards/BMS/Src/App/states/App_ChargeState.c
pasitlaothamatas: that or just checking if the max cell voltage has reached 4.2V

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: hm yeah i can see how the charging complete would be nice to inform the user that the charger needs to be removed

pasitlaothamatas: @Eddy13013 are there any intuitive ways to inform the user that we have finished charging?

boards/BMS/Src/App/states/App_ChargeState.c
pasitlaothamatas: to be more specific, we dont have to check for elapsed time when the charger has faulted or if the charger is not connected

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: @timbitss why is charging done regarded as an AIR shdn fault?

boards/BMS/Src/Io/Io_Charger.c
timbitss: done

scripts/codegen/CAN/App_CanMsgs.dbc
timbitss: I'll just make a separate can msg for it

boards/BMS/Src/App/states/App_ChargeState.c
Eddy13013: If we are to follow the rules for charging. We will need to charge on the cart. This means that in order to put the accumulator back in the car we will need to power cycle the BMS anyways (power comes from the charger lemo). I feel that even when charging on the car, a power cycle would be a good idea. It's an easy way to avoid further accidental precharge scenarios. For example, say we finish charging and enter into the init state, if the shutdown loop is closed again, we could re-enter the charging state accidentally. I think we should try to avoid these accidental precharges when changing connectors to the accumulator so a power off would be the best option.

PR #753, opened 2022-04-23
DCM Quick Fixes: Load switches, reverse wheel direction, init->drive checks, torque limit

timbitss:
### Summary
-  DCM can only enter drive state if the BMS is in the drive state (AIRs closed, charger disconnected, pre-charge completed) + other checks
-  Turn on inverter load switches directly on entry to init state since inverter fault no longer occurs if the inverters see LV before HV as the `Precharge_Bypassed_EEPROM` parameter is now set to 1
-  Reverse wheel direction of the right inverter (both motors need to be spinning in the same direction)
-  Lower maximum torque request limit to 50.0 Nm for safety until proper current-torque algo is determined

### Testing Done
- Tested on car, faking signals using PCAN

### Checklist

### DIFF
--- a/boards/DCM/Inc/App/configs/App_TorqueRequestThresholds.h
+++ b/boards/DCM/Inc/App/configs/App_TorqueRequestThresholds.h
@@ -2,4 +2,4 @@
 
 // Maximum torque request
 // TODO: Have separate maximum torque requests for motoring and generating
-#define MAX_TORQUE_REQUEST_NM 50.0f;
+#define MAX_TORQUE_REQUEST_NM (50.0f);
--- a/boards/DCM/Inc/App/configs/App_TorqueRequestThresholds.h
+++ b/boards/DCM/Inc/App/configs/App_TorqueRequestThresholds.h
@@ -1,9 +1,5 @@
 #pragma once
 
-/**
- * @brief Maximum torque request to get the maximum motor current per Emrax
- * specification.
- *
- * @see Setting up the Controller to run with Emrax Motor Rev 1.2 PDF pg. 6
- */
-#define MAX_TORQUE_REQUEST_NM 132.0f;
+// Maximum torque request
+// TODO: Have separate maximum torque requests for motoring and generating
+#define MAX_TORQUE_REQUEST_NM 50.0f;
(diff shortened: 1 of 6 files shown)


Eddy13013 (approved):
I think this looks good. I'm not confident enough with the syntax of some of our CAN messages but the general process seems correct to me.

boards/DCM/Inc/App/configs/App_TorqueRequestThresholds.h
```diff
@@ -1,9 +1,5 @@
...
-/**
- * @brief Maximum torque request to get the maximum motor current per Emrax
- * specification.
- *
- * @see Setting up the Controller to run with Emrax Motor Rev 1.2 PDF pg. 6
- */
-#define MAX_TORQUE_REQUEST_NM 132.0f;
+// Maximum torque request
+// TODO: Have separate maximum torque requests for motoring and generating
+#define MAX_TORQUE_REQUEST_NM 50.0f;
```
pasitlaothamatas, line 5:
```suggestion
#define MAX_TORQUE_REQUEST_NM (50.0f);
```

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -588,6 +588,7 @@ BA_DEF_DEF_  "GenSigStartValue" 0;
 BA_ "BusType" "CAN";
 
 BA_ "GenSigStartValue" SG_ 507 Start_Switch 1; 
+BA_ "GenSigStartValue" SG_ 82 Direction_Command_INVR 1;
```
pasitlaothamatas, line 591: would be nice to put these in order (82 before 507)

pasitlaothamatas:
lgtm pending changes requested @timbitss

scripts/codegen/CAN/App_CanMsgs.dbc
pasitlaothamatas: actually how does this actually work? does it automatically set the value in the can table to 1?

timbitss: > actually how does this work? does it automatically set the value in the can table to 1?

Yep

timbitss: > would be nice to put these in order (82 before 507)

resolved

PR #757, opened 2022-04-29
Detect PEC15 error and log when comms error occurs

pasitlaothamatas:
### Summary

- Detect three consecutive PEC15 errors and throw air shdn error (BMS-3)
- Transition to the fault state

### Changelist 

- Optimized code for BMS_OK, BSPD_OK, and IMD_OK
- remove can_tx_interface as an input from Accumulator_RunOnTick, feel that it is better to leave can tx interface within the scope of the state machine for now

### Testing Done

- Disconnect isoSPI header and validated that we transition into the fault state when a PEC15 error/no communication occurs

### Checklist


boards/BMS/Src/Io/Io_LTC6813/Io_LTC6813CellVoltages.c
```diff
@@ -196,17 +196,12 @@ bool Io_LTC6813CellVoltages_ReadVoltages(void)
...
+                !Io_ParseCellVoltageFromAllSegments(curr_reg_group, rx_buffer))
             {
-                // Parse cell voltages from the current register group from all
-                // segments
-                if (!Io_ParseCellVoltageFromAllSegments(
-                        curr_reg_group, rx_buffer))
-                {
-                    return false;
-                }
+                return false;
```
alxshev, line 204: Checking whether exiting the method here is intentional -- if you want to continue reading data from the remaining register groups after a failure this should be setting `status=false` instead.

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -335,10 +335,11 @@ SG_ State : 0|8@1+ (1,0) [0|255] "" DCM
 BO_ 108 BMS_STATE_OF_CHARGE: 4 BMS
 SG_ State_Of_Charge : 0|32@1- (1,0) [0|1E2] "%" DIM
 
-BO_ 109 BMS_AIR_SHUTDOWN_ERRORS: 8 BMS
+BO_ 109 BMS_AIR_SHUTDOWN_ERRORS: 1 BMS
```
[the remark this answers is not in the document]
pasitlaothamatas: We are sending less than a byte of data over, so keep the payload length to a byte here

boards/BMS/Src/Io/Io_LTC6813/Io_LTC6813CellVoltages.c
pasitlaothamatas: Actually i think my intention behind this was to discard the current snapshot of the voltages completely if a PEC15 mismatch occurs. Might not be necessary to do this though

alxshev: Yeah ok I figured that might be the case. Would be helpful to add a comment somewhere (here or header file) explaining how partial failures are handled

pasitlaothamatas: should be nicer now @alxshev

alxshev (approved):

PR #759, opened 2022-04-30
Add missing HB callback for DIM

pasitlaothamatas:
### Summary

- Modify heartbeat monitor to support DIM callback when HB is not detected
- Add board IDs. Refer to `App_SharedHeartbeatMonitor.h` to see how these are currently implemented

### Testing Done

- add #define for NDEBUG, test that we transition into the fault state and that 888 is displayed on the sseg display

### Checklist

### DIFF
--- a/boards/shared/Src/App/App_SharedHeartbeatMonitor.c
+++ b/boards/shared/Src/App/App_SharedHeartbeatMonitor.c
@@ -46,7 +46,7 @@ bool App_SharedHeartbeatMonitor_Tick(struct HeartbeatMonitor *const heartbeat_mo
 
 #ifdef NDEBUG
         // Check if the board received all the heartbeats it's listening for
-        status = (heartbeat_monitor->heartbeats_to_check == heartbeat_monitor->heartbeats_checked_in);
+        status = heartbeat_monitor->heartbeats_to_check == heartbeat_monitor->heartbeats_checked_in;
 #endif
 
         // Clear the list of boards that have checked in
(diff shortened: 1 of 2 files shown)


boards/shared/Src/App/App_SharedHeartbeatMonitor.c
```diff
@@ -79,3 +87,18 @@ void App_SharedHeartbeatMonitor_CheckIn(
 {
     heartbeat_monitor->heartbeats_checked_in |= heartbeat;
 }
+
+#if (BOARD_ID == DIM_ID)
```
kirons96, line 91: this indicates to me that this should be in a component specific file i.e. DIM file, maybe with a weak function or something

pasitlaothamatas: I think using a function pointer here might not be the best way to invoke your callback

pasitlaothamatas: Might be much easier to read if we had the heartbeat tick function return false when a heartbeat signal is not received. We can implement the "callback" in the state machine

pasitlaothamatas: Rlly wanna be careful with fxn ptrs for next year and make sure we use it only when its needed. Makes the code rlly hard to follow tbh

pasitlaothamatas:
image

PR #771, opened 2022-05-08
BRUSA Charging 2.0

timbitss:
### Summary
This PR will hopefully ensure that the BRUSA NLG513 does start charging the battery pack when it enters the CHARGE state by asserting the `PON` pin. The `FLT` pin is also monitored in case of a fault during the charging process. Two CAN signals are added to indicate when the charging process has been completed and whether a charging fault has occurred.

The current charging procedure is that the BMS transitions from the PRE-CHARGE state to the CHARGE state if the charger is connected to an outlet. While the BMS is in the CHARGE state and both AIRs are closed, an output pin is asserted to enable the charging process. Once the maximum cell voltage reaches a defined threshold, the BMS enters the FAULT state to indicate that the charger should be disconnected and the BMS should be power-cycled.

### Changelist 
- Enable automatic charging within the CHARGE state
- Transition from the CHARGE to the FAULT state when charging is done or if a fault has occurred during charging
- Added CAN msgs for debugging
- Renamed PA11 GPIO label from "GPIO1" to "BRUSA_FLT" for clarity

### Testing Done
- Nothing yet :P

### Checklist


boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -19,23 +28,47 @@ static void ChargeStateRunOnTick1Hz(struct StateMachine *const state_machine)
...
         if (!App_Charger_IsConnected(charger))
         {
             App_CanTx_SetPeriodicSignal_CHARGER_DISCONNECTED_IN_CHARGE_STATE(can_tx, true);
-            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
         }
+        if (App_Charger_HasFaulted(charger))
+        {
+            App_CanTx_SetPeriodicSignal_CHARGER_FAULT_DETECTED(can_tx, true);
+        }
+        if (App_Accumulator_GetMaxVoltage(accumulator, NULL, NULL) > MAX_CELL_VOLTAGE_THRESHOLD)
```
pasitlaothamatas, line 54: This is almost guaranteed to cause some sort of memory corruption. Gotta be careful when you pass in NULL as an arg

PR #782, opened 2022-05-15
Calibrate L & R resolvers, set recommended inverter params, Update Inverter msgs

timbitss:
### Summary
- Calibrate left and right resolvers
- Update EEPROM files with recommended motor params from the Setting Up The Emrax Motor V1.2 pdf 
- Change right wheel direction to CW, left wheel direction to CCW (facing the motor shaft from outside)
- Update fast msgs to 10 ms and slow inverter msgs to 100 ms
- Remove unused inverter CAN msgs from dbc file. See [excel sheet](https://docs.google.com/spreadsheets/d/1I4MKK22DZTSXN9jGVQjMx2Z-H5mULfPpAcJGnoBUxyc/edit?usp=sharing)

### Testing Done
- Tested on car, motors sound/run smoother

### Checklist


pasitlaothamatas:
Could you explain why we are reversing the direction for the right wheel?

pasitlaothamatas:
Lemme know n we should be good to merge

inverter-eeprom-config/saved_EEPROM_configs/left_inv
```diff
@@ -57,19 +57,19 @@ Coast_Lo_EEPROM_(V)_x_100             ,      0
 Coast_Hi_EEPROM_(V)_x_100             ,      4  
 Accel_Max_EEPROM_(V)_x_100            ,    490  
 Pedal_Hi_EEPROM_(V)_x_100             ,    499  
-Motor_Torque_Limit_EEPROM_(Nm)_x_10   ,    500  
+Motor_Torque_Limit_EEPROM_(Nm)_x_10   ,   1320
```
Eddy13013, line 60: Why do we have a torque limit set? Is this number at least much higher than we would request from the DCM?

timbitss: A torque limit of 132 Nm is set in the EEPROM based on the maximum current through the Emrax motor. The maximum torque request from the DCM is 90 Nm currently, so this is not a problem.

timbitss:
> Could you explain why we are reversing the direction for the right wheel?

It's a result of calibrating the resolver. Prior to resolver calibration, sending a forward rotation request was actually rotating the motor shaft in reverse which is not what the ac motor controller expects. See Section 2.2 Verify Resolver and Motor Direction in the Resolver calibration pdf for a better explanation

pasitlaothamatas:
> > Could you explain why we are reversing the direction for the right wheel?
> 
> It's a result of calibrating the resolver. Prior to resolver calibration, sending a forward rotation request was actually rotating the motor shaft in reverse which is not what the ac motor controller expects. See Section 2.2 Verify Resolver and Motor Direction in the Resolver calibration pdf for a better explanation

@timbitss Ok so i gave the doc a read through. The doc states that "The controller does not understand which direction for a vehicle is “Forward” and which is “Reverse”" but also mentions that "Generally for most motors the Forward direction is counter-clockwise rotation when facing the motor shaft."  (salutes to `ly for most motors the Forward direction is counter-clockwise rotation when facing the motor shaft` @kirons96  (thank you please clap))

So before calibrating, sending a positive torque request would turn the shaft counter clockwise (when facing the shaft)? After calibrating we have now defined the forward direction to be the clockwise direction?

pasitlaothamatas:
@timbitss Do we need to update for `GenMsgCycleTime` for the Torque command, motor speed, etc. CAN messages that we receive from the inverters?

timbitss:
> 
> So before calibrating, sending a positive torque request would turn the shaft counter clockwise (when facing the shaft)? After calibrating we have now defined the forward direction to be the clockwise direction?

After calibrating, we have now defined the forward direction to be in the **counter-clockwise** direction. 

`Direction_Command_INVR` = 0 means spin the motor in the reverse / clockwise direction which is what we want

timbitss:
> @timbitss Do we need to update for `GenMsgCycleTime` for the Torque command, motor speed, etc. CAN messages that we receive from the inverters?

Good catch. Functionally I don't think it affects anything but yes I will update those values

pasitlaothamatas (approved):
LGTM

PR #783, opened 2022-05-16
Implement current sense

Eddy13013:
### Summary
Current sense firmware completed.

### Changelist 

- IO functions changed to return floats just like voltage sense
- App code added to the App_TractiveSystem to get TS current
- CAN messages added to the DBC file
- Changed what "high_res" to mean high gain (i.e. 40mV/A) and "low_res" to mean low gain (i.e. 6.67mV/A)

### Testing Done

- Tested by running current from power supply through sensor terminals and reading CAN messages. Verified with an error of +/- 0.2A.

### Checklist


boards/BMS/Src/Io/Io_CurrentSense.c
```diff
@@ -1,13 +1,30 @@
...
+#include <math.h>
 #include "Io_CurrentSense.h"
 
-ExitCode Io_CurrentSense_ConvertToLowResolutionMainCurrent(float adc_voltage, float *low_res_main_current)
+// Offset voltage of output 1. Found to be 2.45V through testing
+#define OUTPUT_1_OFFSET (2.45f)
+// Sensitivity of output 1: 40mV/A
+#define OUTPUT_1_SENSITIVITY (40e-3f)
+// Voltage divider between sensor output and adc input.
+#define OUTPUT_1_DIV (2.155f / (1.1f + 2.155f))
```
pasitlaothamatas, line 10: `OUTPUT_1_DIV` is the same as `OUTPUT_2_DIV`. Can we use condense them into a single #define?

pasitlaothamatas:
Might want to remove the current sense unit test so your builds pass

boards/BMS/Src/Io/Io_CurrentSense.c
Eddy13013: From a firmware perspective that would make sense. But there is actually two different hardware circuits that may change in the future. That is my reason for keeping them separate.

pasitlaothamatas: sounds good w me

boards/BMS/Inc/Io/Io_CurrentSense.h
```diff
@@ -11,7 +11,7 @@
  * @return EXIT_CODE_OUT_OF_RANGE if adc_voltage is negative
  *         EXIT_CODE_INVALID_ARGS if low_res_main_current is NULL
```
pasitlaothamatas, line 12: i'd also update these since we're not returning `ExitCode` anymore

boards/BMS/Src/Io/Io_CurrentSense.c
```diff
@@ -30,22 +47,17 @@ ExitCode Io_CurrentSense_ConvertToLowResolutionMainCurrent(float adc_voltage, fl
     //
     // Sensitivity = 40mV/A
 
-    const float offset             = 2.5f;
-    const float sensitivity        = 40e-3f;
-    const float hsnbv_d06_output_1 = adc_voltage * (1e3f + 2.2e3f) / 2.2e3f;
+    const float hsnbv_d06_output_1 = adc_voltage * 1.0f / OUTPUT_1_DIV;
```
pasitlaothamatas, line 50:
```suggestion
    const float hsnbv_d06_output_1 = adc_voltage / OUTPUT_1_DIV;
```

boards/BMS/Src/Io/Io_CurrentSense.c
```diff
@@ -68,31 +80,31 @@ ExitCode Io_CurrentSense_ConvertToHighResolutionMainCurrent(float adc_voltage, f
     //
     // Sensitivity = 6.67mV/A
 
-    const float offset             = 2.5f;
-    const float sensitivity        = 6.67e-3f;
-    const float hsnbv_d06_output_2 = adc_voltage * (1e3f + 2.2e3f) / 2.2e3f;
+    const float hsnbv_d06_output_2 = adc_voltage * 1 / OUTPUT_2_DIV;
```
pasitlaothamatas, line 83:
```suggestion
    const float hsnbv_d06_output_2 = adc_voltage / OUTPUT_2_DIV;
```

pasitlaothamatas, line 83: could be nice to clarify with a small comment that this is the voltage before the divider. Up 2 u tho im not super picky

pasitlaothamatas:
Need to make sure that the App_TractiveSystem in BMS' `TestStateMachine` is updated (similar to what you have in main.c). After that i think we should be good to go

boards/BMS/Src/Io/Io_CurrentSense.c
```diff
@@ -1,13 +1,30 @@
...
+#define OUTPUT_1_SENSITIVITY (40e-3f)
+// Voltage divider between sensor output and adc input.
+#define OUTPUT_1_DIV (2.155f / (1.1f + 2.155f))
+
+// Offset voltage of output 2. Found to be 2.45V through testing
+#define OUTPUT_2_OFFSET (2.45f)
+// Sensitivity of output 2: 6.67mV/A
+#define OUTPUT_2_SENSITIVITY (6.67e-3f)
+// Voltage divider between sensor output and adc input.
+#define OUTPUT_2_DIV (2.155f / (1.1f + 2.155f))
```
pasitlaothamatas, line 17: Continuing off yesterdays discussion, its better to invert the values for `SENSITIVITY` and `DIV` to save some CPU cycles by using multiplication instead of division

Assembly for reference:
image

Look up VMUL and VDIV in the link below:
https://developer.arm.com/documentation/ddi0439/b/BEHJADED

boards/BMS/Inc/Io/Io_CurrentSense.h
Eddy13013: updated

boards/BMS/Src/Io/Io_CurrentSense.c
Eddy13013: I changed the current divider. But with the sensitivity I would need to divide in the #define anyways. Is this quicker tho? I guess it only runs on build and doesn't need to divide each time we call the function? Does that sound right?

pasitlaothamatas: Should be done in compile time i belive so i think you're good

boards/BMS/Src/Io/Io_CurrentSense.c
```diff
@@ -58,41 +70,40 @@ ExitCode Io_CurrentSense_ConvertToHighResolutionMainCurrent(float adc_voltage, f
...
-    const float sensitivity        = 6.67e-3f;
-    const float hsnbv_d06_output_2 = adc_voltage * (1e3f + 2.2e3f) / 2.2e3f;
+    // Output from current sensor:
+    const float hsnbv_d06_output_2 = adc_voltage * OUTPUT_2_DIV;
 
-    *high_res_main_current = (hsnbv_d06_output_2 - offset) / sensitivity;
-
-    return EXIT_CODE_OK;
+    // Return the current which corresponds to the output voltage
+    return (hsnbv_d06_output_2 - OUTPUT_2_OFFSET) / OUTPUT_2_SENSITIVITY;
```
pasitlaothamatas, line 87: may be a mistake here with the sensitivity (assuming the #define has been flipped)

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -382,6 +382,10 @@ SG_ PRECHARGE_RELAY: 32|1@1+ (1,0) [0|1] "" DEBUG
 BO_ 118 BMS_PACK_VOLTAGE: 4 BMS
 SG_ PACK_VOLTAGE : 0|32@1+ (1,0) [0|600] "" DEBUG
 
+BO_ 119 BMS_CURRENT_SENSE: 8 BMS
+SG_ HIGH_RES_CURRENT : 0|32@1+ (1,0) [-50|50] "" DEBUG
+SG_ LOW_RES_CURRENT : 32|32@1+ (1,0) [-300|300] "" DEBUG
```
pasitlaothamatas, line 387:
```suggestion
SG_ HIGH_RES_CURRENT : 0|32@1+ (1,0) [-50|50] "A" DEBUG
SG_ LOW_RES_CURRENT : 32|32@1+ (1,0) [-300|300] "A" DEBUG
```

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -627,4 +637,4 @@ TEST_F(BmsStateMachineTest, check_state_transition_from_fault_to_init_with_air_n
     ASSERT_EQ(CANMSGS_BMS_STATE_MACHINE_STATE_INIT_CHOICE, App_CanTx_GetPeriodicSignal_STATE(can_tx_interface));
 }
 
-} // namespace StateMachineTest
+} // namespace StateMachineTest
```
pasitlaothamatas, line 640:
```suggestion
} // namespace StateMachineTest

```

pasitlaothamatas (approved):
approved pending changes requested

PR #796, opened 2022-06-04
Advertise derated power from the BMS

Eddy13013:
### Checklist


pasitlaothamatas (approved):

pasitlaothamatas:
need to fix conflicts and we should be good here

PR #799, opened 2022-06-04
Add torque derating on DCM

timbitss:
### Summary
- Limit maximum torque based on power available from BMS
- "Powertrain" efficiency set to conservative value of 0.8 currently for all speed ranges (decrease if exceeding power limit and vice-versa)

### Checklist
- [ x] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).

### DIFF
--- a/boards/DCM/Src/App/states/App_DriveState.c
+++ b/boards/DCM/Src/App/states/App_DriveState.c
@@ -1,62 +1,30 @@
 #include <stdlib.h>
 #include <math.h>
+#include "App_SharedMacros.h"
 #include "states/App_AllStates.h"
 #include "states/App_InitState.h"
 #include "App_SetPeriodicCanSignals.h"
 
 #define MAX_TORQUE_REQUEST_NM (90.0f)
-
-#define MIN(a, b) ((a) < (b) ? (a) : (b))
+#define EFFICIENCY_ESTIMATE (0.80f)
 #define RPM_TO_RADS(rpm) ((rpm) * (float)M_PI / 30.0f)
-
-struct EfficiencyLUTEntry
-{
-    float speed_rpm;
-    float efficiency_estimate;
-};
-
-struct EfficiencyLUTEntry EfficiencyLUT[14] = {
-    [0]  = { .speed_rpm = 500.0f, .efficiency_estimate = 80.0f },
-    [1]  = { .speed_rpm = 1000.0f, .efficiency_estimate = 80.0f },
-    [2]  = { .speed_rpm = 1500.0f, .efficiency_estimate = 80.0f },
-    [3]  = { .speed_rpm = 2000.0f, .efficiency_estimate = 80.0f },
-    [4]  = { .speed_rpm = 2500.0f, .efficiency_estimate = 80.0f },
-    [5]  = { .speed_rpm = 3000.0f, .efficiency_estimate = 80.0f },
-    [6]  = { .speed_rpm = 3500.0f, .efficiency_estimate = 80.0f },
-    [7]  = { .speed_rpm = 4000.0f, .efficiency_estimate = 80.0f },
-    [8]  = { .speed_rpm = 4500.0f, .efficiency_estimate = 80.0f },
-    [9]  = { .speed_rpm = 5000.0f, .efficiency_estimate = 80.0f },
-    [10] = { .speed_rpm = 5500.0f, .efficiency_estimate = 80.0f },
-    [11] = { .speed_rpm = 6000.0f, .efficiency_estimate = 80.0f },
-    [12] = { .speed_rpm = 6500.0f, .efficiency_estimate = 80.0f },
-    [13] = { .speed_rpm = 7000.0f, .efficiency_estimate = 80.0f } // Maximum motor speed = 7000 rpm
-};
 
 void App_SetPeriodicCanSignals_TorqueRequests(struct DcmCanTxInterface *can_tx, struct DcmCanRxInterface *can_rx)
 {
-    uint8_t lut_idx;
-    float   efficiency_estimate;
-    float   max_torque_request, torque_request;
-    float   bms_available_power = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx);
+    float max_torque_request, torque_request;
+    float bms_available_power   = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx);
     float right_motor_speed_rpm = (float)abs(App_CanRx_INVR_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVR(can_rx));
-    float left_motor_speed_rpm  = (float)App_CanRx_INVL_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVL(can_rx);
-
-    // Get the estimated efficiency for the current motor speed
-    for (lut_idx = 0U; right_motor_speed_rpm < EfficiencyLUT[lut_idx].speed_rpm; lut_idx++)
-        ;
-    efficiency_estimate = EfficiencyLUT[lut_idx].efficiency_estimate;
+    float left_motor_speed_rpm  = (float)abs(App_CanRx_INVL_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVL(can_rx));
 
     // Estimate the maximum torque request to draw the maximum power available from the BMS
     // Note that the motors can not exceed a torque of MAX_TORQUE_REQUEST_NM
-    max_torque_request =
-        MIN(bms_available_power * efficiency_estimate /
-                (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm)),
-            MAX_TORQUE_REQUEST_NM);
+    float bms_torque_limit = bms_available_power * EFFICIENCY_ESTIMATE /
+                             (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm));
+    max_torque_request = min(bms_torque_limit, MAX_TORQUE_REQUEST_NM);
 
     // Calculate the actual torque request to transmit
     torque_request =
-        MIN(0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * MAX_TORQUE_REQUEST_NM,
-            max_torque_request);
+        0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * max_torque_request;
 
     // Transmit torque command to both inverters
     App_CanTx_SetPeriodicSignal_TORQUE_COMMAND_INVL(
--- a/boards/DCM/Src/App/states/App_DriveState.c
+++ b/boards/DCM/Src/App/states/App_DriveState.c
@@ -1,20 +1,62 @@
+#include <stdlib.h>
+#include <math.h>
 #include "states/App_AllStates.h"
 #include "states/App_InitState.h"
 #include "App_SetPeriodicCanSignals.h"
 
-// TODO: Have separate maximum torque requests for motoring and generating
 #define MAX_TORQUE_REQUEST_NM (90.0f)
+
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
+#define RPM_TO_RADS(rpm) ((rpm) * (float)M_PI / 30.0f)
+
+struct EfficiencyLUTEntry
+{
+    float speed_rpm;
+    float efficiency_estimate;
+};
+
+struct EfficiencyLUTEntry EfficiencyLUT[14] = {
+    [0]  = { .speed_rpm = 500.0f, .efficiency_estimate = 80.0f },
+    [1]  = { .speed_rpm = 1000.0f, .efficiency_estimate = 80.0f },
+    [2]  = { .speed_rpm = 1500.0f, .efficiency_estimate = 80.0f },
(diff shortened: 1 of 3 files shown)


boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
+#define RPM_TO_RADS(rpm) ((rpm) * (float)M_PI / 30.0f)
+
+struct EfficiencyLUTEntry
 {
-    float torque_request = 0.0f;
+    float speed_rpm;
+    float efficiency_estimate;
+};
+
+struct EfficiencyLUTEntry EfficiencyLUT[14] = {
```
Auhv, line 18: the .speed_rpm part of the struct seems redundant here. You could do the same thing as a simple 1-D array instead of a struct.

uint8_t eff_lut_index = speed_rpm/RPM_INTERVAL; //RPM_INTERVAL is 500 in the LUT you proposed
float efficiency_estimate = EfficiencyLUT[eff_lut_index ];

This way you're not iterating through a for loop to find the right index, and you don't need to waste memory on the struct. Will become important when you're dealing with a giant 2-D LUT with interpolation

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
-    //                        100%
-    //
+void App_SetPeriodicCanSignals_TorqueRequests(struct DcmCanTxInterface *can_tx, struct DcmCanRxInterface *can_rx)
+{
+    uint8_t lut_idx;
+    float   efficiency_estimate;
+    float   max_torque_request, torque_request;
+    float   bms_available_power = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx);
+    float right_motor_speed_rpm = (float)abs(App_CanRx_INVR_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVR(can_rx));
+    float left_motor_speed_rpm  = (float)App_CanRx_INVL_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVL(can_rx);
```
Auhv, line 42: Missing abs function on this value

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
+
+    // Get the estimated efficiency for the current motor speed
+    for (lut_idx = 0U; right_motor_speed_rpm < EfficiencyLUT[lut_idx].speed_rpm; lut_idx++)
+        ;
+    efficiency_estimate = EfficiencyLUT[lut_idx].efficiency_estimate;
+
+    // Estimate the maximum torque request to draw the maximum power available from the BMS
+    // Note that the motors can not exceed a torque of MAX_TORQUE_REQUEST_NM
+    max_torque_request =
+        MIN(bms_available_power * efficiency_estimate /
```
Eddy13013, line 52: maybe do different efficiency estimates per motor. because efficiency depends on speed and because the two motor speeds may be different, there could be

Auhv: Agree

Auhv, line 52: FSM max advertised torque should be in this MIN expression

Auhv (changes requested):

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas, line 18: the efficiency estimate is not a number bewtween [0,1]. the torque requested will be bigger tahn expected

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
+                (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm)),
+            MAX_TORQUE_REQUEST_NM);
+
+    // Calculate the actual torque request to transmit
     torque_request =
-        0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * MAX_TORQUE_REQUEST_NM;
+        MIN(0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * MAX_TORQUE_REQUEST_NM,
+            max_torque_request);
```
pasitlaothamatas, line 59: why compute the min here? shouldnt your derated max torque request already be computed in the equation above?

pasitlaothamatas: should just be 
`0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * max_torque_request`

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
+    // Get the estimated efficiency for the current motor speed
+    for (lut_idx = 0U; right_motor_speed_rpm < EfficiencyLUT[lut_idx].speed_rpm; lut_idx++)
+        ;
+    efficiency_estimate = EfficiencyLUT[lut_idx].efficiency_estimate;
+
+    // Estimate the maximum torque request to draw the maximum power available from the BMS
+    // Note that the motors can not exceed a torque of MAX_TORQUE_REQUEST_NM
+    max_torque_request =
+        MIN(bms_available_power * efficiency_estimate /
+                (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm)),
```
pasitlaothamatas, line 53: dont we want to average the motor speed? we are adding them right now

pasitlaothamatas: actually that might be dumb

boards/DCM/Src/App/states/App_DriveState.c
Auhv: Yeah I think i agree - couldn't all these things be one giant MIN statement?

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: dividing by the sum of the motor speeds seems incorrect though

Auhv: Adding them is correct, it will result in a torque that doesn't need to be divided by two

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,62 @@
...
+#include <math.h>
 #include "states/App_AllStates.h"
 #include "states/App_InitState.h"
 #include "App_SetPeriodicCanSignals.h"
 
-// TODO: Have separate maximum torque requests for motoring and generating
 #define MAX_TORQUE_REQUEST_NM (90.0f)
 
-void App_SetPeriodicCanSignals_TorqueRequests(struct DcmCanTxInterface *can_tx, struct DcmCanRxInterface *can_rx)
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
```
pasitlaothamatas, line 9: I beleive this macro is alrd implemented in App_SharedMacros.h

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas, line 18: I think it would also be nice to place the lookup table in a header file

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: the `bms_available_power` is the combined power of the left motor and the right motor correct? how do we optimally allocate this power to either motor right now? @Auhv

Auhv: Same torque for both @pasitlaothamatas

boards/DCM/Src/App/states/App_DriveState.c
Auhv: This way we're not arbitrarily choosing the RHS motor to calculate efficiency.

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: ok adding would make sense then sounds good

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: in the case that we have time to test the efficiency a bit better we can quickly update their values in the LUT so i dont mind keeping this for now in a header file. In our implementation in the state machine we can do what you suggested

boards/DCM/Src/App/states/App_DriveState.c
timbitss: Only right motor speed needs abs, but I could just do both for now

boards/DCM/Src/App/states/App_DriveState.c
timbitss: resolved

boards/DCM/Src/App/states/App_DriveState.c
timbitss: Resolved

boards/DCM/Src/App/states/App_DriveState.c
timbitss: Sorry is this the torque plausibility that we were talking about awhile ago. I did not have a chance to implement it

timbitss: For now, we can keep with a single efficiency for simplicity

boards/DCM/Src/App/states/App_DriveState.c
timbitss: resolved

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,38 @@
...
-    //
+    uint8_t lut_idx;
+    float   efficiency_estimate;
+    float   max_torque_request, torque_request;
+    float   bms_available_power = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx);
+    float right_motor_speed_rpm = (float)abs(App_CanRx_INVR_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVR(can_rx));
+    float left_motor_speed_rpm  = (float)abs(App_CanRx_INVL_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVL(can_rx));
+
+    // Get the estimated efficiency for the current motor speed
+    lut_idx =
```
pasitlaothamatas, line 22: for simplicity sake we should 
1) remove this variable
2) create a #define for the efficiency estimate and place it in this file
3) remove `EFFICIENCY_LUT_RPM_INTERVAL` in the header file (dont think its used anywhere after lut_index is removed)

we can keep the struct in the header file for now

Should be ready to merge after this

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,38 @@
...
+#include <math.h>
+#include "App_SharedMacros.h"
 #include "states/App_AllStates.h"
 #include "states/App_InitState.h"
 #include "App_SetPeriodicCanSignals.h"
+#include "configs/App_EfficiencyLUT.h"
 
-// TODO: Have separate maximum torque requests for motoring and generating
 #define MAX_TORQUE_REQUEST_NM (90.0f)
+#define RPM_TO_RADS(rpm) ((rpm) * (float)M_PI / 30.0f)
```
pasitlaothamatas, line 10: if some of these values are shared with the wheel speed sensor i suggest placing this macro in App_SharedMacros, and placing some #defines in App_SharedConstants

pasitlaothamatas: M_PI shouldn't require a cast btw. There should be a floating pt version for PI if this M_PI is a double

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -1,20 +1,38 @@
...
+    // Get the estimated efficiency for the current motor speed
+    lut_idx =
+        (uint8_t)min((uint8_t)right_motor_speed_rpm / EFFICIENCY_LUT_RPM_INTERVAL, EFFICIENCY_LUT_NUM_ENTRIES - 1);
+    efficiency_estimate = EfficiencyLUT[lut_idx].efficiency_estimate;
+
+    // Estimate the maximum torque request to draw the maximum power available from the BMS
+    // Note that the motors can not exceed a torque of MAX_TORQUE_REQUEST_NM
+    max_torque_request =
+        min(bms_available_power * efficiency_estimate /
+                (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm)),
```
pasitlaothamatas, line 30: Kinda nit but can we create a new float variable to hold this value and pass it later into the min macro? Will be a bit nicer to read

boards/DCM/Src/App/states/App_DriveState.c
timbitss: I don't see a floating point PI

boards/DCM/Src/App/states/App_DriveState.c
timbitss: left `EFFICIENCY_LUT_RPM_INTERVAL` incase we use it

boards/DCM/Src/App/states/App_DriveState.c
timbitss: resolved

pasitlaothamatas (approved):

boards/DCM/Src/App/states/App_DriveState.c
Auhv: Maybe now is a good time to finish that PR

PR #801, opened 2022-06-05
Various DIM updates

timbitss:
- Individual board RGB LED set to correct colour
- Set error table periodically within can rx thread
- Turn on red IMD and BSPD LEDs if IMD OK and BMS OK CAN signals from the BMS respectively are false
- AMS/BMS RGB LED only turns red if BMS_OK CAN signal is false, otherwise it is off

### Testing Done
None yet


boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -151,6 +161,11 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
 
     for (size_t i = 0; i < NUM_BOARDS; i++)
     {
+        if (i == BMS)
+        {
+            continue;
+        }
```
pasitlaothamatas, line 167: which spec is this implementing?

Eddy13013 (approved):
Approved, but we should verify this on the car first.

boards/DIM/Src/App/states/App_DriveState.c
timbitss: AMS OK indicator light

timbitss:
tested

PR #808, opened 2022-06-10
Reduce freq for debug msgs

pasitlaothamatas:
### Summary

The goal of this PR is to quickly mitigate phantom issues with CAN we were running into earlier. We think the cause of CAN issues are a result of the following:

(1) DIM peripheral not configured properly on cube MX
(2) Too much bus traffic (load reduced from 70% to 36%)
(3) High priority msgs getting pre empted due to CAN MID not being assigned with the appropriate priority

Solve by:
- Move pedal MID to higher pri (relative to torque req)
- Move the DIM switch MID to higher pri (don't want to miss start sequence)
- Remove unnecessary msgs
- Reduce msg frequency for certain CAN msgs
- Bump up msg pri for the start switch to prevent start sequence failure
- Update DIM CAN peripheral cube config
- Adding a delay before peripheral config (idk if this matters or not but we simply do not have the time to fully test this out). Each delay is not the same.
- Apply Io_SharedCan fix as Geoff had suggested
- Nuke modules that pollute the bus atm

DIM messages are no longer getting dropped. Start switch messages received reliably via the dongle

### Testing Done

Lower bus load
### Checklist


scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -199,6 +199,9 @@ SG_ D2_Write_Success : 16|1@1+ (1,0) [0|1] "" DEBUG
 SG_ D3_Data_Response : 32|16@1- (1,0) [-32768|32767] "" DEBUG
 SG_ D1_Parameter_Address_Response : 0|16@1+ (1,0) [0|65535] "" DEBUG
```
pasitlaothamatas, line 200: @Eddy13013  review this

boards/shared/Src/Io/Io_SharedCan.c
```diff
@@ -287,10 +287,12 @@ void Io_SharedCan_TransmitEnqueuedCanTxMessagesFromTask(void)
 {
     xSemaphoreTake(CanTxBinarySemaphore.handle, portMAX_DELAY);
 
-    while (HAL_CAN_GetTxMailboxesFreeLevel(sharedcan_hcan) > 0 && uxQueueMessagesWaiting(can_tx_msg_fifo.handle) > 0)
+    while (uxQueueMessagesWaiting(can_tx_msg_fifo.handle) > 0)
```
pasitlaothamatas, line 290: @Eddy13013 review this

boards/DIM/DIM.ioc
```diff
@@ -11,11 +11,15 @@ ADC2.OffsetNumber-1\#ChannelRegularConversion=ADC_OFFSET_NONE
 ADC2.Rank-1\#ChannelRegularConversion=1
 ADC2.SamplingTime-1\#ChannelRegularConversion=ADC_SAMPLETIME_61CYCLES_5
 ADC2.SubFamily=STM32F302xC
+CAN.ABOM=ENABLE
```
pasitlaothamatas, line 14: @Eddy13013  review this

Eddy13013 (approved):

PR #810, opened 2022-06-11
Implement EV.5.7 APPS / Brake Pedal Plausibility Check

pasitlaothamatas:
### Summary

- This is a rules req. that we had missed. We are allowed to send positive torque request only after the pedal percentage has returned to a position that is lower than 5%.
- Also setting the torque request to 0 when the brake is actuated. I had presumed this was done earlier but it hasn't. Let me know if you guys think this is a good idea

### Checklist

### DIFF
--- a/boards/DCM/Src/App/states/App_DriveState.c
+++ b/boards/DCM/Src/App/states/App_DriveState.c
@@ -9,30 +9,54 @@
 #define EFFICIENCY_ESTIMATE (0.80f)
 #define RPM_TO_RADS(rpm) ((rpm) * (float)M_PI / 30.0f)
 
-void App_SetPeriodicCanSignals_TorqueRequests(struct DcmCanTxInterface *can_tx, struct DcmCanRxInterface *can_rx)
-{
-    const float bms_available_power = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx);
-    const float right_motor_speed_rpm =
-        (float)abs(App_CanRx_INVR_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVR(can_rx));
-    const float left_motor_speed_rpm =
-        (float)abs(App_CanRx_INVL_MOTOR_POSITION_INFO_GetSignal_D2_MOTOR_SPEED_INVL(can_rx));
-    float bms_torque_limit = MAX_TORQUE_REQUEST_NM;
-    float fsm_torque_limit = App_CanRx_FSM_TORQUE_LIMITING_GetSignal_FSM_TORQUE_LIMIT(can_rx);
+// EV.5.7.1 The power to the Motor(s) must be immediately and completely shut down when both of the following exist at
+// the same time: • The mechanical brakes are actuated • The APPS signals more than 25% pedal travel EV.5.7.2 The Motor
+// shut down must remain active until the APPS signals less than 5% pedal travel, with or without brake operation.
+#define BRAKE_PLAUSIBILITY_LATCH_SET_THRESH (25.0f)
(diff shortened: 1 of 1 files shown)


boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -9,30 +9,56 @@
...
-        bms_torque_limit = bms_available_power * EFFICIENCY_ESTIMATE /
-                           (RPM_TO_RADS(right_motor_speed_rpm) + RPM_TO_RADS(left_motor_speed_rpm));
+        calculated_torque_limit = App_CanRx_BMS_AVAILABLE_POWER_GetSignal_AVAILABLE_POWER(can_rx) *
+                                  EFFICIENCY_ESTIMATE / (right_motor_speed_rad_s + left_motor_speed_rad_s);
     }
 
-    // Calculate the maximum torque request to scale pedal percentage off of
-    const float max_torque_request = min(bms_torque_limit, MAX_TORQUE_REQUEST_NM);
+    const float apps_pct = App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx);
+    if (App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx))
```
Eddy13013, line 39:
```suggestion
    if (App_CanRx_FSM_BRAKE_GetSignal_BRAKE_IS_ACTUATED(can_rx) && apps_pct >= 25.0f)
```

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -9,30 +9,56 @@
...
+    {
+        // Set the brake actuated latch high only when the brake is actuated
+        brake_actuated_latch = true;
+    }
+    else if (brake_actuated_latch)
```
Eddy13013, line 44:
```suggestion
    else if (brake_actuated_latch && apps_pct < CLEAR_BRAKE_ACTUATED_LATCH_THRESH)
```

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: check again

boards/DCM/Src/App/states/App_DriveState.c
pasitlaothamatas: check again

pasitlaothamatas:
after some discussion we decided on keeping the 25 pedal percent threshold

boards/DCM/Src/App/states/App_DriveState.c
```diff
@@ -9,30 +9,54 @@
...
 
-    // Calculate the actual torque request to transmit
-    const float torque_request =
-        min(0.01f * App_CanRx_FSM_PEDAL_POSITION_GetSignal_MAPPED_PEDAL_PERCENTAGE(can_rx) * max_torque_request,
-            fsm_torque_limit);
+    // Calculate the torque request to transmit to the inverter
+    float torque_request = 0.0f;
+    if (!brake_actuated_latch)
+    {
+        const float max_torque_request = min(calculated_torque_limit, MAX_TORQUE_REQUEST_NM);
```
Eddy13013, line 56: what happens if motor speeds are zero? calculated_torque_limit would be the same as MAX_TORQUE_REQUEST_NM. I want to confirm that this will not break our min function.

timbitss: No, it will not since the min function is simply a macro that we have defined in `App_SharedMacros.h`: 

 `#define min(a,b) ((a) < (b) ? (a) : (b))`

timbitss:
The FSM already sends 0% mapped pedal percentage when the brake is pressed, and @pasitlaothamatas has already implemented this rule back in August 2020

PR #812, opened 2022-06-12
Send air shdn to DCM from BMS

pasitlaothamatas:
### Summary

Think what @timbitss had earlier was correct. Gonna add it back!

### Checklist


scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -244,15 +244,15 @@ BO_ 108 BMS_STATE_OF_CHARGE: 4 BMS
...
-SG_ CHARGING_EXT_SHUTDOWN_OCCURRED : 8|1@1+ (1,0) [0|1] "" DEBUG
+SG_ CHARGER_DISCONNECTED_IN_CHARGE_STATE : 0|1@1+ (1,0) [0|1] "" DCM
+SG_ MIN_CELL_VOLTAGE_OUT_OF_RANGE : 1|1@1+ (1,0) [0|1] "" DCM
+SG_ MAX_CELL_VOLTAGE_OUT_OF_RANGE : 2|1@1+ (1,0) [0|1] "" DCM
+SG_ HAS_PEC_ERROR: 3|1@1+ (1,0) [0|1] "" DCM
+SG_ MIN_CELL_TEMP_OUT_OF_RANGE: 4|1@1+ (1,0) [0|1] "" DCM 
+SG_ MAX_CELL_TEMP_OUT_OF_RANGE: 5|1@1+ (1,0) [0|1] "" DCM
+SG_ CHARGER_FAULT_DETECTED : 6|1@1+ (1,0) [0|1] "" DCM
+SG_ HAS_REACHED_MAX_V : 7|1@1+ (1,0) [0|1] "" DCM 
+SG_ CHARGING_EXT_SHUTDOWN_OCCURRED : 8|1@1+ (1,0) [0|1] "" DCM
```
Eddy13013, line 255: why does the DCM care about these errors?

Eddy13013: shouldn't we be checking the critical error table to see if it's cleared instead?

timbitss: Yes, we are checking the critical error table to see if it's cleared in the init state as part of the "ready to drive" checks. If any critical error is set, the DCM enters the fault state. The DCM can only leave the fault state if all errors in the error table are set to 0.

@pasitlaothamatas I labelled DCM as the signal receiver as a formality based on that spec, but in the implementation of the error table, I believe it doesn't matter if it's set to `DCM` or `DEBUG`

pasitlaothamatas: The DCM won't be able to read the CAN msg to update the error table then if its not set up as a receiver @timbitss

pasitlaothamatas: @Eddy13013 these are air shdn faults so theyre classified as a critical fault

timbitss: @pasitlaothamatas How does the code in `Io_SharedErrorTable_SetErrorsFromCanMsg(struct ErrorTable *error_table, struct CanMsg *can_msg)` compile if error messages are being sent to `DEBUG` instead of `DCM`? 

I also recall the DCM being able to detect errors and properly transition to the fault state with the receiver set to `DEBUG` unless you've seen otherwise.

PR #836, opened 2022-07-07
Improved ADC driver

pasitlaothamatas:
### Summary

This PR aims to reduce the number of files and functions needed to read from an ADC. Previously, we are maintaining multiple files within each board to deal with signals read from the ADC. I'll describe the previous design in the section below:
- Io_Adc.c is used to read the raw value from each ADC channel
- Io_VoltageSense, or Io_Currentsense will call Io_Adc.h functions to further process the voltage measured at the pin to a "real" value
- Two functions provided `adc_driver_init()` and `adc_driver_getConvertedResult` should be all the user needs to get ADC readings

Looking forward to reviews. I hope to roll this out to other boards soon so we can continue removing unnecessary files in our codebase

### Checklist

### DIFF
--- a/boards/BMS/Inc/Io/adc_driver.h
+++ b/boards/BMS/Inc/Io/adc_driver.h
@@ -1,0 +1,15 @@
+#pragma once
+
+typedef enum
+{
+    // ADC1 Channels
+    TS_VOLTAGE = 0U,
+
+    // ADC2 Channels
+    HIGH_RES_TS_CURRENT_SENSE,
+    LOW_RES_TS_CURRENT_SENSE,
+    NUM_ADC_CHANNELS,
+} AdcChannels_E;
+
+void  adc_driver_init(void);
+float adc_driver_getConvertedResult(AdcChannels_E adc_channel);
--- a/boards/BMS/Src/Io/adc_driver.c
+++ b/boards/BMS/Src/Io/adc_driver.c
@@ -1,0 +1,178 @@
+#include <stm32f3xx.h>
+#include <stm32f3xx_hal_adc.h>
+#include "adc_driver.h"
+
+#define ADC_RESOLUTION (4095.0f)
+#define SINGLE_ENDED_ADC_V_SCALE (3.3f) // Single-ended ADC voltage has a range of 3.3V as we can measure [0, 3.3] V
+#define DIFFERENTIAL_ADC_V_SCALE (6.6f) // Differential ADC voltage has a range of 6.6V as we can measure [-3.3, 3.3] V
+
+// Macros used to convert digital ADC readings to an analog voltage
+#define SINGLE_END_RAW_ADC_TO_V (SINGLE_ENDED_ADC_V_SCALE / ADC_RESOLUTION)
+#define DIFFERENTIAL_RAW_ADC_TO_V (DIFFERENTIAL_ADC_V_SCALE / ADC_RESOLUTION)
+
+/***************************************
+ * ADC 1 SCALE AND OFFSET CONFIGURATIONS
+ ***************************************/
+// TS offset and scale cfgs
+#define TS_VOLTAGE_R_ERROR_COMPENSATION (1.05f)
+#define TS_VOLTAGE_AMPLIFIER_GAIN (8.0f)
+#define TS_VOLTAGE_DIV (1.24e+3f / (499e+3f + 499e+3f + 499e+3f + 499e+3f + 1.24e+3f))
+#define TS_VOLTAGE_SCALE \
+    (DIFFERENTIAL_RAW_ADC_TO_V * TS_VOLTAGE_R_ERROR_COMPENSATION / (TS_VOLTAGE_DIV * TS_VOLTAGE_AMPLIFIER_GAIN))
+#define TS_VOLTAGE_OFFSET_SCALED (0.0f)
+
+/***************************************
+ * ADC 2 SCALE AND OFFSET CONFIGURATIONS
+ ***************************************/
+// High res current sense offset and scale cfgs
+#define OUTPUT_1_SENSITIVITY (1.0f / 40e-3f)      // Sensitivity of output 1: 40mV/A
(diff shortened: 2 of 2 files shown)


boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+    }
+    else if (hadc->Instance == ADC2)
+    {
+        for (uint8_t channel = HIGH_RES_TS_CURRENT_SENSE; channel < NUM_ADC_CHANNELS; channel++)
+        {
+            const uint8_t adc_rank = adc_cfg[channel].rank;
+            adc2_voltages[adc_rank] =
+                GET_CONVERTED_ADC_VALUE(raw_adc2_buffer[channel], adc_cfg[channel].scale, adc_cfg[channel].offset);
+        }
+    }
```
pasitlaothamatas, line 148: I can try to make this more scalable but considering the no. of channels we are working with as well as the number of ADC instances we have on the MCU (2) this should suffice

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+}
+
+void adc_driver_init(void)
```
pasitlaothamatas, line 151: We should initialize the ADC->DMA->mem pipeline within the scope of this file to avoid exposing pointers unnecessarily

pasitlaothamatas: I can also pass in the ADC handles into this function. Pretty indifferent. We will likely need to stick with what we're doing right now if we use Modules (functions need to follow template of `void x_init(void)`)

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+        .offset = TS_VOLTAGE_OFFSET_SCALED,
+    },
+
+    // ADC instance 2 configuration
+    [HIGH_RES_TS_CURRENT_SENSE] =
+    {
+        .adc_instance = ADC_INSTANCE_2,
+        .rank = RANK_ADC2_CHANNEL_1,
+        .scale = HIGH_RES_TS_CURRENT_SCALE,
+        .offset = HIGH_RES_TS_CURRENT_OFFSET_SCALED,
```
pasitlaothamatas, line 118: Note that the value of the offsets prepared during compile time should be in the units of the "real" value, not voltage or a digital ADC value

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+    },
+};
+
+void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef *hadc)
+{
+    if (hadc->Instance == ADC1)
+    {
+        for (uint8_t channel = TS_VOLTAGE; channel < HIGH_RES_TS_CURRENT_SENSE; channel++)
+        {
+            const uint8_t adc_rank = adc_cfg[channel].rank;
```
pasitlaothamatas, line 135: This is not optimized right now. I should move this line outside of the if-else statements

boards/BMS/Inc/Io/adc_driver.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+typedef enum
```
pasitlaothamatas, line 3: Exposing this enum to the user. When a user wishes to read a converted ADC result, simply call `adc_driver_getConvertedResult` and provide any of these enums as an argument

pasitlaothamatas: Would be a good idea to provide a guard in the case that `NUM_ADC_CHANNELS` gets passed in

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+
+// Buffers to store raw ADC values
+static uint16_t raw_adc1_buffer[NUM_ADC1_CHANNELS] = { 0U };
+static uint16_t raw_adc2_buffer[NUM_ADC2_CHANNELS] = { 0U };
+
+// Buffers to store converted ADC values
+static float    adc1_voltages[NUM_ADC1_CHANNELS]   = { 0U };
+static float    adc2_voltages[NUM_ADC2_CHANNELS]   = { 0U };
+
+static AdcConfig_S adc_cfg[NUM_ADC_CHANNELS] =
```
pasitlaothamatas, line 101: should be a static const

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+    HAL_ADC_Start_DMA(&hadc1, (uint32_t *)raw_adc1_buffer, hadc1.Init.NbrOfConversion);
+    HAL_ADC_Start_DMA(&hadc2, (uint32_t *)raw_adc2_buffer, hadc2.Init.NbrOfConversion);
+    HAL_TIM_Base_Start(&htim3);
+}
+
+float adc_driver_getConvertedResult(AdcChannels_E adc_channel)
+{
+    float result = 0.0f;
+
+    switch (adc_cfg[adc_channel].adc_instance)
```
pasitlaothamatas, line 163: reminder to add guard for `NUM_ADC_CHANNEL` enum

boards/BMS/Src/Io/adc_driver.c
```diff
@@ -0,0 +1,178 @@
...
+extern ADC_HandleTypeDef hadc2;
+extern TIM_HandleTypeDef htim3;
+
+// Buffers to store raw ADC values
+static uint16_t raw_adc1_buffer[NUM_ADC1_CHANNELS] = { 0U };
+static uint16_t raw_adc2_buffer[NUM_ADC2_CHANNELS] = { 0U };
+
+// Buffers to store converted ADC values
+static float    adc1_voltages[NUM_ADC1_CHANNELS]   = { 0U };
+static float    adc2_voltages[NUM_ADC2_CHANNELS]   = { 0U };
```
pasitlaothamatas, line 99: these buffers should be renamed

PR #848, opened 2022-10-24
ELEC-43/ELEC-268 BMS Faults

will-chaba:
### Summary

Move BMS fault checks and sending of CAN fault messages to App code and out of All_States. To keep PR's shorter this PR includes fault checks for App_Accumulator, App_TractiveSystem and App_PreChargeRelay. Others will me moved in a subsequent PR. 

The goal is to phase out use of the SharedErrorTable, and the remaining BMS errors contained in the error table will be moved out in the next PR. Decided to split into multiple PRs to make each smaller.

### Changelist 

App_Accumulator:
- Move fault checks and CAN message sending out of All_States and into function contained in App_Accumulator.c.

App_TractiveSystem:
- Implement check for overcurrent condition, implemented in App_TractiveSystem.c. Also sends overcurrent fault status on CAN.

App_PreChargeRelay:
- Move fault check out of PreChargeState and into PreChargeRelay.c. Behaviour slightly changed to trigger a fault if precharge fails 3 times in a row, requiring a power cycle if triggered.

App_CanMsgs.dbc
- Added new fault signals to BMS_AIR_SHUTDOWN_ERRORS. This message is to be updated further in next PR to update message names to follow common convention, as well as rename message to BMS_AIR_SHUTDOWN_FAULTS to be consistent with our determined Fault definition.

### Testing Done

Unit tests for implemented functions can be found in Test_StateMachine.cpp.

### Checklist


boards/BMS/Src/App/App_PreChargeRelay.c
```diff
@@ -32,3 +33,32 @@ void App_PrechargeRelay_Open(const struct PrechargeRelay *const precharge_relay)
...
+    {
+        precharge_fault_count = 0;
+    }
+
+    if (precharge_fault_count == 3)
+    {
+        precharge_shutdown_fault = true;
+    }
+
+    App_CanTx_SetPeriodicSignal_PRECHARGE_FAULT(can_tx, precharge_shutdown_fault);
```
pasitlaothamatas, line 62: Are we supposed to be able to recover from this fault after entering the fault state or will a power cycle be required?

will-chaba: I think the plan is to require a power cycle, with the mentality that if pre-charge fails 3 times in a row, there is likely an issue that should be investigated before trying again

boards/shared/Src/App/App_SharedErrorTable.c
```diff
@@ -69,14 +69,16 @@ struct ErrorTable *App_SharedErrorTable_Create(void)
...
-    INIT_ERROR(BMS_AIR_SHUTDOWN_MAX_CELL_TEMP_OUT_OF_RANGE, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_CELL_OVERVOLTAGE_FAULT, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_CELL_UNDERVOLTAGE_FAULT, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_HAS_PEC_FAULT, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_CELL_UNDERTEMP_FAULT, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_CELL_OVERTEMP_FAULT, BMS, AIR_SHUTDOWN_ERROR);
     INIT_ERROR(BMS_AIR_SHUTDOWN_CHARGER_FAULT_DETECTED, BMS, AIR_SHUTDOWN_ERROR);
     INIT_ERROR(BMS_AIR_SHUTDOWN_HAS_REACHED_MAX_V, BMS, AIR_SHUTDOWN_ERROR);
     INIT_ERROR(BMS_AIR_SHUTDOWN_CHARGING_EXT_SHUTDOWN_OCCURRED, BMS, AIR_SHUTDOWN_ERROR);
+    INIT_ERROR(BMS_AIR_SHUTDOWN_TS_OVERCURRENT_FAULT, BMS, AIR_SHUTDOWN_ERROR);
```
will-chaba, line 80: Updated SharedErrorTable with new faults, as well as updated names of existing faults, however the plan is still to phase the table out, just not in this PR

will-chaba:
Closing due to CI issues caused by changing base branch

will-chaba:
Creating a new PR did not fix problem as hoped, re-opening original

.github/workflows/githubactions.yml
```diff
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - master
+      - development
   pull_request:
     branches:
       - master
+      - development
```
will-chaba, line 10: Added because github actions were not running with development as base branch

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -189,7 +158,9 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
     {
         acc_meas_settle_count++;
     }
-    else if (App_SharedErrorTable_HasAnyCriticalErrorSet(error_table))
+
+    // App_SharedErrorTable_HasAnyCriticalErrorSet left in for now, to be removed entirely in future
+    else if (acc_fault || ts_fault || App_SharedErrorTable_HasAnyCriticalErrorSet(error_table))
```
will-chaba, line 163: Error table check left in for now, will be removed once the rest of the BMS faults are moved out of the error table in a future PR. See [ELEC-279](https://ubcformulaelectric.atlassian.net/browse/ELEC-279).

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -628,4 +628,120 @@ TEST_F(BmsStateMachineTest, check_state_transition_from_fault_to_init_with_air_n
...
+    SetInitialState(App_GetChargeState());
+
+    // Max acceptable charge current is 23.6A * 3 = 70.8A
+    get_high_res_current_fake.return_val = 24.0f;
+    get_low_res_current_fake.return_val  = 24.0f;
+    LetTimePass(state_machine, 10);
+    ASSERT_EQ(App_GetFaultState(), App_SharedStateMachine_GetCurrentState(state_machine));
+}
+
+TEST_F(BmsStateMachineTest, check_precharge_fault)
```
will-chaba, line 724: This test does not strictly test the state machine, could be moved into a different file

.env
```diff
@@ -1 +1 @@
-PYTHONPATH=${PYTHONPATH}
+PYTHONPATH=${PYTHONPATH}:${PWD}
```
will-chaba, line 1: Had to undo a change made in #845 as it was causing my tests to fail, unsure of if there was a reason that this was changed and if I should look for an alternate solution

boards/BMS/Src/App/App_TractiveSystem.c
```diff
@@ -1,8 +1,13 @@
 #include <stdlib.h>
 #include <assert.h>
+#include <stdbool.h>
+#include "App_TractiveSystem.h"
 
 #define HIGH_RES_MAX_CURRENT_READING (50.0f)
 
+#define MAX_TS_CHARGE_CURRENT_AMPS (23.6f)
```
will-chaba, line 8: Please confirm that I interpreted these limits correctly based on [this](https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/12353537/BMS+Faults+Warnings#What-are-faults-%26-warnings%3F)

boards/BMS/Src/App/App_PreChargeRelay.c
```diff
@@ -32,3 +33,32 @@ void App_PrechargeRelay_Open(const struct PrechargeRelay *const precharge_relay)
...
+    else
+    {
+        precharge_fault_count = 0;
+    }
+
+    if (precharge_fault_count >= 3)
+    {
+        precharge_shutdown_fault = true;
+    }
+
```
gtaharaedmonds, line 61: Just curious, where does the 3 come from? 30ms is not much of a debounce time

gtaharaedmonds (commented):
This is great. Nice tests.

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -245,14 +245,16 @@ SG_ State_Of_Charge : 0|32@1- (1,0) [0|1E2] "%" DIM
 
 BO_ 109 BMS_AIR_SHUTDOWN_ERRORS: 2 BMS
```
Eddy13013, line 246: please rename to BMS Faults

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -647,14 +649,16 @@ VAL_ 105 Condition 0 "IMD_SHORT_CIRCUIT" 1 "IMD_NORMAL" 2 "IMD_UNDERVOLTAGE_DETE
 VAL_ 105 OK_HS 0 "NO FAULT" 1 "FAULT";
 VAL_ 107 State 0 "INIT" 1 "AIR_OPEN" 2 "PRE_CHARGE" 3 "CHARGE" 4 "DRIVE" 5 "FAULT";
 VAL_ 109 CHARGER_DISCONNECTED_IN_CHARGE_STATE  0 "FALSE" 1 "TRUE";
-VAL_ 109 MIN_CELL_VOLTAGE_OUT_OF_RANGE 0 "FALSE" 1 "TRUE";
-VAL_ 109 MAX_CELL_VOLTAGE_OUT_OF_RANGE 0 "FALSE" 1 "TRUE";
-VAL_ 109 HAS_PEC_ERROR 0 "FALSE" 1 "TRUE";
-VAL_ 109 MIN_CELL_TEMP_OUT_OF_RANGE 0 "FALSE" 1 "TRUE";
-VAL_ 109 MAX_CELL_TEMP_OUT_OF_RANGE 0 "FALSE" 1 "TRUE";
+VAL_ 109 CELL_UNDERVOLTAGE_FAULT 0 "FALSE" 1 "TRUE";
```
Eddy13013, line 652: can you change to "OK" and "FAULT"

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -628,4 +628,120 @@ TEST_F(BmsStateMachineTest, check_state_transition_from_fault_to_init_with_air_n
...
+
+    LetTimePass(state_machine, 10);
+    ASSERT_EQ(App_GetFaultState(), App_SharedStateMachine_GetCurrentState(state_machine));
+}
+
+TEST_F(BmsStateMachineTest, check_state_transition_to_fault_state_from_all_states_overvoltage)
+{
+    SetInitialState(App_GetInitState());
+
+    get_max_cell_voltage_fake.return_val = 4.3f;
```
Eddy13013, line 645: just wondering if it's bad practice to put a constant here from the .h file. In the case where we change cell chemistry this could prevent a lot of rework. Thoughts?

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -628,4 +628,120 @@ TEST_F(BmsStateMachineTest, check_state_transition_from_fault_to_init_with_air_n
...
+
+    ASSERT_EQ(App_GetFaultState(), App_SharedStateMachine_GetCurrentState(state_machine));
+}
+
+TEST_F(BmsStateMachineTest, check_state_transition_to_fault_state_from_all_states_overtemp_charge_state)
+{
+    SetInitialState(App_GetChargeState());
+
+    // In Charge state acceptible temp range is (0, 45)
+    get_max_temp_degc_fake.return_val = 46.0f;
```
Eddy13013, line 672: this passed without setting the Is_Charger_connected to true?

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -194,3 +202,56 @@ void App_Accumulator_RunOnTick100Hz(struct Accumulator *const accumulator)
...
+    bool communication_fault = App_Accumulator_HasCommunicationError(accumulator);
+
+    uint8_t throwaway_segment = 0U;
+    uint8_t throwaway_loc     = 0U;
+
+    float max_allowable_cell_temp = MAX_CELL_DISCHARGE_TEMP_DEGC;
+    float min_allowable_cell_temp = MIN_CELL_DISCHARGE_TEMP_DEGC;
+
+    // if we are charging, max cell temp is 45C not 60C
+    if (isChargeState)
```
Eddy13013, line 224: I think rather than checking if the charger is connected, we should check if TS current is negative (for charge current)

boards/BMS/Src/App/App_TractiveSystem.c
```diff
@@ -73,3 +78,24 @@ float App_TractiveSystem_GetPower(struct TractiveSystem *ts)
...
     return App_TractiveSystem_GetVoltage(ts) * App_TractiveSystem_GetLowResCurrent(ts);
 }
+
+bool App_TractveSystem_CheckFaults(struct BmsCanTxInterface *can_tx, struct TractiveSystem *ts, bool isChargeState)
+{
+    bool  ts_overcurrent_fault     = false;
+    float max_allowable_ts_current = MAX_TS_DISCHARGE_CURRENT_AMPS;
+
+    // if we are charging, max ts current is 70.8A not 265.5A
+    if (isChargeState)
```
Eddy13013, line 88: this doesn't need to be a conditional statement. Current can never be outside its bounds regardless of state. For context: charge current is negative, discharge current is positive. So you could write a simple statement using the charge & discharge bounds

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -159,16 +128,16 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
...
     App_SendAndReceiveHeartbeat(can_tx, can_rx, hb_monitor);
 
     App_Accumulator_RunOnTick100Hz(accumulator);
     App_CheckCellVoltageRange(can_tx, error_table, accumulator);
     App_CheckCellTemperatureRange(can_tx, error_table, accumulator, state_machine);
 
-    const bool has_acc_comms_error = App_Accumulator_HasCommunicationError(accumulator);
-    App_CanTx_SetPeriodicSignal_HAS_PEC_ERROR(can_tx, has_acc_comms_error);
-    App_SharedErrorTable_SetError(error_table, BMS_AIR_SHUTDOWN_HAS_PEC_ERROR, has_acc_comms_error);
+    const bool acc_fault = App_Accumulator_CheckFaults(can_tx, accumulator, isChargeState);
```
Eddy13013, line 139: ohhhh this looks so clean! Love it!

boards/BMS/Test/Src/Test_StateMachine.cpp
Eddy13013: I'm kinda wondering if some of these tests could be moved to a new fault test

boards/BMS/Src/App/App_TractiveSystem.c
Eddy13013: multiply by 3. we have 3 cells in parallel in the pack and current sensing is at the pack level

Eddy13013 (changes requested):

boards/BMS/Src/App/App_TractiveSystem.c
will-chaba: Gotcha. I'll update that

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -194,3 +202,56 @@ void App_Accumulator_RunOnTick100Hz(struct Accumulator *const accumulator)
...
+    {
+        max_allowable_cell_temp = MAX_CELL_CHARGE_TEMP_DEGC;
+        min_allowable_cell_temp = MIN_CELL_CHARGE_TEMP_DEGC;
+    }
+
+    if (App_Accumulator_GetMinCellTempDegC(accumulator, &throwaway_segment, &throwaway_loc) < min_allowable_cell_temp)
+    {
+        undertemp_fault = true;
+    }
+
```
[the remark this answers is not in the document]
will-chaba: > If you want to be more concise here you can replace these with:
> 
> > const bool undertemp_fault = App_Accumulator_GetMinCellTempDegC(accumulator, &throwaway_segment, &throwaway_loc) < min_allowable_cell_temp;
> 
> (this is fine though)

Neat trick! I didn't know you could assign bools like that, I'll give that a shot

boards/BMS/Test/Src/Test_StateMachine.cpp
will-chaba: is_charger_connected.fake.return_val is set to true in the setup function, but it turned out the state machine was changing to fault state for reasons other than those checked by the test. I've added some test-specific setup conditions such that fault state transitions should be triggered by the condition under test.

boards/BMS/Test/Src/Test_StateMachine.cpp
will-chaba: Personally I'm okay with it, I don't see why it would cause an issue

boards/BMS/Test/Src/Test_StateMachine.cpp
will-chaba: I definitely agree, especially for check_precharge_fault, however the other tests I wrote do test state machine behaviour. I think that given it's only one test maybe it should stay for the time being, but If I write more tests when moving the rest of the BMS faults out of the error table maybe I should separate it to its own test then. 

Thoughts? I just think it would be a little silly to have a dedicated test just for check_precharge_fault

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -62,7 +63,7 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
             (ts_voltage > threshold_voltage) && (elapsed_time <= PRECHARGE_COMPLETION_LOWER_BOUND);
         const bool is_charger_connected = App_Charger_IsConnected(charger);
         const bool has_precharge_fault =
-            (is_charger_connected) ? is_ts_rising_slowly : (is_ts_rising_slowly | is_ts_rising_quickly);
+            App_PrechargeRelay_CheckFaults(can_tx, is_charger_connected, is_ts_rising_slowly, is_ts_rising_quickly);
```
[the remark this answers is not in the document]
will-chaba: Should be updated now to address this, let me know if this implementation better follows the desired behaviour

boards/BMS/Src/App/App_PreChargeRelay.c
will-chaba: I misinterpreted the implementation, it has been reworked. See Edwin's comment on the matter for details

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -194,3 +200,40 @@ void App_Accumulator_RunOnTick100Hz(struct Accumulator *const accumulator)
...
+    struct TractiveSystem *const    ts)
+{
+    uint8_t throwaway_segment = 0U;
+    uint8_t throwaway_loc     = 0U;
+
+    float max_allowable_cell_temp = MAX_CELL_DISCHARGE_TEMP_DEGC;
+    float min_allowable_cell_temp = MIN_CELL_DISCHARGE_TEMP_DEGC;
+
+    // if we are charging, max cell temp is 45C not 60C
+    if (App_TractiveSystem_GetCurrent(ts) < 0.0f)
```
Eddy13013, line 216:
```suggestion
    if (App_TractiveSystem_GetCurrent(ts) < -3.0f)
```
Can you add a little margin for error here. 3A should be enough to account for any error and is still much lower than our max charge current

boards/BMS/Src/App/App_TractiveSystem.c
```diff
@@ -73,3 +74,15 @@ float App_TractiveSystem_GetPower(struct TractiveSystem *ts)
...
+{
+    //    Charge current is negative, discharge current is positive
+    //    TS current should be in the range: (-70.8,265.5)
+    bool ts_current_out_of_bounds =
+        !IS_IN_RANGE(MAX_TS_CHARGE_CURRENT_AMPS, MAX_TS_DISCHARGE_CURRENT_AMPS, App_TractiveSystem_GetCurrent(ts));
+
+    App_CanTx_SetPeriodicSignal_TS_OVERCURRENT_FAULT(can_tx, ts_current_out_of_bounds);
+
+    return ts_current_out_of_bounds;
+}
```
Eddy13013, line 88: add a newline at the end

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -243,16 +243,18 @@ SG_ State : 0|8@1+ (1,0) [0|255] "" DCM
...
+BO_ 109 BMS_FAULTS: 2 BMS
 SG_ CHARGER_DISCONNECTED_IN_CHARGE_STATE : 0|1@1+ (1,0) [0|1] "" DEBUG
-SG_ MIN_CELL_VOLTAGE_OUT_OF_RANGE : 1|1@1+ (1,0) [0|1] "" DEBUG
-SG_ MAX_CELL_VOLTAGE_OUT_OF_RANGE : 2|1@1+ (1,0) [0|1] "" DEBUG
-SG_ HAS_PEC_ERROR: 3|1@1+ (1,0) [0|1] "" DEBUG
-SG_ MIN_CELL_TEMP_OUT_OF_RANGE: 4|1@1+ (1,0) [0|1] "" DEBUG
-SG_ MAX_CELL_TEMP_OUT_OF_RANGE: 5|1@1+ (1,0) [0|1] "" DEBUG
+SG_ CELL_UNDERVOLTAGE_FAULT : 1|1@1+ (1,0) [0|1] "" DEBUG
+SG_ CELL_OVERVOLTAGE_FAULT : 2|1@1+ (1,0) [0|1] "" DEBUG
+SG_ PEC_ERROR: 3|1@1+ (1,0) [0|1] "" DEBUG
```
Eddy13013, line 250:
```suggestion
SG_ MODULE_COMM_ERROR: 3|1@1+ (1,0) [0|1] "" DEBUG
```

Eddy13013 (changes requested):

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -1,7 +1,13 @@
...
-#include <assert.h>
 #include "App_Accumulator.h"
 
+// Min and Max cell temperatures depending on state
+#define MAX_CELL_DISCHARGE_TEMP_DEGC (60.0f)
+#define MAX_CELL_CHARGE_TEMP_DEGC (45.0f)
+#define MIN_CELL_DISCHARGE_TEMP_DEGC (-20.0f)
+#define MIN_CELL_CHARGE_TEMP_DEGC (0.0f)
+#define MAX_CELL_VOLTAGE_CHARGE (4.2f)
+#define MIN_CELL_VOLTAGE_DISCHARGE (3.0f)
```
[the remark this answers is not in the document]
will-chaba: Oh good catch

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -61,12 +59,15 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
...
+        const bool has_precharge_fault  = App_PrechargeRelay_CheckFaults(
+            can_tx, is_charger_connected, is_ts_rising_slowly, is_ts_rising_quickly, &precharge_fault_count);
 
         if (has_precharge_fault)
         {
-            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+            if (precharge_fault_count >= 3)
+                App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+            else
+                App_SharedStateMachine_SetNextState(state_machine, App_GetInitState());
```
gtaharaedmonds, line 70: Instead of checking if the fault count >= 3 here too, why doesn't App_PrechargeRelay_CheckFaults just return whether or not there is a precharge shutdown fault and then all the checking is hidden?

gtaharaedmonds (changes requested):

boards/BMS/Src/App/states/App_PreChargeState.c
gtaharaedmonds: Also I don't think this will work: If there's a precharge fault, before there's been enough time to count to 3, we'll first go to init state. So fault state here is unreachable via this logic.
Also you're resetting the fault count to zero at line 49, so it can't get to 3 ever anyway.

will-chaba: > Also I don't think this will work: If there's a precharge fault, before there's been enough time to count to 3, we'll first go to init state. So fault state here is unreachable via this logic. Also you're resetting the fault count to zero at line 49, so it can't get to 3 ever anyway.

Ah good point. Setting the fault count as a static int should resolve it being reset, but I'm not sure I quite understand your first point. If the fault count does not reset, going back to Init state should not cause any issues should it? From @Eddy13013's comment [here](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/848#discussion_r1013585046) it sounds like that is the intended behavior but if you have something else in mind I'm all ears

will-chaba: > Instead of checking if the fault count >= 3 here too, why doesn't App_PrechargeRelay_CheckFaults just return whether or not there is a precharge shutdown fault and then all the checking is hidden?

I agree it's a bit cludgy, but unless we want to change to init state from inside App_PrechargeRelay_CheckFaults, we need some way to communicate to PrechargeState the difference between a "fault" and a "shutdown fault". I suppose rather than passing a pointer to precharge_fault_count I could just pass a pointer to a bool, which would probably be more along the lines of what you're suggesting?

gtaharaedmonds: Yeah ignore the first point I thought this was still a debounce thing
Would it be a problem then setting the fault count to be zero if a fault is not set, in CheckFaults? Since if there's a fault, we got Init, then back to Precharge, and initially there won't be a fault so the counter will be reset. Probably you only want to zero the fault counter on a successful precharge, i.e. transition to drive state?
I'd also suggest moving all the fault counting to the state machine, then you don't need to pass the pointer, but that's up to you

will-chaba: I agree with setting the counter to zero on a transition to drive state, as for implmenting the counter in CheckFaults vs Statemachine, I think that part of the goal of this transition is to make the state machine code more readable, i.e. more like pseudocode, and "hiding" the counter inside checkfaults might make the fault transitions easier to read. I'll look into the idea though, see what is the most intuitive to read, as that is the goal.

will-chaba: @gtaharaedmonds Okay I made some major changes after consulting @Eddy13013, let me know how you feel about the new implementation

boards/BMS/Src/App/App_PreChargeRelay.c
```diff
@@ -32,3 +32,48 @@ void App_PrechargeRelay_Open(const struct PrechargeRelay *const precharge_relay)
...
+    if (App_PrechargeRelay_GetFaultCounterVal(precharge_relay) >= 3)
+    {
+        *precharge_limit_exceeded = true;
+    }
+    else
+    {
+        *precharge_limit_exceeded = false;
+    }
```
gtaharaedmonds, line 74: `*precharge_limit_exceeded = App_PrechargeRelay_GetFaultCounterVal(precharge_relay) >= 3`

gtaharaedmonds: Also nit: You can #define the 3 to `MAX_PRECHARGE_ATTEMPTS` or something like that

will-chaba: Oh good call I like that

gtaharaedmonds (approved):

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -2,6 +2,14 @@
 #include <assert.h>
 #include "App_Accumulator.h"
 
+// Min and Max cell temperatures depending on state
```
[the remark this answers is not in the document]
Eddy13013: I was more thinking about moving the MAX_CELL_CHARGE_TEMP etc. to the header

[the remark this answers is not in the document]
Eddy13013: but you know what, this is good too

Eddy13013 (approved):

boards/BMS/Src/App/App_Accumulator.c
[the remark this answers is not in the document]
will-chaba: > I was more thinking about moving the MAX_CELL_CHARGE_TEMP etc. to the header

OH *facepalm* I don't know how I confused that. I can move them over

PR #849, opened 2022-10-26
Documentation Updates

Lucien950:
### Summary
README.md changes for Windows Build Environment.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,11 @@ A consolidated repository for gathering all firmware under one roof.
 - [FAQ](#faq)
 
 ## Project Setup
-### Environment Dependencies
+###  Dependencies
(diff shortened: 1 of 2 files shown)


will-chaba:
Trying to run through this myself, ran into a couple of issues thus far:
- Needs note of installing OpenOCD
- Specific Java version should be mentioned (First version of Java 11 I grabbed was incompatible)
- Specific STM32Cube version should be mentioned for simplicity (should be same as on CI)

I'm still not done, will update with more. These may seem like 'dumb' statements, but remember we'd like these instructions to be for the lowest common denominator (first/second year students)

Lucien950:
Perhaps we let this die

gtaharaedmonds:
I believe this is all deprecated with new build system so going to close

PR #851, opened 2022-11-03
Linux building fix

gtaharaedmonds:


boards/FSM/Src/App/states/App_AllStates.c
```diff
@@ -35,9 +35,9 @@ void App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
     }
 
     // Check for torque plausibility
-    float left_torque_req = App_CanMsgs_dcm_invl_command_message_torque_command_invl_decode(
+    float left_torque_req = (float)App_CanMsgs_dcm_invl_command_message_torque_command_invl_decode(
```
Eddy13013, line 38: not relevant to PR scope

Eddy13013 (changes requested):

boards/FSM/Src/App/states/App_AllStates.c
gtaharaedmonds: Oh yeah, can we group this in? Builds fail without this change for me
(I suspect its because we use different compilers for Linux vs. Windows)

will-chaba (approved):

Eddy13013 (commented):
Has this been tested on a linux machine?

gtaharaedmonds:
> Has this been tested on a linux machine?

The Github CI is Linux so yes. Also its just casting double to float, shouldn't be an issue anywhere.

Eddy13013 (approved):

PR #854, opened 2022-11-18
ELEC-279 Remove BMS Faults and Warnings from Error Table

will-chaba:
### Summary

Remove remaining faults and warnings (non-critical-errors) from the SharedErrorTable. This includes removing references to the error table from all BMS App code. 

### Changelist 

Includes changes to any file referencing the SharedErrorTable inside of the BMS code, as well as any file outside the BMS that references BMS errors inside the error table.

BMS motor shutdown errors also removed outright, as they seem to have been a placeholder that was never used. I'd recommend they be removed from other boards in the future unless plans for their use exist.

DIM was updated to check for BMS status over CAN instead of via the shared error table

Also includes changes to AppCanMsgs.dbc and any other dependant file for renaming on certain CAN messages.

### Testing Done

Existing tests were run and modified where needed to remove reliance on error table and to ensure existing functionality is unaffected.

### Resolved Issues

ELEC-273

### Checklist
- [ x] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).


boards/shared/Inc/App/App_BoardEnum.h
```diff
@@ -2,11 +2,10 @@
 
 enum Board
 {
-    BMS,
+    GSM,
```
will-chaba, line 5: GSM moved to enum value 0 such that the boards have the same order and values as Boardleds enum inside App_DriveState.c. GSM is not checked in DIM drive state, so it taking up position 0 where BMS used to be fits perfectly.

Snippet from DriveState:
```
typedef enum
{
    BMS_LED,
    DCM_LED,
    DIM_LED,
    FSM_LED,
    PDM_LED,
    NUM_BOARD_LEDS,
} BoardLeds;
```

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -190,3 +199,27 @@ const struct State *App_GetDriveState(void)
...
+}
+
+bool has_bms_warning(const struct DimCanRxInterface *can_rx)
+{
+    return App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1_HZ(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1_KHZ(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANRX(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANTX(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_WATCHDOG_TIMEOUT(can_rx);
+}
```
gtaharaedmonds, line 225: Hmm I thought the decision was to stop checking individual errors like is done here, previously done by the error table. And instead check if the BMS is in its "FAULT" state.

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -157,8 +152,7 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
         acc_meas_settle_count++;
     }
 
-    // App_SharedErrorTable_HasAnyCriticalErrorSet left in for now, to be removed entirely in future
-    else if (acc_fault || ts_fault || App_SharedErrorTable_HasAnyCriticalErrorSet(error_table))
+    else if (acc_fault || ts_fault)
     {
         status = false;
         App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
```
gtaharaedmonds, line 158: Shouldn't we also check that other boards are not in their fault state?

gtaharaedmonds: I don't want to accidently regress that we're no longer faulting if some error occurs on another board, could be dangerous so lets be careful here

boards/DIM/Inc/App/states/App_DriveState.h
```diff
@@ -1,9 +1,26 @@
...
  */
 const struct State *App_GetDriveState(void);
+
+/**
+ * Check CAN bus for BMS Fault messages
+ * @param can_rx Dim CAN RX interface for recieving messages
+ * @return True if any BMS Fault is present
+ */
+bool has_bms_fault(const struct DimCanRxInterface *can_rx);
+
```
gtaharaedmonds, line 20: Let's stay consistent with naming conventions for now
App_DriveState_HasBmsFault

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -104,19 +101,14 @@ static void DriveStateRunOnTick100Hz(struct StateMachine *const state_machine)
         can_tx, torque_vectoring_switch, App_CanTx_SetPeriodicSignal_TORQUE_VECTORING_SWITCH,
         CANMSGS_DIM_SWITCHES_START_SWITCH_ON_CHOICE, CANMSGS_DIM_SWITCHES_START_SWITCH_OFF_CHOICE);
 
-    struct RgbLed *board_status_leds[5] = {
```
gtaharaedmonds, line 107: nit: Change the 5 to NUM_BOARD_LEDs

boards/shared/Inc/App/App_BoardEnum.h
gtaharaedmonds: Hmm I don't see why this change is necessary. Why do the Board and BoardLed enums need the same order?

gtaharaedmonds (changes requested):

boards/DIM/Src/App/states/App_DriveState.c
will-chaba: That’s a solid point. I don’t know why that slipped my mind I’ll switch to that.

boards/shared/Inc/App/App_BoardEnum.h
will-chaba: It comes down to how the DIM checks for faults in each board. Because runs a for loop increment through the enum values listed above, for each value of i DIM calls App_SharedError_IsBoardInList, which uses the Board enum with the same i value. If they're not the same order it does not run correctly. The alternative fix would be to put BMS_LED at the bottom of BoardLeds, but the effect is the same.

boards/DIM/Inc/App/states/App_DriveState.h
will-chaba: Good call

boards/BMS/Src/App/states/App_AllStates.c
will-chaba: I think the idea is that we want boards (or at least the BMS) to act independently of other boards. @Eddy13013, @joethrstn and I discussed whether there were any boards that could have issues with high voltage being present while in their respective fault states, and couldn't come up with anything. Of course I'll let others chime in if I got any of that wrong.

boards/DIM/Src/App/states/App_DriveState.c
will-chaba: Can you think of any way to implement a cleaner check for warnings however? warnings are not currently state-dependent.

boards/BMS/Src/App/states/App_AllStates.c
gtaharaedmonds: Makes sense. Just wanted to double check there aren't currently any error table errors that we currently rely on to open contactors.

boards/shared/Inc/App/App_BoardEnum.h
gtaharaedmonds: I think we should maintain the same order in both enums, that seems more clear. Also we should not delete BMS from this enum, since it is still a board in the car. That will preserve the same order for all relevant boards with LEDs, and since we only iterate up to NUM_BOARD_LEDs, GSM will be ignored.

Tbh this is not  ideal - we really should not have to worry about enum ordering for this to work. Maybe we refactor this, but that's outside the scope of this PR. For now, can you add a unit tests that just checks that both enums have the same ordering, i.e. check DCM == DCM_LED, etc., so if we accidentally change the ordering we'll know.

boards/shared/Test/Src/Test_ErrorTable.cpp
```diff
@@ -447,6 +439,10 @@ TEST_F(SharedErrorTableTest, get_all_errors_using_every_error)
     ASSERT_EQ(NUM_ERROR_IDS, error_list.num_errors);
     for (size_t id = 0; id < NUM_ERROR_IDS; id++)
     {
+        if (!App_SharedError_IsErrorInList(&error_list, (enum ErrorId)id))
+        {
+            printf("fail");
```
gtaharaedmonds, line 444: Remove

gtaharaedmonds (changes requested):

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -190,3 +199,18 @@ const struct State *App_GetDriveState(void)
 
     return &drive_state;
 }
+
+bool App_DriveState_HasBmsFault(const struct DimCanRxInterface *can_rx)
+{
+    uint8_t test = App_CanRx_BMS_STATE_MACHINE_GetSignal_STATE(can_rx) == CANMSGS_BMS_STATE_MACHINE_STATE_FAULT_CHOICE;
+    return test;
```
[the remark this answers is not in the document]
will-chaba: Oops! `test` was actually only present for debugging, I must have forgotten to remove it! I'll return the comparison directly.

boards/DIM/Src/App/states/App_DriveState.c
```diff
@@ -190,3 +199,18 @@ const struct State *App_GetDriveState(void)
...
+}
+
+bool App_DriveState_HasBmsWarning(const struct DimCanRxInterface *can_rx)
+{
+    return App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1_HZ(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASK1_KHZ(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANRX(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_STACK_WATERMARK_ABOVE_THRESHOLD_TASKCANTX(can_rx) ||
+           App_CanRx_BMS_NON_CRITICAL_ERRORS_GetSignal_WATCHDOG_TIMEOUT(can_rx);
+}
```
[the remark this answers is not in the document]
will-chaba: I agree it's not great, especially once other boards use a similar format. What's not demonstrated here however is that this would allow us to pick and choose which warnings trigger the LED on the DIM, such that warnings irrelevant to the driver can be ignored.

boards/shared/Inc/App/App_BoardEnum.h
will-chaba: I agree BMS should not be removed from this enum. The issue is that leaving BMS in the Board enum increases NUM_BOARDS by 1. NUM_BOARDS is used in many places to indicate to the shared error table how many boards to expect. I've adapted the behavior to always report the BMS as having no errors  inside of `App_SharedErrorTable_GetBoardsWithNoErrors` and adapted the tests to account for this, along with adding descriptive comments to explain the modifications. In the future when the all boards are removed from the table this kind of band-aid workaround won't be necessary

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -610,4 +616,14 @@ TEST_F(DimStateMachineTest, pdm_board_status_led_control_with_multiple_errors)
...
+{
+    // Test to ensure BoardLeds enum matches Boards enum for use in DIM drive state
+    ASSERT_EQ(BMS_LED, BMS);
+    ASSERT_EQ(DCM_LED, DCM);
+    ASSERT_EQ(DIM_LED, DIM);
+    ASSERT_EQ(FSM_LED, FSM);
+    ASSERT_EQ(PDM_LED, PDM);
+}
+
 } // namespace StateMachineTest
```
Eddy13013, line 629: can we add a test here to check that when a fault is set, that we stay in fault state until the fault is cleared. i.e. we don't cycle between init and fault. 

also, I may have missed this test so let me know if I did

boards/shared/Src/App/App_SharedErrorTable.c
```diff
@@ -263,7 +246,13 @@ void App_SharedErrorTable_GetBoardsWithNoErrors(const struct ErrorTable *error_t
 
     for (enum Board board = 0; board < NUM_BOARDS; board++)
     {
-        if (!App_SharedError_IsBoardInList(&boards_with_errors, board))
+        // BMS not in table, always return as not containing errors
```
Eddy13013, line 249: I'm not sure how the error table is defined, but can we remove the BMS entirely from this table without breaking the entire system?

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -238,23 +238,23 @@ SG_ Duty_Cycle : 0|32@1- (1,0) [0|1E2] "%" DEBUG
...
-SG_ CELL_OVERVOLTAGE_FAULT : 2|1@1+ (1,0) [0|1] "" DEBUG
-SG_ MODULE_COMM_ERROR: 3|1@1+ (1,0) [0|1] "" DEBUG
-SG_ CELL_UNDERTEMP_FAULT: 4|1@1+ (1,0) [0|1] "" DEBUG
-SG_ CELL_OVERTEMP_FAULT: 5|1@1+ (1,0) [0|1] "" DEBUG
-SG_ CHARGER_FAULT_DETECTED : 6|1@1+ (1,0) [0|1] "" DEBUG
-SG_ HAS_REACHED_MAX_V : 7|1@1+ (1,0) [0|1] "" DEBUG
-SG_ CHARGING_EXT_SHUTDOWN_OCCURRED : 8|1@1+ (1,0) [0|1] "" DEBUG
-SG_ TS_OVERCURRENT_FAULT : 9|1@1+ (1,0) [0|1] "" DEBUG
-SG_ PRECHARGE_ERROR : 10|1@1+ (1,0) [0|1] "" DEBUG
+SG_ CHARGER_DISCONNECTED_IN_CHARGE_STATE : 0|1@1+ (1,0) [0|1] "" DIM
```
Eddy13013, line 247: Instead of sending all messages to the DIM, we could send the BMS State to the DIM. I'm if this reduces any bus load but I think it would at least simplify this communication

Eddy13013: in fact you did that exactly above. So these can change back to DEBUG probably for now. If we decide to implement more descriptive faults we can change this, but I leave that up to the dashboard team.

scripts/codegen/ErrorId/generate_error_ids.py
```diff
@@ -107,15 +94,8 @@
...
-            raise KeyError('Could not find non critical error message for %s' % board)
-
-        try:
-            can_msg = database.get_message_by_name(board + '_FAULTS')
-            enum_members['faults'][board] = \
-                ['    %s_FAULTS_%s, \\' %(board, signal.name.upper()) for signal in can_msg.signals]
-        except KeyError:
-            if board == "BMS":
-                raise KeyError('Could not find AIR shutdown error message for %s' % board)
+            if board != "BMS":
```
Eddy13013, line 97: again, no way to completely remove the BMS from the list of boards in the error table?

Eddy13013 (changes requested):

scripts/codegen/CAN/App_CanMsgs.dbc
will-chaba: Yup good call, I forgot to switch these back.

scripts/codegen/ErrorId/generate_error_ids.py
will-chaba: The issue here is that BMS_AIR_SHUTDOWN_ERRORS is renamed to BMS_FAULTS where as the rest of the boards use AIR_SHUTDOWN_ERRORS, this band-aid won't been needed once the rest of the board's messages are re-named to _FAULTS. This can be done quickly after the WARNINGS rename...

boards/BMS/Src/App/states/App_FaultState.c
```diff
@@ -24,14 +24,12 @@ static void FaultStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
-    struct ErrorTable *const error_table = App_BmsWorld_GetErrorTable(world);
+    struct BmsWorld *world = App_SharedStateMachine_GetWorld(state_machine);
+    struct Airs *    airs  = App_BmsWorld_GetAirs(world);
 
-    bool is_error_table_cleared = !App_SharedErrorTable_HasAnyAirShutdownErrorSet(error_table);
-    bool is_air_negative_open   = !App_Airs_IsAirNegativeClosed(airs);
+    bool is_air_negative_open = !App_Airs_IsAirNegativeClosed(airs);
 
-    if (is_error_table_cleared && is_air_negative_open)
+    if (is_air_negative_open)
```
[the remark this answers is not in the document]
will-chaba: There is a even simpler fix. Moving the call to `App_AllStatesRunOnTick100Hz(state_machine);` to after this check allows all-states to over-write the next-state to fault state if there are any active faults. I can't see a reason why this change would cause any issues but let me know if it does!

boards/DIM/Test/Src/Test_StateMachine.cpp
will-chaba: I added a test, we were in fact cycling. Also fixed that, see above.

boards/shared/Src/App/App_SharedErrorTable.c
will-chaba: To remove the BMS from the table completely we would need to remove the BMS from the Boards enum. See the discussion above.

boards/DIM/Src/App/states/App_DriveState.c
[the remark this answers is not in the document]
gtaharaedmonds: Good point, and yeah actually there probably will be aperiodic alerts that don't indicate a warning, such as startup messages.

boards/BMS/Src/App/states/App_FaultState.c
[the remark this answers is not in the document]
gtaharaedmonds: Hmm that's a little too jank. Can you just check `App_Accumulator_CheckFaults` and `App_TractveSystem_CheckFaults` here instead?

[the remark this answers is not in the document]
will-chaba: I could, but that would meant that `App_Accumulator_CheckFaults` and `App_TractiveSystem_CheckFaults` would be getting called twice inside `FaultStateRunOnTick100Hz`. If we can afford that then sure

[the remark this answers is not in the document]
gtaharaedmonds: Is the concern performance? It's not ideal, but shouldn't be significant. 
We call the `AllStatesRunOnTick` function first in every state, so let's keep doing so, to avoid any unexpected behavior down the line.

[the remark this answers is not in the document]
will-chaba: Okay so long as it's not a performance concern that sounds like a good idea

boards/BMS/Src/App/states/App_FaultState.c
```diff
@@ -24,14 +24,18 @@ static void FaultStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+    struct Accumulator *      accumulator    = App_BmsWorld_GetAccumulator(world);
+    struct TractiveSystem *   tractiveSystem = App_BmsWorld_GetTractiveSystem(world);
+    struct BmsCanTxInterface *can_tx         = App_BmsWorld_GetCanTx(world);
+
+    const bool is_air_negative_open = !App_Airs_IsAirNegativeClosed(airs);
+    const bool are_bms_faults_cleared =
+        !(App_TractveSystem_CheckFaults(can_tx, tractiveSystem) ||
+          App_Accumulator_CheckFaults(can_tx, accumulator, tractiveSystem));
+
+    if (is_air_negative_open && are_bms_faults_cleared)
```
gtaharaedmonds, line 38: Yeah this looks good

boards/DIM/Inc/App/states/App_DriveState.h
```diff
@@ -1,9 +1,36 @@
...
+typedef enum
+{
+    DCM_LED,
+    DIM_LED,
+    FSM_LED,
+    PDM_LED,
+    BMS_LED,
+    NUM_BOARD_LEDS,
+} BoardLeds;
```
gtaharaedmonds, line 16: Nit: Can you move this back to the `.c`, since I don't think any other file needs access to this.
Generally its better to put the least amount of code in the `.h`.

scripts/codegen/ErrorId/generate_error_ids.py
gtaharaedmonds: Yeah wouldn't recommend removing from `supported_boards` since that's currently used to error check cube and can code generation
As temporary solution this seems fine

Eddy13013 (approved):

PR #855, opened 2022-11-19
BMS-Charging-Update

KelmLelm:
## Summary

Relevant Jira: [ELEC-41](https://ubcformulaelectric.atlassian.net/browse/ELEC-41)
#### Init State:
- If charger is connected, the board waits for a CAN message to enable charging (Enters Pre-Charge)
#### Pre-Charge State:
- Failing Pre-Charge for charging will disable the CAN message . It will also send it back to Init State if Pre-Charge has not failed more than three times and if it is has been more than three times then it is sent into Fault State.
#### All States:
- If the charger is connected, it checks if the charger has thrown a fault it will get fault state.
- If the charger is connected and AIRs- opens (Shutdown loop activates), sends BMS to fault state.
- If the charger is not connected it checks if the car is in charging state which would indicate that the charger was disconnected midway through charging. It would then go into fault state.
#### Charge State:
- Disabling CAN message for charging and charger has been put into RunOnExit.  
- Checks all at once if the CAN message is disabled, external shutdown loop has triggered or if charging has completed and enters Init State if any are true.
- Checking if charging is complete is done by comparing the current being supplied to 0.05C charging.

## Testing Done

- With no enabling CAN message to charge while the charger is connected in Init state, the BMS should remain in Init state waiting for charging CAN message.
- When the charger is connected in Init state and the CAN message is given, the BMS should go into Precharge state
- After charger is connected and the CAN message enabling charging is sent and the system waits for slightly longer, BMS should go into Charge state (Successful precharge should close AIRs+.
- Starting in Charge state, with CAN message enabled and the charger connected with less than full charge, the BMS will remain in Charge state.
- In Charge state, when current from charger indicates full charge on the cells, the BMS stops charging, switches off the charging CAN message and goes back into Init State and opens AIRs+.
- In Charge state, will exit back to Init state in similar manner when false CAN charging message is sent.
- In Charge state, will go into fault state if the charger gives a fault. Will also disable CAN charging message.
- In Charge state, will go into fault state if charger is connected and the shutdown loop activates (AIRs- is open)
- Modified earlier test with charger disconnecting while in Charge state sending BMS to fault state.

## Checklist

[ELEC-41]: https://ubcformulaelectric.atlassian.net/browse/ELEC-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -1,5 +1,13 @@
 #include "App_Accumulator.h"
 
+// Min and Max cell temperatures depending on state
```
Eddy13013, line 3: this group is present in Accumulator.h. you can remove this

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -157,6 +158,16 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
         status = false;
         App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
     }
+    // Checks if the charger has thrown a fault, the disabling of the charger, etc is done with ChargeStateRunOnExit
+    else if (App_Charger_IsConnected(charger))
```
Eddy13013, line 162: this should be part of its own if statement. I believe the if above is designed to wait until the voltage measurements of the Module boards has settled. Charging is independent of this and therefore should be its own standalone if statement

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,16 +27,18 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
-        struct BmsCanTxInterface *can_tx      = App_BmsWorld_GetCanTx(world);
-        struct Charger *          charger     = App_BmsWorld_GetCharger(world);
-        struct Accumulator *      accumulator = App_BmsWorld_GetAccumulator(world);
-        struct Airs *             airs        = App_BmsWorld_GetAirs(world);
+        struct BmsWorld *            world   = App_SharedStateMachine_GetWorld(state_machine);
+        struct BmsCanTxInterface *   can_tx  = App_BmsWorld_GetCanTx(world);
+        struct BmsCanRxInterface *   can_rx  = App_BmsWorld_GetCanRx(world);
+        struct Charger *             charger = App_BmsWorld_GetCharger(world);
+        struct Airs *                airs    = App_BmsWorld_GetAirs(world);
+        struct TractiveSystem *const ts      = App_BmsWorld_GetTractiveSystem(world);
```
Eddy13013, line 35: interesting that this was made constant. was there a specific reason for this?

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,16 +27,18 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
 
         static uint16_t ignore_chgr_fault_counter      = 0U;
         const bool      is_charger_disconnected        = !App_Charger_IsConnected(charger);
         bool            has_charger_faulted            = false;
         bool            has_external_shutdown_occurred = !App_Airs_IsAirNegativeClosed(airs);
+        bool            charge_over_can                = App_CanRx_BMS_CHARGER_GetSignal_IS_CHARGING_ENABLED(can_rx);
```
Eddy13013, line 41:
```suggestion
        bool            charging_enabled                = App_CanRx_BMS_CHARGER_GetSignal_IS_CHARGING_ENABLED(can_rx);
```
to stick with our naming convention

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -45,29 +49,31 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+        App_CanTx_SetPeriodicSignal_IS_CHARGING_COMPLETE(can_tx, has_charging_completed);
         App_CanTx_SetPeriodicSignal_CHARGING_EXT_SHUTDOWN_OCCURRED(can_tx, has_external_shutdown_occurred);
 
-        if (is_charger_disconnected || has_charger_faulted || has_reached_max_v || has_external_shutdown_occurred)
+        // If the current indicates charging is complete, the shutdown loop is initiated, charging is disabled over CAN
+        // or charger is disconnected Go back to init state
+        if (has_charging_completed || has_external_shutdown_occurred || !charge_over_can || is_charger_disconnected)
         {
-            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+            App_SharedStateMachine_SetNextState(state_machine, App_GetInitState());
```
Eddy13013, line 63: we need to ensure that AIR+ is opened again. it should either go in Charge RunOnExit or Init RunOnEntry. I think I prefer Init RunOnEntry because it will also apply to any other state without a transition through fault. Its nice to guarentee that AIR+ is open in Init State

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -45,29 +49,31 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
 static void ChargeStateRunOnExit(struct StateMachine *const state_machine)
 {
-    struct BmsWorld *world   = App_SharedStateMachine_GetWorld(state_machine);
-    struct Charger * charger = App_BmsWorld_GetCharger(world);
-
+    struct BmsWorld *         world   = App_SharedStateMachine_GetWorld(state_machine);
+    struct Charger *          charger = App_BmsWorld_GetCharger(world);
+    struct BmsCanTxInterface *can_tx  = App_BmsWorld_GetCanTx(world);
+    // If the charger is not turned off, or the CAN message for charging is still set to charge when exiting Charge
+    // State Disable them
```
Eddy13013, line 74:
```suggestion
    // state, disable them when leaving
```

boards/BMS/Src/App/states/App_InitState.c
```diff
@@ -31,15 +31,29 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+        struct Charger *          charger = App_BmsWorld_GetCharger(world);
+
+        bool is_charger_connected = App_Charger_IsConnected(charger);
 
 #ifndef BSPD_DEMO_MODE
         // don't allow pre_charge if in BSPD_DEMO_MODE
         if (App_Airs_IsAirNegativeClosed(airs) && (App_TractiveSystem_GetVoltage(ts) < TS_DISCHARGED_THRESHOLD_V))
         {
-            App_SharedStateMachine_SetNextState(state_machine, App_GetPreChargeState());
+            // If the charger is connected and the CAN message for charging is set to enabled
```
Eddy13013, line 46:
```suggestion
            // If the charger is connected AND! the CAN message for charging is set to enabled
```
just to highlight the AND part. At first glance this if statement seemed useless

Eddy13013:
```suggestion
            // If the charger is connected, the CAN message must also be set to continue to 
            // precharge
```
actually I think this describes the logic here better. 

as an aside: when writing comments, try to describe the higher-level function rather than restating the code in regular language

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -259,6 +259,7 @@ SG_ PRECHARGE_ERROR : 10|1@1+ (1,0) [0|1] "" DEBUG
 BO_ 110 BMS_CHARGER: 1 BMS
 SG_ Is_Connected : 0|1@1+ (1,0) [0|1] "" DEBUG
 SG_ IS_CHARGING_COMPLETE : 1|1@1+ (1,0) [0|1] "" DEBUG
+SG_ IS_CHARGING_ENABLED : 2|1@1+ (1,0) [0|1] "" BMS
```
Eddy13013, line 262:
```suggestion
SG_ CHARGING_STATUS : 2|1@1+ (1,0) [0|1] "" BMS
```
to make it more user friendly. i'm not totally sold on this name

scripts/codegen/CAN/App_CanMsgs.dbc
```diff
@@ -660,6 +661,7 @@ VAL_ 109 CHARGING_EXT_SHUTDOWN_OCCURRED 0 "OK" 1 "FAULT";
 VAL_ 109 TS_OVERCURRENT_FAULT 0 "OK" 1 "FAULT";
 VAL_ 109 PRECHARGE_FAULT 0 "OK" 1 "FAULT";
 VAL_ 110 IS_CHARGING_COMPLETE 0 "FALSE" 1 "TRUE";
+VAL_ 110 IS_CHARGING_ENABLED 0 "FALSE" 1 "TRUE";
```
Eddy13013, line 664:
```suggestion
VAL_ 110 CHARGING_STATUS 0 "DISABLED" 1 "ENABLED";
```

Eddy13013 (changes requested):
Looks very good so far. Just a few small things to take a look at.

will-chaba (commented):
Looks good so far! just a couple of nitpicks

boards/BMS/Src/App/states/App_InitState.c
```diff
@@ -31,15 +31,29 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+            // Go into Pre-Charge State
+            if (is_charger_connected && App_CanRx_BMS_CHARGER_GetSignal_IS_CHARGING_ENABLED(can_rx))
+            {
+                App_SharedStateMachine_SetNextState(state_machine, App_GetPreChargeState());
+            }
+            // If charger isn't connected, go into drive state
```
gtaharaedmonds, line 52: This comment is incorrect - we definitely want to precharge!

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -45,29 +49,31 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
             has_charger_faulted = App_Charger_HasFaulted(charger);
         }
 
-        uint8_t    segment = 0U;
-        uint8_t    cell    = 0U;
-        const bool has_reached_max_v =
-            App_Accumulator_GetMaxVoltage(accumulator, &segment, &cell) > MAX_CELL_VOLTAGE_THRESHOLD;
+        const bool has_charging_completed = App_TractiveSystem_GetCurrent(ts) <= CURRENT_AT_MAX_CHARGE;
```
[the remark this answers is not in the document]
gtaharaedmonds: We could also OR this with cells have reached max voltage, just in case the charger tries to overcharge the cells. But if that's not a realistic worry, this is fine.

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,16 +27,18 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
-        struct Accumulator *      accumulator = App_BmsWorld_GetAccumulator(world);
-        struct Airs *             airs        = App_BmsWorld_GetAirs(world);
+        struct BmsWorld *            world   = App_SharedStateMachine_GetWorld(state_machine);
+        struct BmsCanTxInterface *   can_tx  = App_BmsWorld_GetCanTx(world);
+        struct BmsCanRxInterface *   can_rx  = App_BmsWorld_GetCanRx(world);
+        struct Charger *             charger = App_BmsWorld_GetCharger(world);
+        struct Airs *                airs    = App_BmsWorld_GetAirs(world);
+        struct TractiveSystem *const ts      = App_BmsWorld_GetTractiveSystem(world);
 
         static uint16_t ignore_chgr_fault_counter      = 0U;
```
gtaharaedmonds, line 37: For viz: This counter doesn't look like it actually does anything since its redefined to 0 at every time this function is called. Out of scope of this PR but can be addressed in another ticket

boards/BMS/Inc/App/App_Accumulator.h
```diff
@@ -15,6 +15,7 @@
 #define MIN_CELL_CHARGE_TEMP_DEGC (0.0f)
 #define MAX_CELL_VOLTAGE (4.2f)
 #define MIN_CELL_VOLTAGE (3.0f)
+#define C_RATE_TO_AMPS (17.7f)
```
gtaharaedmonds, line 18: Nit: It's good to explain where you got this number for constants like these

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -157,6 +158,16 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
...
+    {
+        if (App_Charger_HasFaulted(charger))
+        {
+            status = false;
+            App_CanTx_SetPeriodicSignal_IS_CHARGING_ENABLED(can_tx, false);
+            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+        }
```
gtaharaedmonds, line 169: Doesn't look like fault state checks that charger faults are cleared before going back into init, so this would cause cycling from fault <-> init. Prevent fault -> init transition until charger faults are cleared.

boards/BMS/Src/App/states/App_ChargeState.c
gtaharaedmonds: Would this also cause cycling? If charging complete go to init, but charger could still be plugged in, so precharge again and back to charge state, etc. etc.

gtaharaedmonds: Could be best to add it in both places tbh. Charge RunOnExit for clarity that contactors open after charger state, and Init RunOnEntry to be redundant.

gtaharaedmonds (changes requested):
Looks good overall! Can you provide some background on the IS_CHARGING_ENABLED signal however? Looks like its sent and received by the BMS, which doesn't really make sense.

scripts/codegen/CAN/App_CanMsgs.dbc
KelmLelm: What do we think of CHARGE_SWITCH?

boards/BMS/Inc/App/App_Accumulator.h
KelmLelm: Sounds good, I added the explanation in the ".h" file. I'll also mention it in my write-up.

boards/BMS/Src/App/states/App_ChargeState.c
KelmLelm: Sounds good, should be added in both places now

boards/DCM/Inc/stm32f3xx_hal_conf.h
```diff
@@ -5,7 +5,7 @@
   ******************************************************************************
   * @attention
   *
-  * <h2><center>&copy; COPYRIGHT(c) 2022 STMicroelectronics</center></h2>
+  * <h2><center>&copy; COPYRIGHT(c) 2023 STMicroelectronics</center></h2>
```
KelmLelm, line 8: Hey @gtaharaedmonds this won't cause any problems will it? I think I saw one of the tests that are automatically run when pushing to a branch flag a difference like this.

gtaharaedmonds: Confirmed, no problems

PR #856, opened 2022-11-26
Rename CAN NON_CRITICAL_ERRORS to WARNINGS

will-chaba:
### Summary

Renames CAN non_critical_errors to warnings to follow new [convention](https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/42205200/Alarms+-+Faults+Warnings)

### Changelist 

Any file with affected CAN messages was updated

### Testing Done

existing tests updated and ran to ensure all systems work as intended.

### Checklist


gtaharaedmonds (commented):
Is the plan for these to all be turned into aperiodic messages eventually?

will-chaba:
> Is the plan for these to all be turned into aperiodic messages eventually?

Yes I think the plan is for warnings to become aperiodic

boards/DCM/Test/Src/Test_StateMachine.cpp
```diff
@@ -283,12 +283,12 @@ TEST_F(DcmStateMachineTest, exit_fault_state_if_there_is_no_error)
 }
 
 // DCM-17
-TEST_F(DcmStateMachineTest, exit_fault_state_if_there_is_only_non_critical_error)
+TEST_F(DcmStateMachineTest, exit_fault_state_if_there_is_only_warning)
 {
     SetInitialState(App_GetFaultState());
 
     // Choose any non-critical fault, it doesn't have to come from DCM
```
Eddy13013, line 290:
```suggestion
    // Choose any warning, it doesn't have to come from DCM
```

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -533,10 +533,10 @@ TEST_F(DimStateMachineTest, fsm_board_status_led_control_with_critical_error)
 }
 
 // DIM-2
-TEST_F(DimStateMachineTest, fsm_board_status_led_control_with_non_critical_error)
+TEST_F(DimStateMachineTest, fsm_board_status_led_control_with_warning)
 {
     // Set any non-critical error and check that the FSM LED turns blue
```
Eddy13013, line 538:
```suggestion
    // Set any warning and check that the FSM LED turns blue
```

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -494,10 +494,10 @@ TEST_F(DimStateMachineTest, dcm_board_status_led_control_with_critical_error)
 }
 
 // DIM-2
-TEST_F(DimStateMachineTest, dcm_board_status_led_control_with_non_critical_error)
+TEST_F(DimStateMachineTest, dcm_board_status_led_control_with_warning)
 {
     // Set any non-critical error and check that the DCM LED turns blue
```
Eddy13013, line 499:
```suggestion
    // Set any warning and check that the DCM LED turns blue
```

boards/DIM/Test/Src/Test_StateMachine.cpp
```diff
@@ -581,10 +581,10 @@ TEST_F(DimStateMachineTest, pdm_board_status_led_control_with_critical_error)
 }
 
 // DIM-2
-TEST_F(DimStateMachineTest, pdm_board_status_led_control_with_non_critical_error)
+TEST_F(DimStateMachineTest, pdm_board_status_led_control_with_warning)
 {
     // Set any non-critical error and check that the PDM LED turns blue
```
Eddy13013, line 586:
```suggestion
    // Set any warning and check that the PDM LED turns blue
```

boards/FSM/Src/App/App_SetPeriodicCanSignals.c
```diff
@@ -27,16 +26,16 @@ void App_SetPeriodicSignals_WheelSpeedInRangeChecks(const struct FsmWorld *world
     App_SetPeriodicCanSignals_InRangeCheck(
         can_tx, left_wheel_speed_in_range_check, App_CanTx_SetPeriodicSignal_LEFT_WHEEL_SPEED,
         App_CanTx_SetPeriodicSignal_LEFT_WHEEL_SPEED_OUT_OF_RANGE,
-        CANMSGS_FSM_NON_CRITICAL_ERRORS_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OK_CHOICE,
-        CANMSGS_FSM_NON_CRITICAL_ERRORS_LEFT_WHEEL_SPEED_OUT_OF_RANGE_UNDERFLOW_CHOICE,
-        CANMSGS_FSM_NON_CRITICAL_ERRORS_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OVERFLOW_CHOICE);
+        CANMSGS_FSM_WARNINGS_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OK_CHOICE,
```
Eddy13013, line 29:
```suggestion
        CANMSGS_FSM_WARNING_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OK_CHOICE,
```
I personally think this sounds better. It now sounds like "FSM Warning: Left Wheel Speed Out of Range!"

but I see why it was written like this, as in it falls under the warnings 'group'

Eddy13013: if you decide to change please change the rest to match

boards/shared/Src/Io/Io_SharedErrorTable.c
```diff
@@ -4,110 +4,105 @@
...
-    SET_ERROR(error_table, FSM_NON_CRITICAL_WATCHDOG_TIMEOUT, data->watchdog_timeout);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_BSPD_FAULT, data->bspd_fault);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_LEFT_WHEEL_SPEED_OUT_OF_RANGE, data->left_wheel_speed_out_of_range);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_RIGHT_WHEEL_SPEED_OUT_OF_RANGE, data->right_wheel_speed_out_of_range);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_FLOW_RATE_OUT_OF_RANGE, data->flow_rate_out_of_range);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_STEERING_ANGLE_OUT_OF_RANGE, data->steering_angle_out_of_range);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_BRAKE_PRESSURE_OUT_OF_RANGE, data->brake_pressure_out_of_range);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_BRAKE_PRESSURE_OPEN_OC, data->brake_pressure_open_oc);
-    SET_ERROR(error_table, FSM_NON_CRITICAL_BRAKE_PRESSURE_OPEN_SC, data->brake_pressure_open_sc);
+    SET_ERROR(error_table, FSM_WARNING_WATCHDOG_TIMEOUT, data->watchdog_timeout);
```
Eddy13013, line 80: with reference to my earlier comment, I think this reads nicely

Eddy13013 (changes requested):
I like this a lot. Much much cleaner and easier to read! Just a few small changes to make and I will approve

Eddy13013:
> > Is the plan for these to all be turned into aperiodic messages eventually?
> 
> Yes I think the plan is for warnings to become aperiodic

Yes that is the plan

boards/FSM/Src/App/App_SetPeriodicCanSignals.c
will-chaba: I looked into this quite a bit, and while I agree it looks nice, I don't think its worth what it'd cost to achieve. In order to get the codegen to generate the can message choices like you have above, the message 'groups' in the .dbc would need to be changed to "FSM_WARNING" instead of "FSM_WARNINGS". 
```
BO_ 300 FSM_WARNING: 8 FSM
...
SG_ LEFT_WHEEL_SPEED_OUT_OF_RANGE: 9|2@1+ (1,0) [0|2] "" DEBUG
SG_ RIGHT_WHEEL_SPEED_OUT_OF_RANGE: 11|2@1+ (1,0) [0|2] ""  DEBUG
...
```

All that being said, the actual calls to send the messages in the code would not change, as they do not include "warning" in them IE:
 ```
App_CanTx_SetPeriodicSignal_LEFT_WHEEL_SPEED_OUT_OF_RANGE(...)
```
Only the generated choices would change:
```
#define CANMSGS_FSM_WARNING_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OK_CHOICE (0u)
#define CANMSGS_FSM_WARNING_LEFT_WHEEL_SPEED_OUT_OF_RANGE_UNDERFLOW_CHOICE (1u)
#define CANMSGS_FSM_WARNING_LEFT_WHEEL_SPEED_OUT_OF_RANGE_OVERFLOW_CHOICE (2u)
```

If we are okay with re-naming the groups to "_WARNINGS" then it's easily accomplished, in fact I've already pushed the required changes, but I feel like we should revert and leave it as "_WARNINGS"

boards/shared/Src/Io/Io_SharedErrorTable.c
will-chaba: Its definitely more concise

boards/FSM/Src/App/App_SetPeriodicCanSignals.c
gtaharaedmonds: I agree with Will here, I think keeping it "_WARNINGS" makes sense because that comes from the CAN message name, which is the set of warnings, not a single warning. 
But I don't mind either way, up to you guys.

Eddy13013 (approved):

PR #859, opened 2022-12-11
Moved accumulator voltage stats to App from IO

Eddy13013:
### Summary
Moved cell voltages from IO to App. Added code to get an individual cell voltage in App. Cleaned up IO.

### Changelist 

- Voltages are stored in app instead of io, makes it easier to view variables when debugging
- Moved voltage stat calculations to App
- Moved accumulator #defines to Accumulator.h
- Cleaned up IO code

### Testing Done
- BMS unit tests are passing on my machine

### Checklist


boards/BMS/Inc/App/App_Accumulator.h
```diff
@@ -8,6 +8,21 @@
...
+typedef enum
+{
+    ACCUMULATOR_SEGMENT_0 = 0U,
+    ACCUMULATOR_SEGMENT_1,
+    ACCUMULATOR_SEGMENT_2,
+    ACCUMULATOR_SEGMENT_3,
+    ACCUMULATOR_SEGMENT_4,
+    ACCUMULATOR_SEGMENT_5,
+    ACCUMULATOR_NUM_OF_SEGMENTS,
+} AccumulatorSegments_E;
```
gtaharaedmonds, line 21: We should really clarify our conventions. Whether or not we typedef, whether or not we do _E/_T, etc.
I don't care for this PR because there's a big mix in our repo rn, but probably best we make conventions and stick to them (maybe at start of Jan?)

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -100,30 +103,105 @@ void App_Accumulator_Destroy(struct Accumulator *accumulator)
     free(accumulator);
 }
 
+void App_Accumulator_InitRunOnEntry(const struct Accumulator *const accumulator)
+{
+    // Configure the cell monitoring chips. Disable discharge at startup
+    accumulator->config_monitoring_chip();
+}
```
gtaharaedmonds, line 110: Nit: Rename this function to what it actually does, instead of where it's run. This doesn't really give any useful info when reading InitState.

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -20,6 +20,20 @@ enum AccumulatorMonitorState
...
+    struct
+    {
+        uint8_t  segment;
+        uint8_t  series_elem;
+        uint16_t voltage;
+    } min, max;
+    uint32_t segment_avg[ACCUMULATOR_NUM_OF_SEGMENTS];
+    uint32_t segment[ACCUMULATOR_NUM_OF_SEGMENTS];
+    uint32_t pack_avg;
+    uint32_t pack;
```
gtaharaedmonds, line 34: Nit: Include voltage in these variable names so it's clear these are all voltages

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -100,30 +103,105 @@ void App_Accumulator_Destroy(struct Accumulator *accumulator)
...
+        // sum the segment voltages into a pack voltage
+        temp_voltage_stats.pack += temp_voltage_stats.segment[segment];
+        // average of se voltages in pack
+        temp_voltage_stats.pack_avg = temp_voltage_stats.pack / segment * ACCUMULATOR_NUM_OF_SERIES_ELEM_PER_SEGMENT;
+    }
+
+    accumulator->voltage_stats = temp_voltage_stats;
+}
+
+float App_Accumulator_GetSEVoltage(const struct Accumulator *const accumulator, uint8_t segment, uint8_t series_elem)
```
gtaharaedmonds, line 168: I think its more clear to use the enum types in the parameters

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -165,7 +243,7 @@ void App_Accumulator_RunOnTick100Hz(struct Accumulator *const accumulator)
     {
         case GET_CELL_VOLTAGE_STATE:
 
-            UPDATE_PEC15_ERROR_COUNT(accumulator->read_cell_voltages(), accumulator->num_comm_tries)
+            App_Accumulator_ReadCellVoltage(accumulator);
 
             // Write to configuration register to configure cell discharging
             // TODO: re-enable
```
gtaharaedmonds, line 249: The state variable here will always be GET_CELL_VOLTAGE_STATE - is that intentional?

boards/BMS/Inc/App/App_Accumulator.h
```diff
@@ -16,18 +31,16 @@
...
     bool (*config_monitoring_chip)(void),
     bool (*write_cfg_registers)(void),
     bool (*start_voltage_conv)(void),
-    bool (*read_cell_voltages)(void),
-    float (*get_min_cell_voltage)(uint8_t *, uint8_t *),
-    float (*get_max_cell_voltage)(uint8_t *, uint8_t *),
-    float (*get_segment_voltage)(AccumulatorSegments_E),
-    float (*get_pack_voltage)(void),
-    float (*get_avg_cell_voltage)(void),
+    bool (*read_cell_voltages)(uint16_t[ACCUMULATOR_NUM_OF_SEGMENTS][ACCUMULATOR_NUM_OF_SERIES_ELEM_PER_SEGMENT]),
```
gtaharaedmonds, line 43: Didn't know you could pass arrays like this in C -> is this still the same under the hood as just passing by pointer? Or will the array be copied?

boards/BMS/Src/App/App_Accumulator.c
gtaharaedmonds: (out of scope of this PR)

boards/BMS/Inc/App/App_Accumulator.h
will-chaba: C always passes pointers when passing arrays, so in theory, this should be the same

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -20,6 +20,20 @@ enum AccumulatorMonitorState
...
 };
 
+struct VoltageStats
+{
+    struct
+    {
+        uint8_t  segment;
+        uint8_t  series_elem;
+        uint16_t voltage;
+    } min, max;
```
will-chaba, line 30: super nit: I'd add a line here for readability 
```suggestion
    } min, max;
    
```

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -100,30 +103,105 @@ void App_Accumulator_Destroy(struct Accumulator *accumulator)
...
-    // Configure the cell monitoring chips. Disable discharge at startup
-    accumulator->config_monitoring_chip();
+    struct VoltageStats temp_voltage_stats = { .min = { .segment = 0U, .series_elem = 0U, .voltage = UINT16_MAX },
+                                               .max = { .segment = 0U, .series_elem = 0U, .voltage = 0x0000U },
+                                               .segment_avg = { 0U },
+                                               .segment     = { 0U },
+                                               .pack_avg    = 0U,
+                                               .pack        = 0U };
+
+    // find the min and max voltages
```
will-chaba, line 131: Nit: should stay consistent with the capitalization of first character in these comments (more below)
```suggestion
    // Find the min and max voltages
```

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -80,11 +82,17 @@ static void App_AdvertisePackPower(struct Accumulator *accumulator, struct Tract
...
-        min(App_SharedProcessing_LinearDerating(
+        MIN(App_SharedProcessing_LinearDerating(
                 max_cell_temp, MAX_POWER_LIMIT_W, CELL_ROLL_OFF_TEMP_DEGC, CELL_FULLY_DERATED_TEMP),
             MAX_POWER_LIMIT_W);
 }
 
+void App_AllStates_Init()
+{
+    // Reset accumulator settle count
+    acc_meas_settle_count = 0U;
```
gtaharaedmonds, line 93: We should avoid declaring static variables in functions, since there is no way to reset them between test runs.
There were some bugs in the tests that the settle count would count to 3 on the first test, and then stay 3 for the rest.

boards/shared/Inc/App/App_SharedMacros.h
```diff
@@ -13,8 +13,7 @@
 #define __BASENAME__(path) \
     (__builtin_strrchr(path, '/') ? __builtin_strrchr(path, '/') + 1 : path)
 
-#define min(x, y) (((x) < (y)) ? (x) : (y))
-
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
```
gtaharaedmonds, line 16: There was a name conflict with `min` when running for x86. Not sure why this started now, but I suspect this is a windows thing.

gtaharaedmonds (approved):

will-chaba (approved):
Looks good to me! found a small nitpick that should be checked

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -100,30 +157,72 @@ void App_Accumulator_Destroy(struct Accumulator *accumulator)
...
+    {
+        return 0.0f;
+    }
+
+    return accumulator->voltage_stats.segment_voltages[segment] / ACCUMULATOR_NUM_SERIES_CELLS_PER_SEGMENT;
+}
+
+float App_Accumulator_GetSegmentVoltage(const struct Accumulator *const accumulator, uint8_t segment)
+{
+    if (segment > ACCUMULATOR_NUM_SEGMENTS)
```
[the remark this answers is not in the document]
gtaharaedmonds: Good catch, should be all fixed now

PR #865, opened 2023-01-22
Port DCM to f4 

philam2001:
### Summary

Update F3 DCM board firmware to be compatible with F4 chip

### Changelist 

- DCM now uses the F4 chip
- Removed references to io_rgbled and io_inverter as updated boards don't require these
- Update pin mapping for gpios and other pins
- Update GPIO naming references from other files 
- Update file references from F3 to F4
- Update Io_Hal and Io_SharedAdc in the shared folders to account for DCM not using ADC 

### Checklist
- [x] Need to disable adc and spi for DCM in the IOC, right now can't build without it enabled
- [ ] Need to test and flash on DCM


boards/DCM/DCM.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<targetDefinitions xmlns="http://openstm32.org/stm32TargetDefinitions" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://openstm32.org/stm32TargetDefinitions stm32TargetDefinitions.xsd">
+  <board id="DCM">
+    <name>DCM</name>
+    <mcuId>STM32F412RGTx</mcuId>  <!-- mcu-->
+    <dbgIF>SWD</dbgIF>
+    <dbgDEV>ST-LinkV2-1</dbgDEV>
+  </board>
+</targetDefinitions>
```
gtaharaedmonds, line 9: I think this file can be deleted

boards/DCM/Src/main.c
```diff
@@ -217,28 +204,23 @@ int main(void)
 
     /* Create the thread(s) */
     /* definition and creation of Task1Hz */
-    osThreadStaticDef(Task1Hz, RunTask1Hz, osPriorityLow, 0, TASK1HZ_STACK_SIZE, Task1HzBuffer, &Task1HzControlBlock);
+    osThreadStaticDef(Task1Hz, RumTask1Hz, osPriorityLow, 0, 512, Task1HzBuffer, &Task1HzControlBlock);
```
gtaharaedmonds, line 207: Typo: Rum -> Run

gtaharaedmonds (commented):
Looks great. Can you delete the boards/DCM/Drivers/STM32F3xx_HAL_Driver folder entirely? Can't tell if you already did or not

gtaharaedmonds (approved):
lgtm

PR #884, opened 2023-03-05
Torque vectoring

akoen:
### Summary
Implements a torque vectoring algorithm as described at https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/51904893/Traction+Control+Torque+Vectoring

### Changelist 

- Implements the new algorithm in place of SetPeriodicTorqueRequests

### Testing Done
- None

### Checklist


boards/DCM/Src/App/torquevectoring/App_ActiveDifferential.c
```diff
@@ -0,0 +1,37 @@
...
+    outputs->torque_left_Nm = torque_left_Nm * scale;
+    outputs->torque_right_Nm = torque_right_Nm * scale;
+}
+
+float App_ActiveDifferential_SteeringAngleToSpeedDelta(float wheel_angle) {
+    // FIXME(akoen): Current implementation assumes
+    // angle > 0 = right
+    // angle < = left
+
+    return (TRACK_WIDTH_mm*tanf(wheel_angle * ((float) M_PI / 180.0f))) / (2 * WHEELBASE_mm);
```
gtaharaedmonds, line 32: General comment for this entire PR: Consider breaking calculations into multiple lines where it would improve readability. For ex. this would probably read better as:
```c
const float wheel_angle_rad = wheel_angle * ((float) M_PI / 180.0f));
return TRACK_WIDTH_MM * tanf(wheel_angle_rad) / (2 * WHEELBASE_MM);
```

gtaharaedmonds: Also it's pretty common to implement trig functions via lookup table on embedded, idk if that's necessary here though

gtaharaedmonds: Also I'd imagine the degree to rad conversion could be added to the `App_SharedMacros.h` file, seems useful to have in shared code

boards/DCM/Src/App/torquevectoring/App_PowerLimiting.c
```diff
@@ -0,0 +1,26 @@
...
+    // ============== Calculate max powers =================
+    float P_max_motor_temps =
+        max_motor_temp > MOTOR_TEMP_CUTOFF_c
+            ? POWER_LIMIT_CAR_kW
+            : POWER_LIMIT_CAR_kW - (max_motor_temp - MOTOR_TEMP_CUTOFF_c) * MOTOR_TEMP_POWER_DECREMENTING_RATIO;
+    float P_max_accelerator = power_limiting_inputs->accelerator_pedal_percent * POWER_LIMIT_CAR_kW;
+    // =========== Take min of max powers ==================
+    return fminf(
+        P_max_motor_temps,
+        fminf(power_limiting_inputs->available_battery_power_kW, P_max_accelerator)); // triple min function
```
gtaharaedmonds, line 25: We have a `MIN3` macro in `App_SharedMacros.h`

boards/DCM/Src/App/torquevectoring/App_PowerLimiting.c
```diff
@@ -0,0 +1,26 @@
...
+ * @return A float for the maximum power allowed from the motor,
+ */
+float App_PowerLimiting_ComputeMaxPower(struct PowerLimiting_Inputs *power_limiting_inputs)
+{
+    float max_motor_temp = fmaxf(power_limiting_inputs->left_motor_temp_C, power_limiting_inputs->right_motor_temp_C);
+    // ============== Calculate max powers =================
+    float P_max_motor_temps =
+        max_motor_temp > MOTOR_TEMP_CUTOFF_c
+            ? POWER_LIMIT_CAR_kW
+            : POWER_LIMIT_CAR_kW - (max_motor_temp - MOTOR_TEMP_CUTOFF_c) * MOTOR_TEMP_POWER_DECREMENTING_RATIO;
```
gtaharaedmonds, line 20: Is the greater than sign incorrect here?

gtaharaedmonds: Would also recommend changing this to an if/else so it's easier to read

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
```diff
@@ -0,0 +1,97 @@
...
+        // Traction Control
+        traction_control_inputs.motor_speed_left_rpm = motor_speed_left_rpm;
+        traction_control_inputs.motor_speed_right_rpm = motor_speed_right_rpm;
+        traction_control_inputs.torque_left_Nm = active_differential_outputs.torque_left_Nm;
+        traction_control_inputs.torque_right_Nm = active_differential_outputs.torque_right_Nm;
+        App_TractionControl_ComputeTorque(&traction_control_inputs, &traction_control_outputs);
+
+        // Inverter Torque Requests
+        App_CanTx_DCM_LeftInverterCommand_TorqueCommand_Set(traction_control_outputs.torque_left_final_Nm);
+        App_CanTx_DCM_RightInverterCommand_TorqueCommand_Set(traction_control_outputs.torque_right_final_Nm);
```
gtaharaedmonds, line 79: Final torque clamp? Doesn't look like ComputeTorque does that, and imo it's better to do it here anyway to be super explicit about clamping

gtaharaedmonds, line 79: This looks awesome

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
```diff
@@ -0,0 +1,97 @@
...
+    if (accelerator_pedal_percent > 0.0f) {
+        // Reset control loops if timeout elapsed
+        TimerState timeout = App_Timer_UpdateAndGetState(&pid_timeout);
+        if (timeout == TIMER_STATE_EXPIRED) {
+            App_PID_RequestReset(&pid_power_correction);
+            pid_power_correction_factor = 1.0f;
+            App_PID_RequestReset(traction_control_inputs.pid);
+
+        }
+        App_Timer_Restart(&pid_timeout);
```
gtaharaedmonds, line 46: Imo the code will be easier to read + maintain if you only handle the state machine logic between regen, torque vectoring, and zero torque here. And then call other functions in this file that handle what each state actually does. 
Right now it's a little tricky to read what code is for the algorithm and what code is for the state machine, which I'd imagine will only get worse as the algorithms get more complicated.

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
```diff
@@ -0,0 +1,97 @@
...
+    App_Timer_InitTimer(&pid_timeout, PID_TIMEOUT_ms);
+}
+
+void App_TorqueVectoring_Run(void)
+{
+    // FIXME: Replace with CAN messages
+    float accelerator_pedal_percent = 2.0;
+    bool regen_engaged = true;
+
+    if (accelerator_pedal_percent > 0.0f) {
```
gtaharaedmonds, line 37: Should probably be some small threshold, not exactly zero

gtaharaedmonds (commented):
Looks great so far

boards/DCM/Src/App/torquevectoring/App_PID.c
```diff
@@ -0,0 +1,38 @@
...
+    if (output < pid->out_min)
+    {
+        output = pid->out_min;
+    }
+    else if (output > pid->out_max)
+    {
+        output = pid->out_max;
+    }
+
+    return output;
```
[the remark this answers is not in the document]
akoen: Derivative-on-measurement implemented. I like having clamping within the implementation. It's used by both pid controllers

boards/DCM/Inc/App/torquevectoring/App_TractionControl.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+#include "torquevectoring/App_TorqueVectoring.h"
+
+void App_TractionControl_ComputeTorque(TractionControl_Inputs *inputs, TractionControl_Outputs *outputs);
+
+float ComputeSlip(float motor_speed_rpm, float front_wheel_speed_rpm);
```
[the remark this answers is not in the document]
akoen: Don't I need to header if I'm going to write tests for it?

boards/DCM/Inc/App/torquevectoring/App_TorqueVectoring.h
```diff
@@ -0,0 +1,45 @@
...
+    float motor_speed_left_rpm;
+    float motor_speed_right_rpm;
+    float power_max_kW;
+} ActiveDifferential_Inputs;
+
+typedef struct ActiveDifferential_Outputs
+{
+    float torque_left_Nm;
+    float torque_right_Nm;
+} ActiveDifferential_Outputs;
```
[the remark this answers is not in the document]
akoen: We're going to keep it this way for now, but may refactor in the future.

akoen:
Made a bunch of changes, should be very little left to do in terms of code for the current implementation.

boards/DCM/Inc/App/torquevectoring/App_TractionControl.h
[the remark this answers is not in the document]
gtaharaedmonds: whoops yep good point

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
```diff
@@ -0,0 +1,148 @@
...
+float wheel_speed_front_left_kph;
+float wheel_speed_front_right_kph;
+float motor_speed_left_rpm;
+float motor_speed_right_rpm;
+float battery_voltage;
+float current_consumption;
+float left_motor_temp_C;
+float right_motor_temp_C;
+float available_battery_power_kW;
+float steering_angle_deg;
```
gtaharaedmonds, line 37: I think these should all be static: https://stackoverflow.com/questions/2271902/static-vs-global

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
```diff
@@ -0,0 +1,148 @@
...
+    motor_speed_right_rpm       = (float)App_CanRx_INVR_MotorPositionInfo_MotorSpeed_Get();
+    battery_voltage             = App_CanRx_BMS_TractiveSystem_TsVoltage_Get();
+    current_consumption         = App_CanRx_BMS_TractiveSystem_TsCurrent_Get();
+    left_motor_temp_C           = App_CanRx_INVL_Temperatures3_MotorTemperature_Get();
+    right_motor_temp_C          = App_CanRx_INVR_Temperatures3_MotorTemperature_Get();
+    // TODO(akoen): Available power will soon be replaced by current + voltage messages
+    available_battery_power_kW = App_CanRx_BMS_AvailablePower_AvailablePower_Get();
+    steering_angle_deg = App_CanRx_FSM_Steering_SteeringAngle_Get();
+
+    if (accelerator_pedal_percent > 0.01f)
```
gtaharaedmonds, line 65: Maybe something a bit higher than this is safer, like 10%
Also note that PappsMappedPedalPercentage is sent in percent so 0-100, so this value above here is 0.01%

boards/DCM/Test/Src/torquevectoring/Test_ActiveDifferential.cpp
```diff
@@ -0,0 +1,131 @@
...
+    float expected_ratio_t1 = cl_t1/cr_t1;
+    float calculated_ratio_t1 = expected_torque_left_t1/expected_torque_right_t1;
+    float actual_ratio_t1 = actual_active_diff_outputs_t1.torque_left_Nm/actual_active_diff_outputs_t1.torque_right_Nm;
+    ASSERT_FLOAT_EQ(expected_torque_left_t1, actual_active_diff_outputs_t1.torque_left_Nm);
+    ASSERT_FLOAT_EQ(expected_torque_right_t1, actual_active_diff_outputs_t1.torque_right_Nm);
+    ASSERT_TRUE(actual_active_diff_outputs_t1.torque_right_Nm <= MOTOR_TORQUE_LIMIT_Nm);
+    ASSERT_TRUE(actual_active_diff_outputs_t1.torque_left_Nm <= MOTOR_TORQUE_LIMIT_Nm);
+    ASSERT_FLOAT_EQ(expected_ratio_t1,calculated_ratio_t1);
+    ASSERT_FLOAT_EQ(expected_ratio_t1, actual_ratio_t1);
+};
```
gtaharaedmonds, line 65: These could definitely be more readable
Maybe break this into sections + add more comments 
Right now its hard to tell what is actually being tested
Also any common setup/teardown code can go in SetUp and TearDown above, so any values used in all the tests can be declared as class variables and set in SetUp

boards/DCM/Test/Src/torquevectoring/Test_PID.cpp
```diff
@@ -0,0 +1,69 @@
...
+    protected:
+        void SetUp() override
+        {
+            return;
+        }
+        void TearDown() override
+        {
+            return;
+        }
+};
```
gtaharaedmonds, line 22: You can just delete this class if its not used
I believe the class is only used if you use the TEST_F macro vs. the TEST macro

gtaharaedmonds (commented):
looks awesome, main thing is documentation is lacking (especially in the tests)

boards/DCM/Src/App/torquevectoring/App_TorqueVectoring.c
akoen: Set it to 10%, although note that the percentage is scaled 0-1 above

boards/DCM/Test/Src/torquevectoring/Test_PID.cpp
akoen: @1mackenziekyle

boards/DCM/Test/Src/torquevectoring/Test_ActiveDifferential.cpp
akoen: @1mackenziekyle

gtaharaedmonds:
I believe this is long outdated so closing. Please reopen if that's incorrect.

PR #887, opened 2023-03-08
BMS Charging Update (fixed)

gtaharaedmonds:
## Summary

Relevant Jira: [ELEC-41](https://ubcformulaelectric.atlassian.net/browse/ELEC-41)
#### Init State:
- If charger is connected, the board waits for a CAN message to enable charging (Enters Pre-Charge)
#### Pre-Charge State:
- Failing Pre-Charge for charging will disable the CAN message . It will also send it back to Init State if Pre-Charge has not failed more than three times and if it is has been more than three times then it is sent into Fault State.
#### All States:
- If the charger is connected, it checks if the charger has thrown a fault it will get fault state.
- If the charger is connected and AIRs- opens (Shutdown loop activates), sends BMS to fault state.
- If the charger is not connected it checks if the car is in charging state which would indicate that the charger was disconnected midway through charging. It would then go into fault state.
#### Charge State:
- Disabling CAN message for charging and charger has been put into RunOnExit.  
- Checks all at once if the CAN message is disabled, external shutdown loop has triggered or if charging has completed and enters Init State if any are true.
- Checking if charging is complete is done by comparing the current being supplied to 0.05C charging.

## Testing Done

- With no enabling CAN message to charge while the charger is connected in Init state, the BMS should remain in Init state waiting for charging CAN message.
- When the charger is connected in Init state and the CAN message is given, the BMS should go into Precharge state
- After charger is connected and the CAN message enabling charging is sent and the system waits for slightly longer, BMS should go into Charge state (Successful precharge should close AIRs+.
- Starting in Charge state, with CAN message enabled and the charger connected with less than full charge, the BMS will remain in Charge state.
- In Charge state, when current from charger indicates full charge on the cells, the BMS stops charging, switches off the charging CAN message and goes back into Init State and opens AIRs+.
- In Charge state, will exit back to Init state in similar manner when false CAN charging message is sent.
- In Charge state, will go into fault state if the charger gives a fault. Will also disable CAN charging message.
- In Charge state, will go into fault state if charger is connected and the shutdown loop activates (AIRs- is open)
- Modified earlier test with charger disconnecting while in Charge state sending BMS to fault state.

## Checklist

[ELEC-41]: https://ubcformulaelectric.atlassian.net/browse/ELEC-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


boards/BMS/Inc/App/states/App_ChargeState.h
```diff
@@ -1,9 +1,11 @@
...
 #include "App_SharedStateMachine.h"
-
+#include "App_Accumulator.h"
 /**
  * Get a pointer to the Charge State.
  * @return A pointer to the Charge State. THIS SHOULD NOT BE MODIFIED.
  */
 const struct State *App_GetChargeState(void);
+// 0.05C is standard for a boundary to consider full charge
+#define CURRENT_AT_MAX_CHARGE (0.05f * C_RATE_TO_AMPS);
```
will-chaba, line 11: This should be placed above function definitions, also add an empty line to end of file

boards/BMS/Test/Src/Test_Bms.cpp
```diff
@@ -1,3 +1,72 @@
 #include "Test_Bms.h"
 
 DEFINE_FFF_GLOBALS;
+//
```
will-chaba, line 4: Was the addition of this file intentional? Looks like it could be removed

can-bus/json/BMS/BMS_rx.json
```diff
@@ -3,6 +3,7 @@
         "DCM_Vitals",
         "FSM_Vitals",
         "PDM_Vitals",
-        "DIM_Vitals"
+        "DIM_Vitals",
+        "DEBUG_ChargingSwitch"
     ]
 }
```
will-chaba, line 9: missing newline at end of file

can-bus/json/DEBUG/DEBUG_alerts.json
```diff
@@ -0,0 +1,4 @@
+{
+    "msg_id": 700,
+    "alerts": []
+}
```
will-chaba, line 4: missing newline at end of file

can-bus/json/DEBUG/DEBUG_enum.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
will-chaba, line 2: missing newline at end of file

can-bus/json/DEBUG/DEBUG_rx.json
```diff
@@ -0,0 +1,3 @@
+{
+    "messages": []
+}
```
will-chaba, line 3: missing newline at end of file

can-bus/json/DEBUG/DEBUG_tx.json
```diff
@@ -0,0 +1,12 @@
...
+        "msg_id": 701,
+        "cycle_time": 100,
+        "description": "Value set by user to start charging BMS.",
+        "signals": {
+            "StartCharging": {
+                "bits": 1
+            }
+        }
+    }
+}
```
will-chaba, line 12: missing newline at end of file

will-chaba (changes requested):
Looks good, just some minor items to clean up

will-chaba (dismissed):
Looks good to me

boards/BMS/Inc/App/states/App_ChargeState.h
```diff
@@ -1,9 +1,11 @@
 #pragma once
 
 #include "App_SharedStateMachine.h"
-
+#include "App_Accumulator.h"
+// 0.05C is standard for a boundary to consider full charge
+#define CURRENT_AT_MAX_CHARGE (0.05f * C_RATE_TO_AMPS);
```
gtaharaedmonds, line 6: This isn't used in any other file, so you should define it in the .c file instead to limit scope

Pipfile.lock
gtaharaedmonds, line 1: Delete this file, we don't track this anymore

boards/BMS/Src/App/states/App_InitState.c
```diff
@@ -30,24 +32,32 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+            // to continue into Pre-Charge State, then Charge State
+            if (is_charger_connected && App_CanRx_DEBUG_ChargingSwitch_StartCharging_Get())
+            {
+                App_SharedStateMachine_SetNextState(state_machine, App_GetPreChargeState());
+            }
+            // If charger isn't connected, go into Pre-Charge State, then Drive State
+            else if (!is_charger_connected)
+            {
+                App_SharedStateMachine_SetNextState(state_machine, App_GetPreChargeState());
+            }
```
gtaharaedmonds, line 58: I feel like the use of an if/else here is confusing since both conditions lead to precharge state. 
Consider changing to a single if statement, something like:
```c
bool precharge_for_charging = is_charger_connected && App_CanRx_DEBUG_ChargingSwitch_StartCharging_Get();
if(precharge_for_charging || !is_charger_connected)
{
    App_SharedStateMachine_SetNextState(state_machine, App_GetPreChargeState());
}
```
```

File: boards/BMS/Src/App/states/App_InitState.c [hunk 10]
```diff
@@ -30,24 +32,32 @@ static void InitStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
         }
 #endif
-        //      #else?
-        struct HeartbeatMonitor *hb_monitor    = App_BmsWorld_GetHeartbeatMonitor(world);
-        const bool               is_missing_hb = !App_SharedHeartbeatMonitor_Tick(hb_monitor);
-        App_CanTx_BMS_Warnings_MissingHeartBeat_Set(is_missing_hb);
-        if (is_missing_hb)
-        {
-            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
-        }
```
gtaharaedmonds, line 50: You should revert this change here, I don't think this is supposed to be part of your PR

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -62,20 +62,24 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
...
+            if (next_state == App_GetFaultState())
+            {
+                App_PrechargeRelay_ResetFaultCounterVal(precharge_relay);
+            }
             App_SharedStateMachine_SetNextState(state_machine, next_state);
         }
-        else if (missing_hb)
-        {
-            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
-        }
```
gtaharaedmonds, line 78: Did you mean to delete all this heartbeat stuff?

boards/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -62,20 +62,24 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
...
         if (has_precharge_fault)
         {
             const struct State *next_state =
                 (precharge_fault_limit_exceeded) ? App_GetFaultState() : App_GetInitState();
+            App_CanRx_DEBUG_ChargingSwitch_StartCharging_Update(false);
+            if (next_state == App_GetFaultState())
+            {
+                App_PrechargeRelay_ResetFaultCounterVal(precharge_relay);
+            }
             App_SharedStateMachine_SetNextState(state_machine, next_state);
```
gtaharaedmonds, line 78: This logic might be clearer like:
```c
if(precharge_fault_limit_exceeded)
{
    App_PrechargeRelay_ResetFaultCounterVal(precharge_relay);
}
else
{
    App_SharedStateMachine_SetNextState(state_machine, App_GetInitState());
}
```

I feel like setting the next state ptr and then `if`ing based on that is confusing

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -43,24 +44,16 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
 
-        uint8_t    segment = 0U;
-        uint8_t    cell    = 0U;
-        const bool has_reached_max_v =
-            App_Accumulator_GetMaxVoltage(accumulator, &segment, &cell) > MAX_CELL_VOLTAGE_THRESHOLD;
+        const bool charging_completed = App_TractiveSystem_GetCurrent(ts) <= CURRENT_AT_MAX_CHARGE;
 
-        App_CanTx_BMS_Faults_ChargerDisconnectedInChargeState_Set(is_charger_disconnected);
         App_CanTx_BMS_Faults_ChargerFault_Set(has_charger_faulted);
-        App_CanTx_BMS_Charger_IsChargingComplete_Set(has_reached_max_v);
```
gtaharaedmonds, line 53: IsChargingComplete and ChargerDisconnectedInChargeState are important signals - you should change this back so they are actually being set

scripts/codegen/CAN/App_CanMsgs.dbc
gtaharaedmonds, line 1: We don't track the DBC file anymore, you should delete this

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -155,6 +156,23 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
...
+            App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+        }
+    }
+    else if (App_SharedStateMachine_GetCurrentState(state_machine) == App_GetChargeState())
+    {
+        status = false;
+        App_CanTx_BMS_Faults_ChargerDisconnectedInChargeState_Set(true);
+        App_CanRx_DEBUG_ChargingSwitch_StartCharging_Update(false);
+        App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
+    }
```
gtaharaedmonds, line 175: Why is this check here now vs. in App_ChargeState previously? I think it makes more sense there since that's where all the other charging faults are handled

boards/BMS/Src/App/states/App_ChargeState.c
gtaharaedmonds: Also the faults here don't actually put the board into fault state

gtaharaedmonds (commented):
Overall looks great but I think needs some work, take a look at the comments.
I haven't reviewed tests yet, will do so when other stuff is ironed out.

boards/BMS/Src/App/states/App_AllStates.c
will-chaba: This was part of Edwin's charging diagram, it was found to not work here at all because of how the charger sends its fault signal, basically, it sends the fault signal whenever it is not currently charging, even if there is nothing "wrong", so checking in all-states will perpetually send the BMS into fault state. 

TLDR: moved the check back to charge state

scripts/codegen/CAN/App_CanMsgs.dbc
will-chaba: what's the difference between this file and can-bus/dbcs/CanMsgs.dbc?

boards/BMS/Src/App/states/App_ChargeState.c
```diff
@@ -25,48 +29,77 @@ static void ChargeStateRunOnTick100Hz(struct StateMachine *const state_machine)
...
+        if (ignore_chgr_fault_counter >= CYCLES_TO_IGNORE_CHGR_FAULT)
         {
-            ignore_chgr_fault_counter++;
+            has_charger_faulted = App_Charger_HasFaulted(charger);
+            timeout_passed      = true;
+            UNUSED(ts);
+            // TODO: Re-implement charging_completed
+            //            if(-App_TractiveSystem_GetCurrent(ts) <=  0.85f && (App_TractiveSystem_GetCurrent(ts) < 0)){
+            //                charging_completed = true;
+            //            }
```
will-chaba, line 54: charging_completed was magically going true immediately even though the current was much greater than 0.85, could be a current sense glitch, but CAN read the value as significantly higher than that (greater than 2.0A minimum during our tests if I recall correctly.) The current plan is just to babysit the charger and manually stop charging using the CAN message, but this should absolutely be looked into further in the future time permitting

will-chaba:
@gtaharaedmonds I'd request your review but you're the PR author

boards/BMS/Src/App/App_Accumulator.c
```diff
@@ -306,7 +306,7 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
     float min_allowable_cell_temp = MIN_CELL_DISCHARGE_TEMP_DEGC;
 
     // if we are charging, max cell temp is 45C not 60C
-    if (App_TractiveSystem_GetCurrent(ts) < -3.0f)
+    if (App_TractiveSystem_GetCurrent(ts) > 3.0f)
```
KelmLelm, line 309: I'm not completely sure why the voltage we were checking for on the tractive system before was negative. Is it at all related to getting the ts voltage for charging finished?

boards/BMS/Src/App/App_TractiveSystem.c
```diff
@@ -48,7 +48,7 @@ float App_TractiveSystem_GetCurrent(struct TractiveSystem *ts)
     float low_res_current  = App_TractiveSystem_GetLowResCurrent(ts);
     float high_res_current = App_TractiveSystem_GetHighResCurrent(ts);
 
-    if (low_res_current < HIGH_RES_MAX_CURRENT_READING)
+    if (low_res_current < HIGH_RES_MAX_CURRENT_READING && low_res_current > -HIGH_RES_MAX_CURRENT_READING)
```
KelmLelm, line 51: I'm not too sure what's going on here. Under different circumstances, can the ts return negative values for current?

boards/BMS/Src/App/states/App_AllStates.c
```diff
@@ -140,19 +145,20 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
...
     App_CanTx_BMS_OkStatuses_BmsOk_Set(App_OkStatus_IsEnabled(bms_ok));
     App_CanTx_BMS_OkStatuses_ImdOk_Set(App_OkStatus_IsEnabled(imd_ok));
     App_CanTx_BMS_OkStatuses_BspdOk_Set(App_OkStatus_IsEnabled(bspd_ok));
 
-    const bool dcm_fault              = App_CanAlerts_BoardHasFault(DCM_ALERT_BOARD);
-    const bool fsm_fault              = App_CanAlerts_BoardHasFault(FSM_ALERT_BOARD);
-    const bool pdm_fault              = App_CanAlerts_BoardHasFault(PDM_ALERT_BOARD);
-    const bool dim_fault              = App_CanAlerts_BoardHasFault(DIM_ALERT_BOARD);
-    const bool fault_from_other_board = dcm_fault || fsm_fault || pdm_fault || dim_fault;
+    const bool dcm_fault = App_CanAlerts_BoardHasFault(DCM_ALERT_BOARD);
```
KelmLelm, line 156: Are there any faults we can have while charging that would be dangerous if not detected from other boards?

boards/BMS/Src/App/App_Accumulator.c
will-chaba: The TS current convention we want to use is that negative current is current leaving the pack (discharge), so charging current should be positive. I believe this was backwards before so that's why it had to switch

boards/BMS/Src/App/App_TractiveSystem.c
will-chaba: Yes. Depending on charging vs discharging

boards/BMS/Src/App/states/App_AllStates.c
will-chaba: Charging is a process that is handled solely by the BMS, no other boards are involved, so there shouldn't be any need to check for faults from other boards while charging specifically. (additionally charging is often done outside the car where the other boards aren't connected) Hence why I added a conditional operator on `fault_from_other_board` on line 161

@gtaharaedmonds did I get that right?

boards/BMS/Src/App/states/App_ChargeState.c
KelmLelm: Not sure if you changed the code from what's in the commented out block, but if the ts current is positive when charging, the negative sign in front of the App_Tract... would be incorrect. But if you have the check for if ts current is negative as well, then the charging_completed = true should never be reached so I'm not sure.

Also, unrelated I realized we didn't use the CURRENT_AT_MAX_CHARGE variable here and just put 0.85f.

boards/BMS/Src/App/states/App_AllStates.c
KelmLelm: That does make sense, if that's the case then this file should be good. I'll wait for some confirmation though

boards/BMS/Src/App/states/App_ChargeState.c
will-chaba: Yeah the commented code was altered during testing, the reason we weren’t using the macro was a troubleshooting step, I’ll clean it up.

As for the negative sign in front of ts_current, it was reading as positive when it should have been negative, because the previous convention was to have power coming in be positive and it hasn’t fully been switched over yet. Again, I’ll clean up the commented block to be as you left it

will-chaba (approved):
Worked on car, battery no go boom :fire:

boards/BMS/Src/App/states/App_AllStates.c
will-chaba: I confirmed with @Eddy13013 that other boards will always be disconnected when charging

KelmLelm:
> Worked on car, battery no go boom 🔥

SICKK

PR #894, opened 2023-03-21
New BMS reading from module

gtaharaedmonds:
### Summary
New BMS can read from a single module via the pseudo segment board, and can correctly measure voltages.

### Changelist 
- Lowered SPI bit rate to 750kbit/s
- Enabled pull-up resistor on the MISO line (required)
- Enabled pull-down resistors on other SPI lines to match old BMS (NSS, MOSI, SCK) (may or may not be required)
- Configured ADC, but haven't tested it yet

### Testing Done

New BMS can measure voltages on the pseudo segment board with a single module. Tested by removing a cell and applying 0.5-5V in its place, all voltages were measured correctly. Temperatures not tested.

### Checklist


will-chaba (approved):
Looks good to me

PR #896, opened 2023-03-25
Changes for motors spinning in the car

gtaharaedmonds:
### Summary
Misc. changes for motors spinning in the car

### Changelist 
- Configured BMS ADC module, however the ADC was not being triggered by TIM3 as it is on other boards so I had to start the conversions manually in the 1kHz task. Will have to look into why this is the case.
- Refactored some BMS IO code for voltage/current sensing
- Removed 1.5s delay on the startup of the DCM: The inverters fault if they do not see a torque request msg within 1s of being powered on, so DCM should talk on CAN immediately on power up.
- Enabled CAN RX interrupts on the DIM
- No longer test IO level code for current and voltage sense

### Testing Done

Spun motors on the car

### Checklist


boards/BMS/CMakeLists.txt
```diff
@@ -10,14 +10,11 @@ set(ARM_BINARY_IO_INCLUDE_DIRS
...
 
-set(X86_COMPATIBLE_IO_SRCS
-        "${CMAKE_CURRENT_SOURCE_DIR}/Src/Io/Io_VoltageSense.c"
-        "${CMAKE_CURRENT_SOURCE_DIR}/Src/Io/Io_CurrentSense.c")
 set(ARM_BINARY_X86_COMPATIBLE_SRCS
         ${ARM_BINARY_APP_SRCS}
         ${X86_COMPATIBLE_IO_SRCS})
 
-list(REMOVE_ITEM ARM_BINARY_IO_SRCS ${X86_COMPATIBLE_IO_SRCS})
+list(REMOVE_ITEM ARM_BINARY_IO_SRCS "")
```
will-chaba, line 17: Is this line necessary? There is no item passed to be removed from ARM_BINARY_IO_SRCS.

If this is autogenerated disregard, doesn't hurt anything

will-chaba, line 17: Is this line necessary? There is no item passed to be removed from ARM_BINARY_IO_SRCS.

If this is autogenerated disregard, doesn't hurt anything

boards/BMS/Test/Src/Test_VoltageSense.cpp
```diff
@@ -1,18 +0,0 @@
-#include "Test_Bms.h"
-
-extern "C"
-{
-#include "Io_VoltageSense.h"
-}
-
-TEST(VoltageSenseTest, tractive_system_voltage_calculation)
```
will-chaba, line 8: Would it be beneficial to add a test for voltage sense somewhere else now that it's in App code?

will-chaba (approved):
LGTM, couple of minor concerns

boards/BMS/CMakeLists.txt
gtaharaedmonds: This line is necessary now that Io_Voltage/CurrentSense files call the Io_Adc code directly
So can no longer test Io_Voltage/CurrentSense

will-chaba: Right, but per the `REMOVE_ITEM` syntax, this line isn't doing anything is it? is an item contained in `""`?

https://cmake.org/cmake/help/v3.0/command/list.html

boards/BMS/Test/Src/Test_VoltageSense.cpp
gtaharaedmonds: In theory yes, but all it would test is the resistor divider values
Since we'll have to test this on hardware anyway and we don't test IO code anywhere else I think it's fine

PR #899, opened 2023-03-31
SBG Ellipse N GPS/IMU sensor IO code

gtaharaedmonds:
### Summary
Added IO (and some basic app level code) to read from the SBG Ellipse N sensor and broadcast values over CAN. Currently read lateral accel, angular velocity, euler angles, timestamp, and status. 

STILL TODO: 
- Look at how much time the CPU takes decoding these logs
- Add some way to determine the board is still receiving valid data from the sensor
- Better status checking
- Export an official sbgCenter config file into the repo
- Stop clang-formatting third party code since this is just a waste of time

Since this PR is already 5k lines, I want to leave these for another PR.

### Changelist 
- Added Io_EllipseImu.h/.c which reads data from the sensor via UART DMA. Uses SBG's provided library to parse the binary logs.
- Added App_EllipseImu.h/.c which is (currently) a barebones wrapper for the IO code. Broadcasts data over CAN.
- Integrated SBG's provided library into `boards/shared/third_party/sbgECom`. 
- Added a simple ring queue implementation (FIFO queue which requires no dynamic memory, AKA a circular buffer) used by the IO code for storing UART packets.
- Updated DCM task stack sizes to 4096 vs 512 words. Was getting errors related to stack corruption in 100Hz task, I suspect related to stack overflow since changing to 4096 fixed the problem. We should look into if 4096 is excessive, and if the other tasks can stay at 512.

WORTH REVIEWING: 
- IO code: Io_EllipseImu.h/.c
- Ring buffer implementation: App_RingQueue.h/.c

### Testing Done

Tested with the sensor, and added a few unit tests for the ring queue.

### Checklist


boards/shared/Src/App/App_RingQueue.c
```diff
@@ -0,0 +1,72 @@
...
+    // Insert the new value
+    queue->elements[queue->tail] = value;
+
+    // The queue has overflowed if the tail idx has "circled around" and caught up to the head idx,
+    // so increment the head idx. Also check there are elements in queue so this is not set for
+    // an empty queue (which has tail and head idx equal).
+    const bool queue_overflowed = (queue->tail == queue->head) && queue->count > 0;
+    if (queue_overflowed)
+    {
+        // Count doesn't if overflowed, since new element replaced an old element
```
will-chaba, line 36:
```suggestion
        // Count doesn't change if overflowed, since a new element replaced an old element
```

will-chaba (approved):
LGTM, like the comments!

PR #900, opened 2023-04-01
All Boards Update Heartbeat

Lucien950:
### Summary
- Change all boards CAN Vitals cycle time to 100ms
- Make boards go into fault state if heartbeat is not detected from other boards
- On every 100Hz
    - broadcast heartbeat vitals to be true to all boards
    - for each board it is dependent on
        - check in if the CAN vital value is true
        - set the CAN vital value to false
    - tick function (rename?)
    - if heartbeat status not ok, send to fault state, broadcast warning on CAN bus

### Testing Done
- 

### Checklist


gtaharaedmonds (approved):
LGTM

PR #903, opened 2023-04-04
Add EEPROM Code to BMS

will-chaba:
### Summary

Add I/O code for reading and writing an array of bytes and floats to EEPROM

### Changelist 

Adds file Io_Eeprom.c with functions:

Io_Eeprom_WriteByte
Io_Eeprom_ReadByte
Io_Eeprom_PageErase

Adds file App_Eeprom.c with functions

App_Eeprom_WriteFloats
App_Eeprom_ReadFloats

Adds EEPROM tests in
Test_Eeprom.cpp

### Testing Done

Tested on BMS-Core to ensure data can be written to and successfully read back from EEPROM.

Additionally, float writing/reading conversions were tested in Test_EEPROM.cpp

Encountered a weird bug where if parameters to the read function were changed (such as size, page, offset etc...) and the board re-flashed, the function would not properly read from the EEPROM until a power cycle is performed on the board. The issue was found to be with a 'locked' I2C bus caused by interrupting an in-progress transaction. The solution is to cycle the clk signal to the EEPROM 10 times on startup in case it was in the middle of an operation prior to flashing. This requires switching the pins to GPIO, manually pushing the clock then switching the pins back to I2C functionality, this is done in the main. c's startup region.

An additional bug was found when trying to write to multiple pages, which the EEPROM requires a 5ms write delay between page writes. Unfortunately, we ran into some issues with using the HAL_DELAY function to achieve this functionality, so the solution, for now, is to limit writes to once per task cycle. Given we can write 4 floats on a single page write, this should be sufficient for SOC purposes.

A wiki page was created to document these findings, in case they are investigated in the future: https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/123666441/EEPROM+Firmware+Notes

### Checklist
- [ x] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).


boards/BMS/Src/Io/Io_Eeprom.c
```diff
@@ -0,0 +1,154 @@
...
+
+        // Determine the number of bytes to write to page
+        uint16_t bytes_to_write = Bytes_To_Process(size, offset);
+
+        HAL_I2C_Mem_Write(EEPROM_I2C, EEPROM_ADDR, mem_address, 2, &data[pos], bytes_to_write, 1000);
+
+        startPage++;                                // Increment the page
+        offset = 0;                                 // writing to new page, offset will be zero for next write.
+        size   = (uint16_t)(size - bytes_to_write); // reduce size by the amount of bytes written to page
+        pos    = (uint16_t)(pos + bytes_to_write);  // increase starting position by number of bytes written
```
kironsen, line 89: tried to follow how this is iterating as you're accessing `data[] = bytes_to_write[]` with size `4` at position `pos`, but you appear to be incrementing `pos` beyond size `4` (buffer overread?). some recommendations to improve readability here:

replace `convert_float_to_bytes()` with:
```c
float f = ...;
const uint8_t const * data = (const uint8_t const *)&f; // const pointer to const int
```

simplify `convert_bytes_to_float()` with:
```c
memcpy(&f, data, sizeof(f));
```

this should allow you to remove the following functions entirely: `Io_Eeprom_WriteFloat()`, `Io_Eeprom_ReadFloat()`, `convert_float_to_bytes()`, `convert_bytes_to_float()`

P.S.
- define magic numbers like `4` in one place to improve maintainability
- remember to scope local functions with `static` and not expose them to the rest of the program

will-chaba: Definitely some good recommendations here, I'll be sure to apply them

As for the flow, I'll lay it out best I can and that should either clarify or exposes flaws in my design:

1. `data[]` is an array of bytes to be written to the EEPROM, it can be of any size (theoretically I suppose it should be limited to at least the size of the EEPROM)
2. The EEPROM can only be written to one page at a time, each page being 16 bits wide (4 bytes), `num_pages` determines the number of individual page writes necessary to write contents of `data[]` to the EEPROM, and is dependant on the starting `page`, `offset` as well as the total number of bytes to be written, `size`
3. `Bytes_To_Process` returns the number of bytes to be read or written to the current page, in this case `bytes_to_write` contains the number of bytes to write to the current page.
4. ` HAL_I2C_Mem_Write` writes `bytes_to_write` number of bytes to the current page, starting at `mem_address`
5. `pos` (the index of the input data array) is incremented by the number of bytes written (between 1-4)
6. for loop continues until we have written to the necessary number of pages to write entirety of `data[]` to the EEPROM

I think that's the important bits at least... obviously I could do better to clean up and comment to make this more clear. It should be noted that this has not yet been tested, and so there very well could be bugs in my code. Hopefully this clears up at least a little of the confusion as to my intentions at least

kironsen: thanks for the explanation 👍. i may be misunderstanding but my main concern was passing `bytes_to_write[]` of size 4 to `Io_Eeprom_Write()` as `data[]` and accessing that at `pos` which is being incremented beyond size 4 (?)

```c
void Io_Eeprom_WriteFloat(uint16_t page, uint16_t offset, float data)
...
    uint8_t bytes_to_write[4];
...
    Io_Eeprom_Write(page, offset, bytes_to_write, 4);
...
void Io_Eeprom_Write(uint16_t page, uint16_t offset, uint8_t *data, uint16_t size)
...
        HAL_I2C_Mem_Read(EEPROM_I2C, EEPROM_ADDR, mem_address, 2, &data[pos], bytes_to_read, 1000);
...
        pos    = (uint16_t)(pos + bytes_to_read);  // increase starting position by number of bytes read
```

if i'm misunderstanding, a check that `pos < size` would still help. anyways probs only worth refactoring after clean up + testing

gtaharaedmonds: You can use `sizeof(float)` too instead of 4 since that indicates what 4 represents in this case

will-chaba: Oh I see the confusion, you're referencing `Io_Eeprom_WriteFloat` `bytes_to_write`, yeah that's a terrible variable name, my bad lol.

As for `pos`, in the case you have above, `Io_Eeprom_WriteFloat` writes a single 4-byte float value to the EEPROM, because this would fit inside a single page (16 bytes) the for loop inside `Io_Eeprom_Write` would only run for a single cycle, so while `pos` is updated, the updated value is never used, as `num_pages` is 1 and the for loop will break. 

if `num_pages` was larger than 1, (say writing 20 bytes to the EEPROM), then the updated value of `pos` would indicate the starting index in `data` to write to the second page.

So yes, `pos` does go greater than `size`, but the for loop will break under these conditions. An extra sanity check could be added, but I don't think it would ever be exercised.

boards/BMS/Inc/App/App_EepromExitCode.h
```diff
@@ -0,0 +1,13 @@
...
+ * @brief  EEPROM Status structures definition
+ */
+typedef enum
+{
+    EEPROM_OK           = 0x00U,
+    EEPROM_ADDR_ERROR   = 0x01U,
+    EEPROM_I2C_ERROR    = 0x02U,
+    EEPROM_SIZE_ERROR   = 0x03U,
+    EEPROM_DATA_CORRUPT = 0x04U
+} EEPROM_StatusTypeDef;
```
will-chaba, line 13: Not a huge fan of having this in it's own file, but since I wanted it to be used in IO, APP, and Test files, this seemed the easiest way. Put in App header directory similar to App_SharedExitCode.h

boards/BMS/Test/Src/Test_Eeprom.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    // capture converted byte array
+    for (int i = 0; i < PAGE_SIZE; i++)
+    {
+        byte_values[i] = write_page_fake.arg2_val[i];
+    }
+
+    // Unfortunately, cannot easily access the bytes converted back to floats due to FFF limitations
+    // Because function used is static and located in App_Eeprom.c, need to make a copy of that static function in this
+    // file This is NOT maintainable, any changes to convert_bytes_to_float must be added here as well.
```
will-chaba, line 80: For this test, I wanted to write a floats and read them back to test that they are being converted to bytes correctly. With our current test/mock setup, there is no way for me to access the byte-array in App_Eeprom_ReadFloats. As such the next best thing is to test the function that converts bytes into floats directly, problem being that I have it implemented as a static function that is inaccessible to the test file. So the terrible bandaid solution was to copy the entire static function and have a copy of it word-for-word in the test file. This works and allows me to test it, but if there is ever a need to modify this function, it would need to be updated in the test file as well. 

Frankly, I don't know if there is a better way of doing this, besides making the conversation function no longer static, (I just don't know where I'd put it if I did that). I just wanted to write more tests, as I feel they are extremely valuable, but to say that I had to do some jerry-rigging to be able to write tests for these functions at all is an understatement.

tldr. static function not visible to test file. copied it here. it's ghetto, but at least it's tested.

kironsen: you guys should use some define magic to fix this:
```c
#if TEST
#define STATIC
#else
#define STATIC static
#endif // TEST
```

then define your function as:
```c
STATIC float convert_bytes_to_float...
```

boards/BMS/Src/App/App_Eeprom.c
```diff
@@ -0,0 +1,207 @@
...
+#define SOC_ADDR_PAGE (0U)
+#define SOC_ADDR_OFFSET (0U)
+#define SOC_ADDR_SIZE_BYTES (1U)
+#define NUM_PAGES (128U)
+#define SAVED_COPIES 3U
+#define BYTES_PER_FLOAT sizeof(float) / sizeof(uint8_t)
+#define BYTES_PER_SHORT sizeof(uint16_t) / sizeof(uint8_t)
+#define SHORT_SIZE ((uint8_t)sizeof(uint16_t))
+
+static void convert_float_to_bytes(uint8_t *bytes, float float_to_convert)
```
kironsen, line 12: i think that all of these conversion functions are a bit overcomplicated. also if i remember correctly, accessing unions like this isn't portable due to endianness and you're lucky that ARM and x86 are both little endian

as per https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/903#discussion_r1159114345:

you can replace convert_float_to_bytes() with:
```c
float f = ...;
const uint8_t const * bytes = (const uint8_t const *)&f; // const pointer to const int
```

and simplify convert_bytes_to_float() with:
```c
memcpy(&f, bytes, sizeof(f));
```

you can keep them in functions if you want to keep the tests

boards/BMS/Test/Src/Test_Eeprom.cpp
will-chaba: I tried something like this but ran into a couple issues (maybe you can clear some light on this)
1. even if I did this, wouldn't I still need to put function definitions in the header? I guess header definitions could be stuck within a #ifdef block

2. I'd have to figure out where to define TEST, as the test files are compiled last, so defining it at the start of a test before including App_Eeprom.h doesn't work. 

Another fix I saw was to include the .c file rather than the header in the test file, however this would re-compile anything included in the .c file, increasing executable size and compile time, but would include the static functions. This would require messing with CMake, as doing so right now results in the compiler not being able to find the .c file.

kironsen: 1. yep, but that's fine. can add a separate section of function declarations with a comment
2. something like that may already be defined in the highest level CMakeLists.txt, or you can define it yourself

> Another fix I saw was to include the .c file rather than the header in the test file
seems really messy, the define magic is pretty industry standard afaik

boards/BMS/Src/App/App_Eeprom.c
will-chaba: ah good point, I overlooked this in your previous comment. I'll look into switching to this. tbh the static functions are causing just as much as a testing headache as just using these inline, might be worth abandoning.

boards/BMS/Src/App/App_Eeprom.c
```diff
@@ -0,0 +1,207 @@
...
+EEPROM_StatusTypeDef
+    App_Eeprom_WriteFloats(struct Eeprom *eeprom, uint16_t page, uint8_t offset, float *input_data, uint8_t num_floats)
+{
+    // Check if data to write fits into single page
+    if (num_floats * sizeof(float) > PAGE_SIZE)
+    {
+        return EEPROM_SIZE_ERROR;
+    }
+    // for floats, offset should be aligned to multiple of 4
+    else if (offset % 4 != 0)
```
gtaharaedmonds, line 125: nit: 4 could be BYTES_PER_FLOAT or sizeof(float)

boards/BMS/Src/App/App_Eeprom.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+EEPROM_StatusTypeDef App_Eeprom_PageErase(struct Eeprom *eeprom, uint16_t page)
+{
+    return eeprom->page_erase(page);
+}
+
+EEPROM_StatusTypeDef App_Eeprom_WriteAddress(struct Eeprom *eeprom, uint16_t page, uint16_t address)
+{
+    uint16_t num_bytes = SAVED_COPIES * 2; // saving 3 copies of address, each 2 bytes
```
gtaharaedmonds, line 175: nit: 2 could be BYTES_PER_SHORT or sizeof(uint16_t)

boards/BMS/Src/App/App_Eeprom.c
```diff
@@ -0,0 +1,207 @@
...
+    uint16_t short_array[SAVED_COPIES];
+
+    for (uint8_t i = 0; i < SAVED_COPIES; i++)
+    {
+        short_array[i] = convert_bytes_to_short(&byte_array[i * 2]);
+    }
+
+    // TODO: Implement error checking using 3 copies of address stored in short_array
+    *address = short_array[0];
+
```
gtaharaedmonds, line 205: I think it makes sense to do all the voting in the same file: Since it looks like you're doing voting in the upcoming SOC file for the floats, might as well do the address voting there too. So consider making this a "ReadAddresses" function like the "ReadFloats" above.

gtaharaedmonds (commented):
looks great. +1 on either making a static macro (as long as this compiles, is a static function in a header file allowed?) or making an inline function in the header. after that ill approve

gtaharaedmonds (approved):

boards/BMS/Inc/App/App_Eeprom.h
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Write an address 3 times to EEPROM
+ * @note SHOULD ONLY BE CALLED ONCE EVERY 5ms, DOING SO MORE QUICKLY WILL VIOLATE EEPROM SPECS
+ * @param eeprom Eeprom to erase a page from
+ * @param page the number of the start page to set to all zeroes
+ * @param addresses array of 3 addresses to be chosen from. Should all contain same value but may not due to
+ * data-corruption in EEPROM
+ * @return EEPROM_StatusTypeDef returns success status for debug
+ */
```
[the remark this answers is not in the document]
will-chaba: oops.

PR #911, opened 2023-05-01
Add precharge tests

gtaharaedmonds:
### Summary
Add precharge tests before testing on hardware.

### Changelist 
Added a single test for precharging to Test_StateMachine.cpp, with start conditions (TS voltage < 10V, AIR- closes) and precharge duration as parameters. Check state transitions and AIR+ close not requested until a successful precharge. Tests for precharge faults are in Test_Faults.cpp.

### Testing Done

PR only changes tests so unit tests are sufficient

### Checklist


boards/BMS/Inc/App/App_PreChargeRelay.h
```diff
@@ -6,6 +6,20 @@
...
+#define PRECHARGE_CAPACITANCE_F (INV_CAPACITANCE_F * NUM_OF_INVERTERS)
+#define PRECHARGE_RC_MS         (S_TO_MS(PRECHARGE_RESISTOR_OHMS * PRECHARGE_CAPACITANCE_F))
+
+// 2.7RC corresponds to time to reach ~93% charged
+#define PRECHARGE_COMPLETION_MS          ((float)PRECHARGE_RC_MS * 2.7f)
+#define PRECHARGE_COMPLETION_LOWER_BOUND ((uint32_t)(PRECHARGE_COMPLETION_MS * 0.5f))
+#define PRECHARGE_COMPLETION_UPPER_BOUND ((uint32_t)(PRECHARGE_COMPLETION_MS * 2.0f))
+
+// clang-format on
+
```
gtaharaedmonds, line 22: Moved this here so it can be accessed in the tests

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -279,30 +276,26 @@ class BmsStateMachineTest : public BaseStateMachineTest
     struct Clock *            clock;
 };
 
-// BMS-31
 TEST_F(BmsStateMachineTest, check_init_state_is_broadcasted_over_can)
 {
     SetInitialState(App_GetInitState());
     EXPECT_EQ(BMS_INIT_STATE, App_CanTx_BMS_Vitals_CurrentState_Get());
 }
```
gtaharaedmonds, line 283: We no longer track requirements so removed all the comments from Test_StateMachine.cpp

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -648,4 +628,117 @@ TEST_F(BmsStateMachineTest, check_remains_in_fault_state_until_fault_cleared_the
...
+TEST_F(BmsStateMachineTest, check_precharge_state_transitions_and_air_plus_status)
+{
+    struct
+    {
+        bool  air_negative_closes;
+        float initial_ts_voltage;
+        float precharge_duration;
+        bool  expect_precharge_starts;
+        bool  expect_precharge_successful;
+    } test_params[5] = { {
```
will-chaba, line 641: Although I like the conciseness of using this struct to avoid writing 5 tests, it might make debugging slightly more difficult if a test fails in the future, as the test framework will point to "check_precharge_state_transitions_and_air_plus_status", but won't indicate which step failed without manual debugging. I can live with it as is, but an FYI for future tests

will-chaba (approved):
Seems fine, have concerns about using test_params struct for debug message clarity reasons, otherwise looks good

boards/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -167,10 +168,6 @@ class BmsStateMachineTest : public BaseStateMachineTest
         RESET_FAKE(get_low_res_current);
         RESET_FAKE(get_high_res_current);
 
-        // The charger is connected to prevent other tests from entering the
-        // fault state from the charge state
-        is_charger_connected_fake.return_val = true;
-
         // Set initial voltages to nominal value
```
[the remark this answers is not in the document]
gtaharaedmonds: Looks like its currently added as part of the setup, and added on a test-by-test basis as well. So that's why I can remove it here.

boards/BMS/Test/Src/Test_StateMachine.cpp
gtaharaedmonds: Yeah this is a good point, seems like a bit of a limitation of gtest that there's no easy way to do stuff like this
I don't love repeating the test 5 times since then we get massive testing files, and if we have to change the test down the line then we have to modify it in 5 places
But you're right there's probably a better way, like breaking it into a few test cases (all the reasons precharge will/won't start, slow precharge, fast precharge) and then do some smaller parameterizing afterwards 
But I'll leave it as-is for this PR

gtaharaedmonds: Would also be nice maybe to centralize the main testing fixture so it's really easy to add seperate test files, and then we can get add lots of cases without getting mega files

PR #918, opened 2023-05-27
Prefix signal names with message name in generated DBC

gtaharaedmonds:
### Summary
In the DBC generated by jsoncan, signal names are now prefixed by their message name. For example, what was previously the signal `TsVoltage` in the message `BMS_TractiveSystem` is now the signal `BMS_TractiveSystem_TsVoltage` in the message `BMS_TractiveSystem`. This has no effect on any firmware, the signals will just show up differently in PCAN and when uploaded to the database. 

The purpose of this PR is to fix name collisions caused by signals on different boards and/or different messages having the same name. This is really just a workaround - signal names are now too long when viewing, but I don't want to diff any firmware.

Actual code change is in the jsoncan repo, this just updates the submodule.

### Testing Done
None. This doesn't diff any embedded code so fine without testing.

### Checklist


gtaharaedmonds:
cleaning up old stuff

PR #920, opened 2023-06-01
Update TS current fault check

will-chaba:
### Summary

Update the polarity of TS current fault check.

### Changelist 

Updated TS current fault check to reflect polarity change (discharge current is negative, charge current is positive)

Also changed low-res/high-res current check condition to use IS_IN_RANGE macro for easier reading

### Checklist


gtaharaedmonds (approved):
LGTM

PR #921, opened 2023-06-02
Add Odometer to BMS (using EEPROM)

will-chaba:
### Summary

Implement Odometer that calculates distance travelled and saves it periodically to the EEPROM.

Function takes in wheel speed from FSM, takes average of both wheels (maybe should take max of two wheels instead?), and uses the circumfrance of the tires to calculate distance travelled.

Saves to EEPROM once every 30 seconds so we don't need to deal with wear levelling for now (could be a longer interval, this isn't a mission-critical statistic in the case of a sudden shutdown)

### Changelist 

Added App_Odometer.c and App_Odometer.h, Added function in App_Eeprom.c to read/write 4 floats with error checking (reset to 0 if error detected, this makes sense for odometer but might not in the future for other uses)

### Checklist

### DIFF
--- a/boards/BMS/Inc/App/App_BmsWorld.h
+++ b/boards/BMS/Inc/App/App_BmsWorld.h
@@ -14,6 +14,7 @@
 #include "App_TractiveSystem.h"
 #include "App_SharedClock.h"
 #include "App_Eeprom.h"
+#include "App_Odometer.h"
 
 struct BmsWorld;
 
@@ -39,7 +40,8 @@ struct BmsWorld *App_BmsWorld_Create(
     struct PrechargeRelay *  precharge_relay,
     struct TractiveSystem *  tractive_system,
     struct Clock *           clock,
-    struct Eeprom *          eeprom);
+    struct Eeprom *          eeprom,
+    struct Odometer *        odometer);
 
 /**
  * Deallocate the memory used by the given world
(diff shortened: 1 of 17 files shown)


boards/BMS/Inc/App/App_BmsWorld.h
```diff
@@ -39,7 +40,8 @@ struct BmsWorld *App_BmsWorld_Create(
     struct PrechargeRelay *  precharge_relay,
     struct TractiveSystem *  tractive_system,
     struct Clock *           clock,
-    struct Eeprom *          eeprom);
+    struct Eeprom *          eeprom,
+    struct Odometer *        odometer);
```
lehroatic, line 44:
```suggestion
    struct Odometer *        odometer); // ur dumb lol
```

boards/BMS/Src/App/App_Eeprom.c
```diff
@@ -144,3 +143,59 @@ EEPROM_StatusTypeDef App_Eeprom_Read4CopiesOfAddresses(struct Eeprom *eeprom, ui
...
+    EEPROM_StatusTypeDef read_status =
+        App_Eeprom_ReadFloats(eeprom, address, DEFAULT_OFFSET, floats_read, SAVED_COPIES);
+
+    // If read unsucessful, reset to 0
+    if (read_status != EEPROM_OK)
+    {
+        return 0.0f;
+    }
+    // If any two read values are the same, return the shared value
+    else if (floats_read[0] == floats_read[1])
```
kironsen, line 160: the entire block of if statements could be simplied by a double for loop with `j` starting one past `i`. not sure if that's less clear tho
```suggestion
    for (i = 0; i <= 3; i++)
    {
        for (j = i + 1; j <= 3; j++)
        {
            if (floats_read[i] == floats_read[j]) return floats_read[i];
        }
    }
```

will-chaba:
No longer needed

PR #924, opened 2023-06-05
Lehan/rotary switch

lehroatic:
### Summary 
The rotary switch has a conversion from the drive mode 4-bit binary to an integer position value, with additional checks if the position is valid (in the range of 1-12). The applications of this rotary switch are endless. We can use it for torque vectoring states or "drive modes". Different drive modes can ideally be used to allow the car to perform differently in different events during the competition.

### Changelist 

Io_DriveModeSwitch.c/.h, App_RotarySwitch.c, DIM_tx.json, and App_DriveState.c are the main files that were changed during this project. The Io code was updated to convert the 4-bit binary signal to an integer that ranges from 1-12. The app code for the rotary switch itself had a boolean signal to check if the position was a valid number. App drive state is where CAN signals are used for the position and is_valid checks, and ideally where we can have multiple drive states.
### Testing Done

As the recent commit states, this feature currently works with 8/12 functionality. There was testing code written in the IO code to check if the 4-bit binary number was working as expected:

`bool pins_set[4];`
`pins_set[0] = HAL_GPIO_ReadPin(DRIVE_MODE_0_GPIO_Port, DRIVE_MODE_0_Pin) == GPIO_PIN_RESET;`
`pins_set[1] = HAL_GPIO_ReadPin(DRIVE_MODE_1_GPIO_Port, DRIVE_MODE_1_Pin) == GPIO_PIN_RESET;`
`pins_set[2] = HAL_GPIO_ReadPin(DRIVE_MODE_2_GPIO_Port, DRIVE_MODE_2_Pin) == GPIO_PIN_RESET;`
`pins_set[3] = HAL_GPIO_ReadPin(DRIVE_MODE_3_GPIO_Port, DRIVE_MODE_3_Pin) == GPIO_PIN_RESET;`
`UNUSED(pins_set);`

Again, this code checks if the GPIO pins are on or off depending on which position the rotary switch is in. For some reason 4/12 of the expected position values show up as 0 and this code also equates all the pins as 0 during this section. These null positions are integer values 4-6.

### Checklist
- [ ] If this pull request is longer than **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).


boards/DIM/Src/App/App_RotarySwitch.c
```diff
@@ -41,3 +42,14 @@ ExitCode App_RotarySwitch_GetSwitchPosition(
 
     return exit_code;
 }
+
+bool App_RotarySwitch_IsValid(const struct RotarySwitch *const rotary_switch)
+{
+    uint32_t position = rotary_switch->get_switch_position();
```
kironsen, line 48: nit - could make the function a bit more readable by simplifying
```suggestion
    return (rotary_switch->get_switch_position() < rotary_switch->num_switch_positions);
```

gtaharaedmonds:
I don't believe this is still in scope so closing for now. Re-open if I'm wrong.

PR #925, opened 2023-06-05
Demote Pedal disagreement to warning

will-chaba:
### Summary

Demote Pedal disagreement condition from fault to warning in FSM

Triggering a fault will cause the BMS to fault, according to rules, the only thing that needs to occur under these conditions is a toque request of zero.

See EV.5.5, EV.5.7

### Checklist


can-bus/json/FSM/FSM_alerts.json
forbesashli, line 1: looks gud

forbesashli (approved):
looks

PR #928, opened 2023-07-08
Dockerize CI

gtaharaedmonds:
### Summary
Switch CI pipeline to run inside a Docker container. This gives some speed improvements: Averaging over 5 CI runs I saw an improvement of 2m55s, down from 4m8s. Also I think Docker containers are nicer to work with compared to setting up and caching everything in the Github Actions workflow `.yml`, which we do currently. This will be useful since I expect we'll add a few more jobs this year.

### Changelist 
1. Added `Dockerfile` and README file to `environment` directory.
2. Moved environment setup scripts from `scripts/environment_setup` to `environment/scripts`. Hopefully this should improve organization: I want everything related to repo setup to live in the `environment` directory, and everything required by the Docker container.
3. Updated the `.github/workflows/ci.yml` to use the new container. Should be identical to our current CI pipeline.
4. Deleted deprecated `.travis.yml` and `.series_ci.sh` (ancient past CI solutions, before my time)
5. Updated ARM compiler to Version 10.3-2021.10 (latest version).

### Testing Done
CI runs and passes.

### Checklist


.github/workflows/ci.yml
```diff
@@ -0,0 +1,167 @@
...
+
+      # Uploads .elf binary to Github Actions workflow
+      - name: Save built artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.board }}.elf
+          path: ${{ env.BUILD_DIR }}/${{ matrix.board }}/${{ matrix.board }}.elf
+  
+  test:
+    name: Build and Run Tests
```
will-chaba, line 131: I like that we're separating out the x86 tests per board (easier to debug) but I think you missed the shared tests?

will-chaba (changes requested):
Overall looks great! Really simplifies dependency setup for each check! But I think we're no longer running the shared tests, please check

.github/workflows/ci.yml
gtaharaedmonds: yup, added them in now
also added memcheck in for the tests which I forgot about before

will-chaba (approved):

PR #929, opened 2023-07-10
Speedup clang-format

gtaharaedmonds:
### Summary
Used python's multiprocessing package to run clang-formatting in parallel, speeding it up a lot. I'm seeing <3s total lint time, down from ~18s.

### Changelist 
Only modified `clang-format/fix_formatting.py`. Also refactored the file a lot.

### Testing Done
Works locally (Windows) and remotely (Linux). Can't test myself on local Linux or Mac, but I haven't changed any of the command invocation so I think it'll probably still work.

### Checklist


will-chaba:
Joke's on you it runs in ~3s on Linux to start with :laughing:

In all seriousness though looks good, but with the changes to platform detection we should try this on Linux (does anyone use mac on the team?) before merging

gtaharaedmonds:
> Joke's on you it runs in ~3s on Linux to start with 😆
> 
> In all seriousness though looks good, but with the changes to platform detection we should try this on Linux (does anyone use mac on the team?) before merging

That means it'll now run in 0.5s :)
Runs on windows (currently untested on Linux, but CI should handle that). No one's built on mac in years so I'm not worried about this change specifically, I'm sure there will be bigger changes that pop up next time someone does. For what its worth I tried not to change the mac invocation so if it worked before, it should work now.

will-chaba (approved):

PR #931, opened 2023-07-17
Refactor DIM to test out proposed refactor plan

gtaharaedmonds:
### Summary
Testing out refactoring plan ideas on DIM.

#### Directory structure
1. Put all of STM32's autogenerated stuff into `stm32` directory, including the `.ioc`.
2. Add source directories: `app`, `device`, `driver`, and `driver_shared`. **Note**: `driver_shared` will actually live in the `shared` directory, and just be called `driver`, but I put it here for now to hopefully reduce confusion.
3. `.h` and `.c` files now live in the same directory.
4. Also pulled out task tick functions and initialization to a `tasks.h/c` file in the `driver` level. Reduces the clutter of developing stuff in `main.c`. 

#### Coding conventions
1. The `app`, `device`, and `driver` directories each represent an abstraction layer. `app` for high-level behavior, `device` for wrappers around physical devices, chips, etc., and `driver` for hardware peripherals. 
2. Removed abstraction layer prefix. **Note**: Please let me know what you guys think of this!
3. Enums and structs are `PascalCase`, and all are `typedef`ed.
5. Functions are now `camelCase` instead of `PascalCase`. I think this looks better and its easier to distinguish between enums/structs.
6. Allocate data statically in modules, for modules that only will ever have 1 instance.

#### New code
1. Added GPIO and ADC shared drivers, which will eventually live in `shared` and be used across all boards.
2. Changed the "world" to "globals" and simplified it to be an `extern`ed global variable. I think globals is a better name since it no longer encapsulated all of a board's hypothetical "world", but is more just a container for all global state.
3. General refactoring and simplification of existing code.

#### What I haven't done yet

1. Any CMake changes (this code will absolutely not compile!)
2. Refactor any shared code(with the exception of adding the shared GPIO and ADC drivers).
3. Any documentation, the new functions have no comments. But I think its clear enough for now?
4. Any updates to testing.
5. Any changes to RTOS tasks.

These are both huge projects so I want feedback on the plan before actually putting in the effort to change anything. Keep in mind that this doesn't compile, and probably has lots of bugs.

#### What to review

1. General directory structure and coding conventions.
2. Global state plan (renaming "world" to "globals" and making it a global variable).
3. How stuff gets initialized in `main.c`.
7. How stuff gets called in `driveState.c`.
8. Any of the new drivers and devices.


will-chaba:
Had a look, definitely an improvement though admittedly struggled to break some old habits, so had a couple of concerns:

- I'm not sure I'm sold on removing the abstraction layer prefix, especially given that we're adding a new abstraction layer, the prefix makes it easy for someone to find the definition (even if they don't know how to use the IDE to go to definition)
- For no particular reason I'm bothered that there is no 'src' folder, I know it's not needed and 'app' is plenty descriptive but it just bothers me
- Why a dedicated 'stm32' folder? I know it encapsulates autogenerated code, but it makes it unintuitive to find where `main.c` is. 
- I like the idea of `globals`, but could we prefix the globals struct with the board name? makes it more clear that it is all the global variables for a given board
- The mix of `camelCase` and `underscore_case` we have going right now seems odd, I understand the separation between the file name and function (`fileName_function`) but having the function be lowercase after having an uppercase word in the file name seems un-elegant. Not sure what standard practice is, but what if we upper-cased the word after the underscore, something like `fileName_Function`? Curious about what everyone else thinks, I guess doing so would be a weird combination of `camelCase`, `underscore_case`, and `PascalCase`.

Unrelated, would it be possible to move the top-level CMakeLists.txt to the repo root? Not strictly required if it's a headache, but right now any files outside the 'boards' directory are not immediately displayed in the project tab of CLion, and it would be more intuitive to new users if they could see the entire repo I think (also maybe rename `boards` to `src` in this case? just an idea. (not sure if that'd be confusing due to having src code for other things in our repo now or in the future (especially if SW team starts using the repo)

gtaharaedmonds:
> * having .c and .h in the same folder has its advantages, but I am worried about more complicated boards with more files, trying to read through the `app` folder hurt my brain a little for some reason, it was like stripes of files with the same name rather than a list. (I mostly work with rtl code at work which doesn't use headers, so if this is common and I'm just not used to it ignore my concern)

I think both same directory and separate are common (I've seen both in past jobs). I think having them in the same place is nice since you so often switch between a module's `.h` and `.c`, but agreed it takes some getting used to.

> * I'm not sure I'm sold on removing the abstraction layer prefix, especially given that we're adding a new abstraction layer, the prefix makes it easy for someone to find the definition (even if they don't know how to use the IDE to go to definition)

Yeah this is a fair concern I think this would take the most getting used to. Curious to see what other people think, and I'm not tied to either way. These prefixes could work:
1. Application: `app`
2. Device: `dev`
3. Driver: `drv`, or Hardware: `hw`

> * Why a dedicated 'stm32' folder? I know it encapsulates autogenerated code, but it makes it unintuitive to find where `main.c` is.

The reason is to hide all auto-generated code. I think having that at the top-level is intimidating for new members, is ugly since it uses different conventions, and is also just not useful to have to sort through 99% of the time. Hiding `main.c` is indeed a downside, but I think this is fixable by just documenting where the new `main.c` is located. Also, hoping to pull out all the meat and potatoes out of main to `tasks.h/.c` at the driver level so we don't need to have to develop inside of the auto-generated clutter in the main file.

> * The mix of `camelCase` and `underscore_case` we have going right now seems odd, I understand the separation between the file name and function (`fileName_function`) but having the function be lowercase after having an uppercase word in the file name seems un-elegant. Not sure what standard practice is, but what if we upper-cased the word after the underscore, something like `fileName_Function`? Curious about what everyone else thinks, I guess doing so would be a weird combination of `camelCase`, `underscore_case`, and `PascalCase`.

Also curious to see what other people think. What I've done in this PR is standard `camelCase`, i.e. `prefix_moduleName_coolFunctionName`. Agreed it's a bit of a weird convention to use all 3.
Imo `underscore_case` is by far the most readable and I'd prefer using that for everything, but it gets tricky since we do prefix + module name. Open to any suggestions though.

> Unrelated, would it be possible to move the top-level CMakeLists.txt to the repo root? Not strictly required if it's a headache, but right now any files outside the 'boards' directory are not immediately displayed in the project tab of CLion, and it would be more intuitive to new users if they could see the entire repo I think

Yes! This is a pretty easy change, we should definitely do this. 

These are good suggestions: Let's wait to hear what other people have to say. Happy to make some changes.

gtaharaedmonds:
This was just a demo, and I think has served its purpose, so closing in the effort of cleaning up outdated PRs.

PR #937, opened 2023-07-28
Reorganize repo directories

gtaharaedmonds:
### Summary
Reduce clutter and improve repo organization for Quadruna.

**Recommend checking out this branch locally and looking around, GitHub is very slow for big PRs like this**

### Changelist 
1. Moved `clang-format` code into the scripts directory.
2. Moved Cascadia PM100DZ inverter config tool into `tools` directory
3. Deleted all the old tutorials. Pretty much no information there and we use Confluence for that now anyway. Moved PCAN setup to `tools`.
4. Deleted `specs.md` since we no longer track specs in the repo
5. Renamed `boards` to `firmware`. This is because there will soon be a lot of non-firmware in the repo, for wireless telemetry for ex. Also there will be boards not running firmware i.e. the Linux display. 
6. Moved all boards into a subdirectory of `firmware` called `thruna`. This will make it clear which boards are for which car when we add a `quadruna` directory later.

### Testing Done

- CI passes


Lucien950 (approved):
image
image

will-chaba (approved):

PR #938, opened 2023-08-16
BMS Auxiliary Thermistor Sense + Integration Changes

will-chaba:
### Summary

Add capability to read thermistor values

- inside all-states, in 100hz task read thermistor value (output from MUX) then update the mux channel to give time to switch before next read.
- broadcast thermistor temps over CAN

Also add Integration changes from Aug 19
- Add dead zone to apps pedals (still able to reach 100%)
- Fix issues related to latched faults on BMS
- Update paths to openocd on <board>.elf files

### Changelist 

- BMS_tx.json
- App_Accumulator.h/.c
- Io_ThermistorReadings.h/.c
- BMS App_AllStates.c

### Testing Done

- Testing will be done on BMS/car as a whole saturday

### Checklist


firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
```diff
@@ -0,0 +1,136 @@
...
+    2007.1f,  1972.3f,  1938.3f,  1904.9f,  1872.2f,  1840.1f,  1808.7f,  1777.9f,  1747.7f,  1718.0f,  1689.0f,
+    1660.5f,  1632.6f,  1605.2f,  1578.3f,  1552.0f,  1526.1f,  1500.8f,  1475.9f,  1451.5f,  1427.6f,  1404.2f,
+    1381.1f,  1358.5f,  1336.4f,  1314.6f,  1293.3f,  1272.4f,  1251.8f
+};
+
+static void intToBinaryArray(uint8_t input_int, GPIO_PinState *output_arr)
+{
+    for (uint8_t i = 0; i < 4; i++)
+    {
+        output_arr[i] = GPIO_PIN_RESET;
```
will-chaba, line 38: may want to use intermediary variable here

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -465,3 +486,27 @@ bool App_Accumulator_BalancingEnabled(struct Accumulator *const accumulator)
...
+    accumulator->aux_thermistor_temps[thermistor_channel] = accumulator->read_thermistor_temp();
+
+    if (thermistor_channel < (NUM_AUX_THERMISTORS - 1))
+    {
+        thermistor_channel++;
+    }
+    else
+    {
+        thermistor_channel = 0;
+    }
```
gtaharaedmonds, line 501: This can be: 
```c
thermistor_channel++;
thermistor_channel %= NUM_AUX_THERMISTORS 
```

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
```diff
@@ -0,0 +1,136 @@
...
+    7097.0f,  6950.1f,  6806.6f,  6666.6f,  6529.7f,  6396.1f,  6265.6f,  6138.1f,  6013.5f,  5891.8f,  5772.9f,
+    5656.7f,  5543.2f,  5432.3f,  5323.9f,  5217.9f,  5114.4f,  5013.2f,  4914.2f,  4817.5f,  4722.9f,  4630.5f,
+    4540.1f,  4451.7f,  4365.3f,  4280.8f,  4198.1f,  4117.3f,  4038.2f,  3960.9f,  3885.2f,  3811.2f,  3738.8f,
+    3668.0f,  3598.7f,  3530.9f,  3464.6f,  3399.7f,  3336.1f,  3273.9f,  3213.1f,  3153.5f,  3095.2f,  3038.1f,
+    2982.3f,  2927.6f,  2874.0f,  2821.6f,  2770.3f,  2720.0f,  2670.8f,  2622.6f,  2575.3f,  2529.1f,  2483.8f,
+    2439.5f,  2396.0f,  2353.4f,  2311.7f,  2270.9f,  2230.9f,  2191.6f,  2153.2f,  2115.6f,  2078.7f,  2042.5f,
+    2007.1f,  1972.3f,  1938.3f,  1904.9f,  1872.2f,  1840.1f,  1808.7f,  1777.9f,  1747.7f,  1718.0f,  1689.0f,
+    1660.5f,  1632.6f,  1605.2f,  1578.3f,  1552.0f,  1526.1f,  1500.8f,  1475.9f,  1451.5f,  1427.6f,  1404.2f,
+    1381.1f,  1358.5f,  1336.4f,  1314.6f,  1293.3f,  1272.4f,  1251.8f
+};
```
gtaharaedmonds, line 32: Thermistor resistance vs. temp can be modelled via the [Steinhart-Hart equation](https://en.wikipedia.org/wiki/Steinhart%E2%80%93Hart_equation). Requires a `ln` calculation but I think we have enough CPU headroom that that should be fine. Specifically:

image

`B` and `R_0` at `T_0=25C` are in the datasheet. Make sure you use Kelvin for the actual calculation.

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
```diff
@@ -0,0 +1,136 @@
...
+    }
+    if (input_int >= 2)
+    {
+        output_arr[1] = GPIO_PIN_SET;
+        input_int     = (uint8_t)(input_int - 2U);
+    }
+    if (input_int == 1)
+    {
+        output_arr[0] = GPIO_PIN_SET;
+    }
```
gtaharaedmonds, line 59: Try this: 
```c
output_arr[3] = (input_int & 1<<3) ? GPIO_PIN_SET : GPIO_PIN_RESET;
output_arr[2] = (input_int & 1<<2) ? GPIO_PIN_SET : GPIO_PIN_RESET;
output_arr[1] = (input_int & 1<<1) ? GPIO_PIN_SET : GPIO_PIN_RESET;
output_arr[0] = (input_int & 1<<0) ? GPIO_PIN_SET : GPIO_PIN_RESET;
```
nit: This is probably simple enough you don't need the helper func and can do something like this below:
```c
HAL_GPIO_WritePin(AUX_TSENSE_MUX0_GPIO_Port, AUX_TSENSE_MUX0_Pin, (channel & 1<<0) ? GPIO_PIN_SET : GPIO_PIN_RESET);
...
```

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
```diff
@@ -0,0 +1,136 @@
...
+    1660.5f,  1632.6f,  1605.2f,  1578.3f,  1552.0f,  1526.1f,  1500.8f,  1475.9f,  1451.5f,  1427.6f,  1404.2f,
+    1381.1f,  1358.5f,  1336.4f,  1314.6f,  1293.3f,  1272.4f,  1251.8f
+};
+
+static void intToBinaryArray(uint8_t input_int, GPIO_PinState *output_arr)
+{
+    for (uint8_t i = 0; i < 4; i++)
+    {
+        output_arr[i] = GPIO_PIN_RESET;
+    }
```
gtaharaedmonds, line 39: No need to clear here since you're setting each index below.

gtaharaedmonds (changes requested):
where do you switch the MUX channel?

will-chaba:
> where do you switch the MUX channel?

Ah yes good catch, I'm missing a call to `accumulator->select_mux_channel` after the read

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
will-chaba: Ah using shifts is smart. I'll look into simplifying that using your suggestions

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
will-chaba: I had a look and the existing LUT is very close to what I calculated for the new thermistor, could try implementing the actual Steinhart-Hart equation, but will likely try with LUT first

firmware/thruna/BMS/Src/App/App_Accumulator.c
will-chaba: I'm getting the sense someone has done a bit of leetcode...

But yes that would be much cleaner

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
will-chaba: Left the helper function for now, I do like the separation for easier readability/debugging but could sway either way. Leaving conversation un-resolved until decision made

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
will-chaba: I tried looking into this briefly but had issues using the log function from math.h, we can try again in the future but the lookup table was working well for now

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
will-chaba: Decided to keep the helper function, I think It makes it more readable but I agree it's unncessary

firmware/thruna/FSM/Src/App/states/App_AllStates.c
```diff
@@ -3,6 +3,10 @@
 #include "states/App_DriveState.h"
 #include "App_SharedConstants.h"
 
+#define IGNORE_HEARTBEAT_CYCLES 1000U
```
will-chaba, line 6: @Eddy13013 was this necessary?

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -444,9 +465,9 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
 void App_Accumulator_BroadcastLatchedFaults(struct Accumulator *const accumulator)
 {
     // Check Latched fault pins for IMD and BSPD
-    const bool imd_latched_fault  = accumulator->check_imd_latched_fault;
-    const bool bspd_latched_fault = accumulator->check_bspd_latched_fault;
-    const bool bms_latched_fault  = accumulator->check_bms_latched_fault;
+    const bool imd_latched_fault  = accumulator->check_imd_latched_fault();
+    const bool bspd_latched_fault = accumulator->check_bspd_latched_fault();
+    const bool bms_latched_fault  = accumulator->check_bms_latched_fault();
```
will-chaba, line 470: Turns out you need the brackets for it to properly call the function. Who knew? 🤦‍♂️

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
gtaharaedmonds: Ok sounds good. What were the issues?

will-chaba: I’d include math.h but trying to use any of the log functions would give a undeclared function error

can_bus/json/BMS/BMS_tx.json
```diff
@@ -292,5 +292,48 @@
                 "unit": "W"
             }
         }
+    },
+    "BMS_AuxThermistors": {
+        "msg_id": 121,
+        "cycle_time": 100,
+        "description": "BMS Auxiliary Thermistor Temperatures",
+        "signals": {
+            "ThermTemp0": {
```
gtaharaedmonds, line 301: Are these populated in permanent locations? We should give them better names if so. Fine to leave that for a follow-up though.

firmware/thruna/BMS/Src/App/states/App_AllStates.c
```diff
@@ -11,7 +11,7 @@
 #define CELL_FULLY_DERATED_TEMP (60.0f)
 
 // Num of cycles for voltage and cell temperature values to settle
-#define NUM_CYCLES_TO_SETTLE (3U)
+#define NUM_CYCLES_TO_SETTLE (30U)
```
gtaharaedmonds, line 14: Why did we change this?

firmware/thruna/DCM/Src/App/states/App_AllStates.c
```diff
@@ -3,6 +3,10 @@
 #include "states/App_AllStates.h"
 #include "App_EllipseImu.h"
 
+#define IGNORE_HEARTBEAT_CYCLES 1000U
+
+static uint16_t num_cycles = 0;
+
```
gtaharaedmonds, line 9: Can we use an `App_Timer` for this?

gtaharaedmonds (commented):
lgtm, few minor things and ill approve

firmware/thruna/BMS/Src/Io/Io_ThermistorReadings.c
gtaharaedmonds: Oh linker error right? You probably need to link the math library.
Pretty easy change in CMake if we ever want to get better accuracy later.

firmware/thruna/BMS/Src/App/states/App_AllStates.c
will-chaba: Was seeing over voltage fault on startup, 30 may be excessive but 3 was too little. Perhaps more time should be taken to tune this and check if a FSM missing hb ignore timer is actually needed. I obviously won’t be around but if the car is together with the accumulator in, would be nice to check these things over

gtaharaedmonds: ok fair enough. don't love that we need to extend this all of a sudden. 
maybe drop a TODO and we can look into reducing this later.

will-chaba: It could be that it’s been silently faulting for a while now, only noticed it while debugging a issue with the latched fault signal going to the DIM. Ended up being completely unrelated, but dipping into fault state and setting the BMS ok status low should be enough to latch the fault in hardware and cause the DIM to display a latched fault. 

My suspicion is that this was actually happening at comp as well, and that it bounced out of fault state and set the ok status back high before hardware was able to latch the fault, though I have no way to prove that.

Definitely should add a TODO: and investigate further

gtaharaedmonds (approved):

can_bus/json/BMS/BMS_tx.json
will-chaba: At the time of testing they had not been, will create a follow up with better labels after consulting with Mallory

firmware/thruna/DCM/Src/App/states/App_AllStates.c
will-chaba: Sure, looks like Edwin applied a 10 second ignore period, seems a bit much going to try with one second. Still using a static flag to indicate startup, let me know if you can think of a better method of identifying startup condition. Could use entry into init state, but would run again after startup.

will-chaba: Actually, before I go through with that I'd like to see if this change is even necessary, will test with and without the timer next time I get the chance to see if it's even needed.

gtaharaedmonds: Ah we have an `App_AllStates_Init` func, put it in there. You also need to call `App_Timer_Restart` to get it to actually start running, `UpdateAndGetState` just checks the current state. 

Also add a `App_Timer_SetCurrentTimeMS` into the 1kHz run task to set the timebase for timers. See FSM `main.c` for example.

gtaharaedmonds: Actually yeah on second thought why does only the DCM needs this, and why now?
The BMS and FSM have 1sec `HAL_Delay` at startup, the DCM and DIM do not, so it could  be due to that? But I don't know why this is only happening now? 
(this was apparently to fix CAN issues at startup)

will-chaba: I think it may have been added to debug an unrelated issue which has now been [resolved](https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/938#discussion_r1306574393), so I'll double-check that it's needed

will-chaba: I tried removing the timer to ignore heartbeats but DCM does fault on startup. Good news is that only a 1 cycle delay was required to prevent it. I tuned the timer to wait 3 cycles just to add some breathing room, but for a delay this short, I don't think the complexity of using the App_Timer function is worth it.

PR #943, opened 2023-09-02
Integrate SEGGER SystemView

gtaharaedmonds:
### Summary
Integrate SEGGER SystemView so we can use SystemView to track CPU load and gain visibility into the RTOS. The embedded SEGGER SystemView sources were downloaded from https://www.segger.com/downloads/systemview/ and added into a new submodule under the `UBCFormulaElectric` organization. 

Using SEGGER SystemView should now be as simple as:

1. Updating Git submodules.
2. Installing it from https://www.segger.com/downloads/systemview/.
3. Connecting a J-Link and opening up SystemView.

I also removed the old SEGGER RTT and TraceRecorder sources since SystemView comes bundled with RTT, and TraceRecorder integration with SystemView looked difficult (both couple tightly with FreeRTOS). TraceRecorder is also paid, while SystemView is free, so I think we should just stick with SystemView.

TraceRecorder does offer student licenses however, so it might be worth looking into getting one and trying it out. 

### Testing Done

- [x] Tested on old DCM.
- [x] More testing on board.


will-chaba (approved):
Looks good to me, we should add documentation on how to use SystemView to confluence, I think you're the only person who's used it before

gtaharaedmonds:
> Looks good to me, we should add documentation on how to use SystemView to confluence, I think you're the only person who's used it before

Yeah this could be a good thing to add to the week 3 onboarding, I'd like to rely on this more significantly this year.

PR #988, opened 2023-09-22
BMS Coulomb Counting SOC

will-chaba:
### Summary

Track state-of-charge using coulomb counting on BMS. No current limiting implemented, solely statistical tracking. 

SOC is saved to EEPROM on 1s intervals, utilizing a basic wear-leveling algorithm
- Address of up-to-date soc stored in page 0 of EEPROM, this is only written once per power cycle, no need to wear-level
- SOC value stored in page 1-127, written once per second. Active SOC page written to will increment once per power cycle.
- SOC value read is checked for corruption:
   - When SOC value saved to EEPROM, 4 copies saved to page
   - When SOC value read, 4 values are checked to see if they match and are within range
   - If not, read SOC from Ocv to SOC lookup table (inaccurate but better than corrupted value)

SOC can be reset in one of two ways, both controlled by debug CAN messages
- Update from Ocv-Soc LUT
- Update with manually specified SOC value

### Changelist 

- Added SOC tracking logic
- Added EEPROM reading/writing
- Added simple test for SOC

### Testing Done

- Test checking for accuracy added to TestStateMachine
- LUT lookup tested in TestStateMachine (will be moved during refactor)
- Tested on car:
-  SOC can be reset using both custom value and via OCV lookup from CAN debug message
- Test value saved and recovered between power cycles from EEPROM
- Checked that SOC tracks linearly with current draw (tested with full charge cycle)

### Checklist


firmware/thruna/BMS/Inc/App/App_Soc.h
```diff
@@ -1,6 +1,37 @@
...
+ * @return soc related to open circuit voltage
+ */
+float App_Soc_GetSocFromVoc(float voltage);
+
+/**
+ * Function to retrieve open circuit voltage closest to given soc
+ * @param soc_percent state of charge in percent
+ * @return open circuit voltage closest to given state of charge
+ */
+float App_Soc_GetVocFromSoc(float soc_percent);
```
1mackenziekyle, line 19: Pedantic but Voc should be Ocv

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -949,4 +961,97 @@ TEST_F(BmsStateMachineTest, check_precharge_state_transitions_and_air_plus_statu
...
+    expected_soc = 100.0;
+
+    ASSERT_FLOAT_EQ(soc, expected_soc);
+}
+
+TEST_F(BmsStateMachineTest, soc_to_voc_lut_test)
+{
+    float test_soc = 0.0f;
+
+    float voc = App_Soc_GetVocFromSoc(test_soc);
```
1mackenziekyle, line 1032: same thing as above, voc -> ocv (up to you if this actually matters)

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -459,7 +459,10 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
     App_CanAlerts_BMS_CellOvertempFault_Set(overtemp_fault);
     App_CanAlerts_BMS_ModuleCommunicationFault_Set(communication_fault);
 
-    return (overtemp_fault || undertemp_fault || overvoltage_fault || undervoltage_fault || communication_fault);
+    const bool acc_fault =
+        overtemp_fault || undertemp_fault || overvoltage_fault || undervoltage_fault || communication_fault;
+
+    return acc_fault;
```
will-chaba, line 465: Personally find this easier to debug without needing to set custom breakpoints

firmware/thruna/BMS/Src/App/App_Eeprom.c
```diff
@@ -1,10 +1,13 @@
 #include "App_Eeprom.h"
 #include "string.h"
 
-#define NUM_PAGES (128U)
+#define NUM_PAGES (16U)
```
will-chaba, line 4: Ran into issues with address corruption above 16, could look into this further but with SD card on the way I'm not too sure how valuable that'd be

firmware/thruna/BMS/Src/App/App_Soc.c
```diff
@@ -1,5 +1,50 @@
 #include <math.h>
 #include "App_Soc.h"
+#include <stdint.h>
+#include <float.h>
+#include "App_SharedConstants.h"
+#include "App_SharedProcessing.h"
+#include "lut/App_Lut_CellVoltageToSoc.h"
+
+#define STATE_OF_HEALTH (1.0f)
```
will-chaba, line 9: Will update this based on Edwin's experimental testing before merging, was told is closer to 0.93

firmware/thruna/BMS/Src/App/App_Eeprom.c
```diff
@@ -114,33 +117,150 @@ EEPROM_StatusTypeDef App_Eeprom_PageErase(struct Eeprom *eeprom, uint16_t page)
...
+    {
+        float_arr[i] = min_soc;
+    }
+
+    return App_Eeprom_WriteFloats(eeprom, address, DEFAULT_OFFSET, float_arr, SAVED_COPIES);
+}
+
+ExitCode App_Eeprom_ReadMinSoc(struct Eeprom *eeprom, uint16_t address, float *min_soc)
+{
+    float                float_arr[SAVED_COPIES];
```
1mackenziekyle, line 223: Change variable name to something more meaningful. What is float_arr?

firmware/thruna/BMS/Src/App/App_Soc.c
1mackenziekyle: Will you be measuring state of health? Or just going off of Edwin's findings

firmware/thruna/BMS/Src/App/App_Soc.c
```diff
@@ -1,5 +1,50 @@
...
+    if (lut_index == V_TO_SOC_LUT_SIZE)
+    {
+        // Ensures that the index is in the LUT range
+        lut_index--;
+    }
+
+    return ocv_soc_lut[lut_index];
+}
 
 ExitCode App_Soc_Vote(float max_abs_difference, float soc_1, float soc_2, float soc_3, float *result)
```
1mackenziekyle, line 49: What does this function do?

firmware/thruna/BMS/Src/App/App_Soc.c
will-chaba: In the future yes, but don't have accurate enough data yet. I did get an estimated value of 0.946 from a charging cycle last Thursday, comparing the computed SOC to the value found on the LUT based on settled min-cell-voltage, but given my starting point for the computed SOC was close to the 5% garbage region, I don't know how accurate that is. Definitely need to do something about not having data below 5%.

firmware/thruna/BMS/Src/App/App_Soc.c
will-chaba: This isn't actually used and I should probably remove it, it's leftover from Prom's implementation. But what it's meant to do is to compare 3 soc readings from EEPROM and determine if any of them are corrupt. I do similar functionality in App_Eeprom.c

firmware/thruna/BMS/Src/App/App_Eeprom.c
will-chaba: it's an array of floats to hold the contents read from the EEPROM. EEPROM is written in bytes, so the ReadFloats function reads in bytes and converts them to floats which are written to the input float array. I can update the name

firmware/thruna/BMS/Src/main.c
```diff
@@ -312,15 +313,40 @@ int main(void)
...
+    uint16_t soc_address;
+
+    eeprom = App_Eeprom_Create(Io_Eeprom_WritePage, Io_Eeprom_ReadPage, Io_Eeprom_PageErase);
+
+    // A negative soc value will indicate to App_SocStats_Create that saved SOC value is corrupted
+    float saved_soc_c = -1.0f;
+    if (App_Eeprom_ReadAddress(eeprom, &soc_address) == EXIT_CODE_OK)
+    {
+        if (App_Eeprom_ReadMinSoc(eeprom, soc_address, &saved_soc_c) == EXIT_CODE_OK)
+            ;
```
philam2001, line 325: why don't you just handle  
if (App_Eeprom_ReadMinSoc(eeprom, soc_address, &saved_soc_c) != EXIT_CODE_OK) case instead of having ; here

philam2001 (commented):
Are you going to broadcast some kind of warning or something if the SOC if corrupt and to not read it during like a 100HZ or a certain cycle?

firmware/thruna/BMS/Inc/App/App_BmsWorld.h
```diff
@@ -103,6 +105,13 @@ struct OkStatus *App_BmsWorld_GetBspdOkStatus(const struct BmsWorld *world);
  */
 struct Accumulator *App_BmsWorld_GetAccumulator(const struct BmsWorld *world);
 
+/**
+ * Get the soc_stats for the given world
+ * @param world The world to get the accumulator for
+ * @return The accumulator for the given world
+ */
+struct SocStats *App_BmsWorld_GetSocStats(const struct BmsWorld *const world);
+
```
gtaharaedmonds, line 114: we love the 🌎

firmware/thruna/BMS/Inc/App/App_Eeprom.h
```diff
@@ -63,21 +65,36 @@ EEPROM_StatusTypeDef
...
+
+/**
+ * Read previous SOC value of cell with lowest voltage to EEPROM and perform error checking
+ * @param eeprom
+ * @param address
+ * @param min_soc
+ * @return ExitCode returns EXIT_CODE_ERROR in case of data corruption or read error
  */
-EEPROM_StatusTypeDef App_Eeprom_Read4CopiesOfAddress(struct Eeprom *eeprom, uint16_t page, uint16_t *addresses);
+ExitCode App_Eeprom_ReadMinSoc(struct Eeprom *eeprom, uint16_t address, float *min_soc);
```
gtaharaedmonds, line 100: This module could probably use a better name. By name it sounds like an application-level wrapper around the EEPROM, but it seems highly specific to SOC. Consider renaming to something more accurate.

firmware/thruna/BMS/Inc/App/App_Soc.h
```diff
@@ -1,6 +1,40 @@
...
+    double charge_c;
+
+    // Charge loss at time t-1
+    float prev_current_A;
+
+    // Indicates if SOC from EEPROM was corrupt at startup
+    bool is_corrupt;
+
+    TimerChannel soc_timer;
+};
```
gtaharaedmonds, line 37: put struct definitions above function definitions

firmware/thruna/BMS/Src/App/App_Eeprom.c
gtaharaedmonds: this seems extremely sus but OK

firmware/thruna/BMS/Src/App/App_Soc.c
gtaharaedmonds: when you update it, please leave a comment on where the number comes from

firmware/thruna/BMS/Src/App/App_Soc.c
gtaharaedmonds: yep please remove if not used

firmware/thruna/BMS/Src/App/App_Soc.c
```diff
@@ -35,3 +80,98 @@ ExitCode App_Soc_Vote(float max_abs_difference, float soc_1, float soc_2, float
...
+    return soc_percent;
+}
+
+float App_SOC_GetMinOcvFromSoc(struct SocStats *soc_stats)
+{
+    float soc_percent = App_SocStats_GetMinSocPercent(soc_stats);
+    return App_Soc_GetOcvFromSoc(soc_percent);
+}
+
+void App_SOC_ResetSocFromVoltage(struct SocStats *soc_stats, struct Accumulator *accumulator)
```
gtaharaedmonds, line 161: nit: stick with either `App_Soc` or `App_SOC`. I like `App_Soc`

firmware/thruna/BMS/Src/main.c
gtaharaedmonds: agreed
also this logic is probably better handled in `App_Eeprom_Create` or some other EEPROM file initialization function

firmware/thruna/BMS/Test/Src/Test_Faults.cpp
```diff
@@ -53,6 +53,8 @@ FAKE_VALUE_FUNC(EEPROM_StatusTypeDef, read_page, uint16_t, uint8_t, uint8_t *, u
 FAKE_VALUE_FUNC(EEPROM_StatusTypeDef, write_page, uint16_t, uint8_t, uint8_t *, uint16_t);
 FAKE_VALUE_FUNC(EEPROM_StatusTypeDef, page_erase, uint16_t);
 
+static float saved_soc = 100.0f;
```
gtaharaedmonds, line 56: do you need this variable?

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -82,6 +87,8 @@ class BmsStateMachineTest : public BaseStateMachineTest
         App_CanTx_Init();
         App_CanRx_Init();
 
+        saved_soc = SERIES_ELEMENT_FULL_CHARGE_C;
```
gtaharaedmonds, line 90: again, do you need this?

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -269,13 +280,14 @@ class BmsStateMachineTest : public BaseStateMachineTest
...
     struct Accumulator *      accumulator;
+    struct SocStats *         soc_stats;
     struct Airs *             airs;
     struct PrechargeRelay *   precharge_relay;
     struct TractiveSystem *   ts;
     struct Clock *            clock;
     struct Eeprom *           eeprom;
 };
-
+#include <assert.h>
```
gtaharaedmonds, line 290: do you need this?

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
```diff
@@ -921,4 +933,97 @@ TEST_F(BmsStateMachineTest, check_precharge_state_transitions_and_air_plus_statu
...
+    ASSERT_FLOAT_EQ(soc, expected_soc);
+
+    // Check that SOC saturates at 100.0%
+    test_voltage = 5.0f;
+
+    soc = App_Soc_GetSocFromOcv(test_voltage);
+
+    expected_soc = 100.0;
+
+    ASSERT_FLOAT_EQ(soc, expected_soc);
```
gtaharaedmonds, line 997: interesting whitespace decisions here

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
gtaharaedmonds: or, does it need to be static?

firmware/thruna/BMS/Inc/App/App_Soc.h
```diff
@@ -37,3 +71,70 @@
...
+ * @return soc % of lowest SOC SE
+ */
+float App_SocStats_GetMinSocPercent(struct SocStats *soc_stats);
+
+/**
+ * Get the minimum series element open circuit voltage approximation given current pack SOC status
+ * @param soc_stats current SOC of each series element in pack
+ * @return float minimum series element open circuit voltage approximation
+ */
+float App_SOC_GetMinOcvFromSoc(struct SocStats *soc_stats);
```
gtaharaedmonds, line 126: Instead of having a bunch of getters that are only sent over CAN, I like the pattern of just having a single "Broadcast" function that handles setting all the TX values. I did this on the accumulator in the BMS refactor and it cuts the file size in half, and is a lot clearer + more maintainable in my opinion

firmware/shared/src/app/App_SharedProcessing.h
```diff
@@ -7,10 +7,10 @@
  * @param f_x1 The current value of f(x) used to approximate the value of the current trapezoid
  * @param d_x The uniform delta-x of each trapezoid
  */
-static inline void App_SharedProcessing_TrapezoidalRule(float *integral, float *f_prev, float f_curr, float d_x)
+static inline void App_SharedProcessing_TrapezoidalRule(double *integral, float *f_prev, float f_curr, float d_x)
 {
     // Calculate the trapezoid and add to the previous integral
-    *integral = *integral + d_x * (*f_prev + f_curr) * 0.5f;
+    *integral = *integral + (double)(d_x * (*f_prev + f_curr) * 0.5f);
```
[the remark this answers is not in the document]
gtaharaedmonds: could also drop a comment on why this is a double (more precision, since this is basically the only place we use doubles (I think))

gtaharaedmonds (changes requested):

firmware/thruna/BMS/Src/App/App_Eeprom.c
will-chaba: Oh it's definitely sus. I'll add a comment.

firmware/thruna/BMS/Inc/App/lut/App_Lut_CellVoltageToSoc.h
```diff
@@ -1,141 +1,30 @@
...
-    3.9071f,  3.90225f, 3.89707f, 3.8919f,  3.88607f, 3.88025f, 3.87508f, 3.86958f, 3.8644f,  3.85923f, 3.85405f,
-    3.8492f,  3.84435f, 3.8395f,  3.83497f, 3.83044f, 3.82591f, 3.82171f, 3.81718f, 3.8133f,  3.80942f, 3.80586f,
-    3.80198f, 3.79842f, 3.79486f, 3.79131f, 3.78775f, 3.78484f, 3.78128f, 3.77837f, 3.77514f, 3.77222f, 3.76931f,
-    3.7664f,  3.76349f, 3.76058f, 3.75832f, 3.75541f, 3.75314f, 3.75088f, 3.74829f, 3.74603f, 3.74376f, 3.74182f,
-    3.73988f, 3.73794f, 3.736f,   3.73438f, 3.73277f, 3.73147f, 3.72985f, 3.72824f, 3.72662f, 3.72533f, 3.72339f,
-    3.72177f, 3.71983f, 3.71756f, 3.7153f,  3.71304f, 3.7098f,  3.70689f, 3.70398f, 3.70075f, 3.69751f, 3.69395f,
-    3.69039f, 3.68651f, 3.68166f, 3.67681f, 3.67228f, 3.66905f, 3.66646f, 3.66452f, 3.66193f, 3.65935f, 3.65611f,
-    3.65191f, 3.64576f, 3.63606f, 3.62086f, 3.59886f, 3.56911f, 3.53029f, 3.47919f, 3.40965f, 3.30648f, 3.07522f
+// index 0 represents 5.0% soc, final index represents 100%. 0.5% increments.
+static const float ocv_soc_lut[V_TO_SOC_LUT_SIZE] = {
```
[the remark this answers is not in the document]
will-chaba: The LUT will be constant yes, but SOC will take into account a state-of-health factor to compensate for aging. This LUT is only accurate when cells have settled with 0 current draw for 1 hour or more.

will-chaba:
> Are you going to broadcast some kind of warning or something if the SOC if corrupt and to not read it during like a 100HZ or a certain cycle?

Yes there is a CAN message to broadcast if SOC value read is corrupt. In this case, it will reset based on min-cell voltage, but that is inaccurate unless cell voltages have settled. So we need to know it's a dirty starting-value.

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
will-chaba: No, not sure why that's there good catch

firmware/thruna/BMS/Test/Src/Test_StateMachine.cpp
will-chaba: Agreed will clean up

firmware/thruna/BMS/Test/Src/Test_Faults.cpp
will-chaba: There needs to be a value input into App_SocStats_Create, for now all tests use SERIES_ELEMENT_FULL_CHARGE_C (100% SOC), but may be changed in the future. I'll replace it with a constant for now.

firmware/thruna/BMS/Src/App/App_Soc.c
will-chaba: Good catch App_SOC was a typo.

firmware/thruna/BMS/Src/main.c
will-chaba: Moved to App_SocStats_Create

firmware/thruna/BMS/Inc/App/App_Soc.h
will-chaba: Only `App_SocStats_GetMinSocPercent` is being broadcast over CAN. `App_SOC_GetMinOcvFromSoc` and `App_SocStats_GetMinSocCoulombs` are implemented for future use, but not currently utilized.

I can remove `App_SocStats_GetMinSocCoulombs` if it bothers you, but `App_SOC_GetMinOcvFromSoc` will be needed for current limiting in the near future.

firmware/thruna/BMS/Inc/App/App_Eeprom.h
will-chaba: I guess one thing I could do is rather than include the EEPROM object in the world, have it be a part of the SocStats struct as it's the only thing using the EEPROM then I could place these SOC-related EEPROM functions in App_Soc.c

can_bus/json/BMS/BMS_tx.json
```diff
@@ -9,6 +9,15 @@
...
                 "enum": "BmsState"
+            },
+            "Soc": {
+                "resolution": 0.1,
+                "min": 0,
+                "max": 100,
+                "unit": "%"
+            },
+            "SocCorrupt": {
+                "bits": 1
```
will-chaba, line 20: Added to Bms_Vitals as we have hit the limit on the # of messages the BMS can send before hitting tx overflow, will need to address this

firmware/thruna/BMS/Src/App/App_Soc.c
will-chaba: Edwin is unsure of where his State of Health number is doccumented, but believes it to be around 93%. Based on math from charging above, I got an estimated value of 94.46%. For now I used a value of 0.94, or 94% splitting the difference. This can be tuned in the future with more testing.

will-chaba:
Testing on Nov 12 2023 with driving testing, started the day at around 77% and ended the day at 30.2%. Referencing the OCV LUT and min cell voltage revealed that the "actual" SOC was 24.9%, so we are about 5% optimistic at the moment. Further tuning will be needed, but otherwise tested working.

gtaharaedmonds (approved):
dub

PR #990, opened 2023-09-22
Build system quickfixes

gtaharaedmonds:
### Summary

Fix the dubious ownership issue by executing this command recursively inside of a `start_environment.sh` bash script.

```sh
# fatal: detected dubious ownership in repository at '/root/Consolidated-Firmware'
# To add an exception for this directory, call:
#
#         git config --global --add safe.directory /root/Consolidated-Firmware
```

Also deprecate the `-DNO_VENV` flag and remove the virtual environment, since it is unnecessary with the container.


Lucien950 (approved):
Do it 😈

Lucien950:
i'm also still hung on the opt in pipenv thing but we can do that later

PR #993, opened 2023-09-24
Day 1 build system hotfixes

gtaharaedmonds:
### Summary

Various build system fixes from meeting 1.


environment/scripts/install_stm32cubemx.py
gtaharaedmonds, line 1: Changes in this file are all from linting

firmware/CMakeLists.txt
```diff
@@ -94,7 +107,7 @@ function(cube_mx_code_generation
     add_custom_command(
         OUTPUT ${IOC_FILE}.md5
         ${LOG4J_PROPERTIES}
-        COMMAND python3
+        COMMAND ${XVFB_PREFIX} python3
```
Lucien950, line 110: I assume that the build system error `xvfb-run: error: Xvfb failed to start` is coming from here.

Lucien950 (changes requested):

firmware/CMakeLists.txt
gtaharaedmonds: Yeah I'm trying to fix this now

gtaharaedmonds: it doesn't happen when building locally

will-chaba (approved):
Seems good to me. If we're going native linux/wsl don't we need the Pipfile back?

gtaharaedmonds:
> Seems good to me. If we're going native linux/wsl don't we need the Pipfile back?

we could but for development personally I don't think python virtual environments add a lot of value
the situation where you need a different version of package for project A vs. B is pretty rare 
also pipenv is very slow

PR #994, opened 2023-09-24
DIM Test Refactor

will-chaba:
### Summary

Experimenting with separating tests into different files by function. This is pretty pointless on the DIM, as there aren't many tests, but for boards such as the BMS which have many tests that end up all being crammed into Test_StateMachine.cpp, having a better mechanism to separate tests into distinct files is valuable. (See Test_Faults.cpp and Test_Eeeprom.cpp for example, they require the same setup as the state machine test, and it means that 3 files need to be maintained any time the state machine test setup or teardown needs to be modified.)

Admittedly the DIM is a poor example as to why this is beneficial, and I would argue that tests have been over-segregated in this example (just to get the point across), however, I cannot re-iterate how much this would benefit other boards.

#### Changes
* Move `DimStateMachineTest` class definition to a header file accessible to any .cpp test file. Also renamed to `DimBaseStateMachineTest` to fall in line with the existing `BaseStateMachineTest` class. 
  * I'm leaning towards naming it something more along the lines of DimBaseTest, as even code that does not require the state machine running can be tested within it (albeit slightly less efficiently, but none of our tests are that demanding anyway and will allow for a single source of truth when it comes to test initialization)
* `test_stateMachine.cpp` should focus on testing state transitions in my eyes (open to some other things residing here as well, please share feedback), and should not be the defacto place to add a new test that is unrelated to state machine behavior, instead create a new test class inheriting from `DimBaseStateMachineTest` and group tests with similar functions/goals into smaller test files.
   * Because the DIM has no state transitions, there is no longer a `test_stateMachine.cpp` file

### Changelist 

- test_stateMachine.cpp split into multiple files.

### Testing Done

- All existing tests pass

### Checklist


firmware/thruna/DIM/test/test_avgPowerCalc.cpp
```diff
@@ -1,15 +1,7 @@
...
-extern "C"
+class AveragePowerCalcTest : public DimBaseStateMachineTest
 {
-#include "app_avgPower.h"
-}
-
-class AveragePowerCalcTest : public testing::Test
-{
-  protected:
-    void SetUp() override { app_avgPower_init(); }
```
will-chaba, line 12: previous AveragePwoerCalcTest only required the initialization of `app_avgPower_init()` rather than initializing the whole state machine, however to better fit the two tests relating to average power calculation together, needed the state machine anyways, and I think the slight performance overhead of initializing the state machine is worth having a cleaner file

firmware/thruna/DIM/test/test_dimCanMsgs.cpp
gtaharaedmonds, line 1: nit: thoughts on removing "DIM" from these file names? In my opinion they already have DIM in their path from the DIM folder so repeating DIM is redundant info. but I don't feel strongly about this

gtaharaedmonds (approved):
this PR is very good

firmware/thruna/DIM/test/test_dimCanMsgs.cpp
will-chaba: hmm yeah, that's a good point, I'd like to keep it on the header file containing the board-specific base test so as to not confuse it with "Test_BaseStateMachineTest.h", but otherwise I think they could be removed. I'll push that and we can see how it looks

PR #997, opened 2023-09-30
remove gcc only feature

gtaharaedmonds:
### Summary

Structs without members are a GCC only feature:
```c
typedef struct 
{
// No members!
} Struct;
```
We've been using this since everyone has been compiling with GCC. However, MacOS seems to enforce compiling with `clang`, which doesn't allow this. So, remove the few cases we use it.


will-chaba (approved):
Looks good

PR #999, opened 2023-09-30
Firmware Tutorial Project

Steven141:
### Summary
In order to walk through all the required steps to configure MCU pins, read from I/O, and send/receive CAN messages, we’re going to walk through a simple project to display data onto the DIM 7-Segment displays.


Lucien950:
Good work 🙂

PR #1005, opened 2023-10-06
Blown Fuse Check

KelmLelm:
### Summary

Initial implementation for a fuse check in App_Accumulator (Battery Cells)
### Changelist 

App_Accumulator.c added the blown fuse check code in CheckFaults,  BMS_Alert.json added blown fuse fault.
### Testing Done

Just initial implementation. Nothing yet.
### Checklist

### Confluence Link
https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/97222845/Software+-+Thruna+FW+-+BMS+-+Blown+Fuse+Check


KelmLelm:
Like the description says, just an initial draft of implementation, not unit tested yet

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -18,6 +18,7 @@
 #define BALANCE_DEFAULT_DUTY (50) // %
 #define BALANCE_TICKS_ON (100U)
 #define BALANCE_TICKS_OFF (100U)
+#define INTERNAL_CELL_RESISTANCE_OHMS (0.0025f)
```
gtaharaedmonds, line 22: Can you provide some way of linking to where you got this number? LiIon cell part number + link to datasheet would be great.

can_bus/json/BMS/BMS_alerts.json
```diff
@@ -25,6 +25,7 @@
         "BMS_FAULT_HAS_REACHED_MAX_V",
         "BMS_FAULT_TS_OVERCURRENT",
         "BMS_FAULT_PRECHARGE_ERROR",
-        "BMS_FAULT_LATCHED_BMS_FAULT"
+        "BMS_FAULT_LATCHED_BMS_FAULT",
+        "BMS_FAULT_BLOWN_CELL_FUSE"
```
gtaharaedmonds, line 29: Can you make this a warning instead? Faults will cause the car to shutdown so until this is reliable I don't want the car to shutdown unnecessarily. Also I don't think fuses blowing is a high enough risk to make this a fault at present.

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -436,13 +437,30 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
...
+            cell_voltage_minimum = cell_voltage;
+        }
+    }
+    float cell_voltage_avg = cell_voltage_sum/4.0;
+    float segment_current = App_TractiveSystem_GetCurrent(ts);
+
+    float num_of_cells = (segment_current)*INTERNAL_CELL_RESISTANCE_OHMS/
+                        (cell_voltage_avg - cell_voltage_minimum + segment_current*INTERNAL_CELL_RESISTANCE_OHMS/3.0);
+
     // if we are charging, max cell temp is 45C not 60C
```
gtaharaedmonds, line 456: Can you move this block to a helper function? I think it'd be a bit more readable.

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -436,13 +437,30 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
...
+        cell_voltage_sum += cell_voltage;
+        if(cell_voltage < cell_voltage_minimum){
+            cell_voltage_minimum = cell_voltage;
+        }
+    }
+    float cell_voltage_avg = cell_voltage_sum/4.0;
+    float segment_current = App_TractiveSystem_GetCurrent(ts);
+
+    float num_of_cells = (segment_current)*INTERNAL_CELL_RESISTANCE_OHMS/
+                        (cell_voltage_avg - cell_voltage_minimum + segment_current*INTERNAL_CELL_RESISTANCE_OHMS/3.0);
```
gtaharaedmonds, line 454: Hmm I don't think `cell_voltage_minimum` and `cell_voltage_avg` are what we want. 

It looks like `cell_voltage_minimum` is the minimum of the average cell voltages for each segment - you want the individual cell with the lowest voltage. This info is in the `accumulator->voltage_status.min_voltage` variable.

`cell_voltage_avg` is `cell_voltage_sum` divided by 4? where does this come from?

gtaharaedmonds: `num_of_cells` isn't super descriptive - maybe `estimated_min_cells_in_parallel`?

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -436,13 +437,30 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
...
+
     // if we are charging, max cell temp is 45C not 60C
     if (App_TractiveSystem_GetCurrent(ts) > 3.0f)
     {
         max_allowable_cell_temp = MAX_CELL_CHARGE_TEMP_DEGC;
         min_allowable_cell_temp = MIN_CELL_CHARGE_TEMP_DEGC;
     }
 
+    bool blown_fuse_fault = num_of_cells < 2.5;
```
gtaharaedmonds, line 463: make a macro for this 2.5 value at the top - also maybe drop a TODO about needing to do testing on the car to determine this 2.5 number. 
(unnecessary if we're going to determine that before we merge this in, but then you should write something about where the number comes from)

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -453,13 +471,14 @@ bool App_Accumulator_CheckFaults(struct Accumulator *const accumulator, struct T
         App_Accumulator_GetMinVoltage(accumulator, &throwaway_segment, &throwaway_loc) < MIN_CELL_VOLTAGE;
     bool communication_fault = App_Accumulator_HasCommunicationError(accumulator);
 
+    App_CanAlerts_SetFault(BMS_FAULT_BLOWN_CELL_FUSE, blown_fuse_fault);
```
gtaharaedmonds, line 474: We should also add a CAN signal indicating the estimated min. number of cells in parallel, as well as the corresponding cell/segment

gtaharaedmonds (changes requested):
Great first draft!
We should add a unit test for this too.

firmware/thruna/BMS/Src/App/App_Accumulator.c
gtaharaedmonds: Also this math isn't super obvious - maybe link to the confluence doc?

gtaharaedmonds: Also also can you add a check to make sure we don't actually divide by zero? I don't think its possible but I don't want the board to crash if that ever does magically happen...

firmware/thruna/BMS/Inc/App/App_Accumulator.h
```diff
@@ -18,20 +18,23 @@ typedef enum
     ACCUMULATOR_NUM_SEGMENTS,
 } AccumulatorSegment;
 
+#define ACCUMULATOR_NUM_CELLS_PER_SEGMENT (3U)
```
will-chaba, line 21:
```suggestion
#define ACCUMULATOR_NUM_PARALLEL_CELLS (3U)
```

firmware/thruna/BMS/Src/App/states/App_AllStates.c
```diff
@@ -162,6 +163,10 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
         status = false;
         App_SharedStateMachine_SetNextState(state_machine, App_GetFaultState());
     }
+    else if (acc_warning)
+    {
+        // Cut torque requests to wheels??
```
will-chaba, line 168: For now, warnings are only displayed to the driver for them to act on, if we can confirm that this algorithm does not produce false-positives, could potentially look into either cutting torque or upgrading to a fault.

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -9,6 +9,12 @@
 
 #define NUM_AUX_THERMISTORS (6U)
 
+// What we will compare against to determine if a segment has a blown fuse and is
+// less than three cells in parallel. We calculate number of cells and compare.
+// TODO: TUNING AND TESTING REQUIRED, STILL A WORK IN PROGRESS THRESHOLD.
+#define BLOWN_FUSE_THRESHOLD (2.5f)
+#define FUSE_CHECK_CURRENT_THRESHOLD (40.0f)
```
will-chaba, line 16: Always good to specify units where applicable:
```suggestion
#define FUSE_CHECK_CURRENT_THRESHOLD_AMPS (40.0f)
```

will-chaba (commented):
Looks good so far!

firmware/thruna/BMS/Src/App/App_Accumulator.c
```diff
@@ -9,6 +9,12 @@
 
 #define NUM_AUX_THERMISTORS (6U)
 
+// What we will compare against to determine if a segment has a blown fuse and is
+// less than three cells in parallel. We calculate number of cells and compare.
+// TODO: TUNING AND TESTING REQUIRED, STILL A WORK IN PROGRESS THRESHOLD.
+#define BLOWN_FUSE_THRESHOLD (2.5f)
+#define FUSE_CHECK_CURRENT_THRESHOLD_AMPS (40.0f)
```
will-chaba, line 16: will need to lower current threshold for testing

will-chaba:
Out of date

PR #1016, opened 2023-10-10
Upgrade STM32CubeMX version from 5.3.0 -> 6.9.2

gtaharaedmonds:
### Summary

Upgrade the boards in our repo from STM32CubeMX 5.3.0 to 6.9.2. We need to switch to 6.9.2 to unlock support for the STM32H7xx family of microcontrollers, which is going to be on our vehicle controller this year. Also, we should just upgrade since they probably fixed a bunch of bugs.

TODO: Test this on other people's machines

### Changelist 

- Migrate every board's `.ioc` file from 5.3.0 to 6.9.2. This means we had to regenerate the drivers/middleware for every board, leading to a crazy diff.
- Switch to CMSIS-RTOS v2, which is supported in 6.9.2 (it wasn't in 5.3.0). CMSIS-RTOS is an API that wraps FreeRTOS, and v2 is way easier to use than regular FreeRTOS and way more powerful than v1. 
- Bumped the version in the container (using a custom tag currently so `ubcformulaelectric/environment` still points to the regular container, and other CI jobs can run).

### Testing Done

- [x] Flashed all boards and spun wheels on jacks
- [ ] This is a big change, so lets do some driving with these changes

### Checklist


.github/workflows/bin-dbc-gen.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    uses: UBCFormulaElectric/Consolidated-Firmware/.github/workflows/build.yml@master
```
Lucien950, line 9: this is going to be fixed in peter/build_DBC_PR, hopefully no conflicts.

environment/Dockerfile
```diff
@@ -14,27 +14,27 @@ FROM ubuntu:22.04
 # openocd: Our embedded GDB server of choice
 # usbutils: USB tools (lsusb)
 # nano: Basic Unix text editor
+# libgtk-3-0: Random library required for running STM32CubeMX in container
+# libxi6: Random library required for running STM32CubeMX V6 in container
+# libxtst6: Random library required for running STM32CubeMX V6 in container
+# libxrender1: Random library required for running STM32CubeMX V6 in container
 RUN apt update -y && apt install -y \
```
Lucien950, line 21: remove comments for unused dependencies

environment/Dockerfile
```diff
@@ -14,27 +14,27 @@ FROM ubuntu:22.04
...
     wget \
     unzip \
     pip \
     git \
     git-lfs \
-    openjdk-11-jdk \
     xvfb \
     libncurses5 \
     valgrind \
     openocd \
```
Lucien950, line 30: i don't think we need this if we're not flashing?

environment/Dockerfile
```diff
@@ -14,27 +14,27 @@ FROM ubuntu:22.04
...
+# libgtk-3-0: Random library required for running STM32CubeMX in container
+# libxi6: Random library required for running STM32CubeMX V6 in container
+# libxtst6: Random library required for running STM32CubeMX V6 in container
+# libxrender1: Random library required for running STM32CubeMX V6 in container
 RUN apt update -y && apt install -y \
     wget \
     unzip \
     pip \
     git \
     git-lfs \
```
Lucien950, line 26: if we clone from outside and volume, perhaps this is not used either. (git, gitlfs, unzip, wget??)

Lucien950 (changes requested):
just a few optimization things.

environment/Dockerfile
gtaharaedmonds: if I remember correctly I did have to install these to get the CI working but that was a long time ago
container takes too long to build for me to want to try

gtaharaedmonds:
> Also I am not familiar about autogenerated STM files but can they autogenerate on each build? (If so can we hide it away in build_arm)

Possibly yes. Currently we invoke CubeMX with the `generate code` command which just updates `main.c` and a few other files. We can also invoke with the `project generate` which should also update the driver files. Although Cube usually doesn't like being generated in a different directory than the `.ioc` so I'm not sure if we can put it in `build_arm`. 

This'll be easier after the refactor since all CubeMX stuff will live in its own folder.

Lucien950:
I'll review once #1040 gets merged in.

Lucien950 (approved):

PR #1018, opened 2023-10-10
Jiajun/build system refactor

Jiajun-Huang:
### Summary
Small change on some CMakelist.txt, allowing the different boards to include or exclude the code in shared directory.
It solves the dependency issue in shared code when one shared code depends on a specific driver that some boards don't have., only the board has the driver gets compiled and others boards will not.
Excluding unnecessary code will also reduce the compiled binary size.

### Changelist 
Add two parameter in `create_arm_binary` and `create_arm_binary_or_tests_for_board` functions in `firmware/CMakeList.txt`
Change the CMakeList.txt for each board accordingly.

### Testing Done
Remove cache in build_arm and build_x64 
Load Cmake test and embedded
Build all embedded binary, all test and shared test

### Checklist


Jiajun-Huang:
The issue in ‘scripts/utilities/supported_boards.py’ seems like just a Python version issue. I can remove the fix later.

firmware/thruna/BMS/CMakeLists.txt
```diff
@@ -50,6 +50,32 @@ list(REMOVE_ITEM GOOGLETEST_TEST_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/main
...
+set(SHARED_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../shared")
+file(GLOB_RECURSE SHARED_APP_SRCS "${SHARED_DIR}/src/app/*.c")
+file(GLOB_RECURSE SHARED_IO_SRCS "${SHARED_DIR}/src/io/*.c")
+file(GLOB_RECURSE SHARED_HW_SRCS "${SHARED_DIR}/src/hw/*.c")
+
+set(SHARED_APP_INCLUDE_DIRS "${SHARED_DIR}/src/app")
+set(SHARED_IO_INCLUDE_DIRS
+    "${SHARED_DIR}/src/io"
+    "${SHARED_DIR}/src/hw"
+)
```
Lucien950, line 63: can we set this in `firmware/CMakeLists.txt` rather than pass it all the way down the function? This saves a lot more code between all the files in `firemware/thruna/<BOARD>/CmakeLists.txt`

Lucien950: Because from what I can tell the variables are the same throughout.

Lucien950: By the way variables defined at higher levels of the folder structure carry over to lower levels.

Lucien950 (changes requested):

firmware/thruna/BMS/CMakeLists.txt
Jiajun-Huang: The primary I made this change is this.
I am currently writing an SD_card driver that is based on the HAL SD_Card driver for f4dev. Gus told me to put this in the shared folder and I think it makes sense because other boards might use this code in the future.  But, if I set it up like this, all other boards can't compile because they do not have the HAL driver for the SD card. 

Also, I am setting up a new board f4dev and I only want it to have HAL SD_Card driver on the project for now. But, if I want to build the binary, I have to add a bunch of other drivers and middleware to solve the dependency issue just to compile the shared code.  

So I want each individual board to have the ability to select the shared code it needs; so that other existing boards can include everything in shared code but exclude the SD_Card driver, and the new f4dev board can only include the SD_card driver and exclude anything else.

In order to do that, I think it has to be done in the CmakeLists.txt for each board.

Lucien950: Can you make it so that `SHARED_APP_SRCS`, `SHARED_IO_SRCS`, `SHARED_HW_SRCS` in `firmare/CMakeLists.txt` and set `ALL_SHARED_SRCS` and `ALL_SHARED_INCLUDE_DIRS` in each CMakeLists.txt?

Lucien950: So you pick everything you want from each board's CMakeLists.txt but you keep the helper variables in the higher context.

Jiajun-Huang: The changes are pushed.

Lucien950 (approved):
The code looks good. Please wait until @gtaharaedmonds reviews to merge to master.

firmware/CMakeLists.txt
```diff
@@ -139,13 +139,14 @@ function(create_arm_binary
...
+    ALL_ARM_SHARED_BINARY_INCLUDE_DIRS
+    ALL_ARM_SHARED_BINARY_SRCS
     STM32_DEVICE_TYPE
 )
     add_executable(${BOARD_NAME}.elf
         ${ARM_BINARY_SRCS}
         ${ARM_BINARY_THIRD_PARTY_SRCS}
-        ${SHARED_ARM_BINARY_X86_COMPATIBLE_SRCS}
-        ${SHARED_ARM_BINARY_X86_INCOMPATIBLE_SRCS}
+        ${ALL_ARM_SHARED_BINARY_SRCS}
```
gtaharaedmonds, line 149: These should probably be renamed to `SELECTED_ARM_BINARY_*`

firmware/thruna/BMS/CMakeLists.txt
```diff
@@ -50,6 +50,24 @@ list(REMOVE_ITEM GOOGLETEST_TEST_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/main
...
+    ${ALL_SHARED_HW_SRCS}
+)
+
+set(SELECTED_SHARED_INCLUDE_DIRS
+    ${ALL_SHARED_APP_INCLUDE_DIRS}
+    ${ALL_SHARED_IO_INCLUDE_DIRS}
+    ${ALL_SHARED_HW_INCLUDE_DIRS}
+)
+
+
```
gtaharaedmonds, line 70: nit: Can you make whitespace consistent on the board files? i.e. remove the double newlines

firmware/thruna/BMS/CMakeLists.txt
```diff
@@ -50,6 +50,24 @@ list(REMOVE_ITEM GOOGLETEST_TEST_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/Test/Src/main
...
+    ${ALL_SHARED_APP_SRCS}
+    ${ALL_SHARED_IO_SRCS}
+    ${ALL_SHARED_HW_SRCS}
+)
+
+set(SELECTED_SHARED_INCLUDE_DIRS
+    ${ALL_SHARED_APP_INCLUDE_DIRS}
+    ${ALL_SHARED_IO_INCLUDE_DIRS}
+    ${ALL_SHARED_HW_INCLUDE_DIRS}
+)
```
gtaharaedmonds, line 68: Is this unnecessary - is there ever a case where a certain board will want to ignore  a certain include directory? 
I think every board will want to look in `app`, `io`, and `hw`. I believe headers only get included in the binary if a C file `#include`s it so this shouldn't affect the binary size either. Can you think if we will ever need this?

gtaharaedmonds (commented):
Looks great. We should probably do this for tests too but that can be a follow up since the SD driver is higher priority

firmware/thruna/BMS/CMakeLists.txt
Jiajun-Huang: I think you are right. It should not be a problem to include more directories for `target_include_directories` than needed. I will make this as a variable in `firmware/Cmakelist.txt`, and remove this argument from the functions.

gtaharaedmonds (approved):
Nice!

PR #1019, opened 2023-10-11
Make CAN signal names unique

gtaharaedmonds:
### Summary

Currently CAN signal names are not unique. This is not a problem when using PCAN or writing code, since signals are categorized by their CAN message. However, this is not the case when uploading data to Influx, since then only signal names are uploaded. This means that there are weird conflicts between signals of the same name. For example, `State` for the BMS and the DCM.

This PR changes so signal names are prefixed by their board, and then we enforce all signals are unique across all boards. To avoid ridiculously long CAN setters/getters in the firmware, only the signal name is now used. 

For example:
|           | Before                                     | After                           |
|-----------|--------------------------------------------|---------------------------------|
| Message   | `BMS_Contactors`                           | `BMS_Contactors`                |
| Signal    | `AirPositive`                              | `BMS_AirPositive`               |
| TX Setter | `App_CanTx_BMS_Contactors_AirPositive_Set` | `App_CanTx_BMS_AirPositive_Set` |

The board name prefixes are also now added automatically to messages/signals, so they've been removed from the `*_tx.json` files.

### Changelist 

- Change `jsoncan` Python to support these changes
- Rename all signals
- Removed a few unused signals

Note: This diff is huge because of the autogenerated example code in `jsoncan`. I should really remove this and add proper unit tests instead.

### Testing Done

- [x] Validated on car

### Checklist


philam2001:
this'll be nice update, lots of potential merge conflicts pending tho lmao.

Lucien950 (approved):

PR #1022, opened 2023-10-13
Liam/commit info

liam-ilan:
### Summary
Every board exposes to the CAN bus the current running commit hash, and if the commit is clean.

### Changelist 
- Added python script utility to generate header files that expose the short git commit hash and if the commit is clean (`scripts/code_generation/commit_info_gen`).
- Modified `firmware/CMakeLists.txt` to generate header files with `commit_info_gen`. 
- Added `BOARDNAME_CommitInfo` message to every `BOARDNAME_tx.json`.
    - 32 bit `Hash`, 1 bit `Clean`.
- Every board broadcasts commit info through CAN bus.

### Testing Done
- Flashed on Car

### Resolved Issues
None

### Checklist


firmware/thruna/PDM/Src/main.c
```diff
@@ -52,6 +52,8 @@
 #include "configs/App_CurrentLimits.h"
 #include "configs/App_VoltageLimits.h"
 #include "configs/App_HeartbeatMonitorConfig.h"
+
+#include "App_CommitInfo.h"
 /* USER CODE END Includes */
 
 /* Private typedef -----------------------------------------------------------*/
```
Lucien950, line 59: PDM does not broadcast commit info, despite importing + creating message

firmware/CMakeLists.txt
```diff
@@ -469,6 +493,9 @@ function(create_arm_binary_or_tests_for_board
     list(APPEND ARM_BINARY_INCLUDE_DIRS
         "${AUTOGENERATED_CAN_IO_INCLUDE_DIRS}")
 
+    list(APPEND ARM_BINARY_INCLUDE_DIRS
+      "${TODO}")
```
Lucien950, line 497: remove/complete

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
```diff
@@ -0,0 +1,56 @@
...
+import git
+
+# validates data dict
+# data["hash"]: 7 char string with short commit hash
+# data["clean"]: "0" if dirty, "1" is clean
+def validateData(data):
+  return (
+    len(data["hash"]) == 7 # check length of hash
+    and all([c in string.hexdigits for c in data["hash"]])  # check hash is valid hexadecimal
+    and (data["clean"] == "0" or data["clean"] == "1") # check clean bit
```
Lucien950, line 14: instead of using comments, use 3 variable names.

Lucien950 (changes requested):
Works on my machine. Small things before we approve. Also, `DIM` does not broadcast commit info (I didn't know where to comment that, so I'm putting it here).

scripts/code_generation/commit_info_gen/.gitignore
```diff
@@ -0,0 +1,2 @@
+# temporary while we use test.h
+test.h
```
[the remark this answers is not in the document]
liam-ilan: Nope, good catch! I think it's also better to include an example header in the `commit_info` directory for reference, will fix.

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
liam-ilan: Fixed!

scripts/code_generation/commit_info_gen/.gitignore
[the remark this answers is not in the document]
liam-ilan: Fixed!

firmware/CMakeLists.txt
liam-ilan: Fixed!

firmware/thruna/PDM/Src/main.c
liam-ilan: Fixed! Good catch.

Lucien950:
formatting probably murks the head file because it is not used, mentioned in review

liam-ilan:
@Lucien950 
> Works on my machine. Small things before we approve. Also, `DIM` does not broadcast commit info (I didn't know where to comment that, so I'm putting it here).

I'll try to get CMake working to add the header for every board, rather than just the DIM, though I have a bit of an issue here...

Currently, as a quick fix, we just add the `App_CommitInfo.h` header file to `ARM_BINARY_SRCS`. Since this is a header file, it should actually be covered under `ARM_BINARY_INCLUDE_DIRS`. `ARM_BINARY_INCLUDE_DIRS` covers `Inc/App`, where `App_CommitInfo.h` is, however the commit info header file is not picked up... I'm not sure why...

Adding `App_CommitInfo.h` as a source, and not an include, works for all boards pre-refactor (everything but the `DIM`).

Does every autogenerated header file need an accompanying source file to be scanned as a dependency? In this case said source file would just include the header, and be empty... But if that's the fix, I can do it.

I'll try to experiment with this.

liam-ilan:
> formatting probably murks the head file because it is not used, mentioned in review

CI checks currently fail on the PDM on the `git diff`... Ran clang formatting... Unsure what the issue is...

Lucien950:
> > formatting probably murks the head file because it is not used, mentioned in review
> 
> CI checks currently fail on the PDM on the `git diff`... Ran clang formatting... Unsure what the issue is...

You have put the functions before the zone alloted for user code. Put it after this line
https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/a55c6cc64360a5b3048f37fee08c65c1a436e2c9/firmware/thruna/PDM/Src/main.c#L193

liam-ilan:
Everything should work on all boards, no exceptions! Next step is to test the software on the car itself...

firmware/CMakeLists.txt
```diff
@@ -19,6 +19,28 @@ elseif("${PLATFORM}" STREQUAL "x86")
     add_custom_target(all_tests)
 endif()
 
+# Generate header file for commit message
+function(commit_info_code_generation
+    HEADER_OUTPUT_PATH
+    SRC_OUTPUT_PATH
+)
+    set(GENERATE_COMMIT_INFO_SCRIPT_PY
+    ${SCRIPTS_DIR}/code_generation/commit_info_gen/src/generate_commit_info.py)
```
Lucien950, line 28: perhaps set this as a global variable, so it doesn't get regenerated on each run, less side effects

Lucien950: or possibly just inline it on L33 (but that would make it slightly less readable)

firmware/thruna/BMS/Src/main.c
```diff
@@ -275,6 +276,10 @@ int main(void)
     state_machine = App_SharedStateMachine_Create(world, App_GetInitState());
     App_AllStates_Init();
 
+    // broadcast
+
+    // broadcast
+
```
Lucien950, line 282: remove extra comments?

Lucien950 (changes requested):
other than a few small things to clean up, it is done

Lucien950 (approved):
once this is tested, this is solid

liam-ilan:
Current status:
- All boards but FSM work.
  - FSM Watchdog Times Out, need to fix. **DO NOT MERGE YET**
- PCAN displays the hash in decimal. Viewing raw hexadecimal results in a sort of reversed hash (eg. `9abcdef` becomes `ef cd ab 09`). I confirmed that the correct hash is being transmitted, this is a formatting issue.
  - Currently unsure if we can change the radix via the dbc file... If we can't, we need to stop PCAN from reversing raw hex data. If this is also not possible, the system is still super helpful, it's just a bit more clunky to read.

gtaharaedmonds:
> > > formatting probably murks the head file because it is not used, mentioned in review
> > 
> > 
> > CI checks currently fail on the PDM on the `git diff`... Ran clang formatting... Unsure what the issue is...
> 
> You have put the functions before the zone alloted for user code. Put it after this line (line 193)
> 
> https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/a55c6cc64360a5b3048f37fee08c65c1a436e2c9/firmware/thruna/PDM/Src/main.c#L190-L198
> 
> I would also suggest adding this function (in all boards) in the 1Hz task, in order to keep refreshing it, in case the CAN viewer is not plugged in when the board starts up. @gtaharaedmonds I believe that is the case, please fact check me on that one

No when you call a CAN setter it sets a static variable (the "CAN table"). This means if you just set the value once it will persist for the lifetime of the program. For that reason, I think you should move this back to `main.c`.

gtaharaedmonds:
> * FSM Watchdog Times Out, need to fix. **DO NOT MERGE YET**

FSM watchdog timeout has been fixed in https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/1038, I don't think is your change.

Okay that's too bad but this is still great to have. Even if its a bit more work to read its great that all of our data logs are now stamped with the commit. We could look back into doing this with strings in the future, but I don't think thats high priority.

scripts/code_generation/commit_info_gen/README.md
```diff
@@ -0,0 +1,46 @@
+# commit_info_gen
```
gtaharaedmonds, line 1: Actual documentation is great

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
```diff
@@ -0,0 +1,67 @@
...
+    # get commit info
+    repo = git.Repo(search_parent_directories=True)
+    commit = repo.head.commit
+    clean = not repo.is_dirty(untracked_files=True)
+
+    # short hash is the first 7 chars of the long hash
+    short_hash = commit.hexsha[0:7]
+
+    data["hash"] = short_hash
+    data["clean"] = "1" if clean else "0"
```
gtaharaedmonds, line 44: Nit: It'd probably be better to do this as a bool and import `<stdbool.h>`

Up to you, this is very minor

gtaharaedmonds (approved):
LGTM. Can you move the CAN setters back to `main.c`, and then let's merge!

scripts/code_generation/commit_info_gen/README.md
Lucien950: also on confluence?

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
Lucien950: I would prefer booleans in this case

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # validate
+    if not validateData(data):
+      raise Exception("Invalid data.")
+
+  except:
```
Lucien950, line 50: i would suggest catching the specific error here we are expecting in order to allow other errors to stop the process

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
liam-ilan: Now using booleans!

scripts/code_generation/commit_info_gen/README.md
gtaharaedmonds: imo its fine to stay in the repo since this is implementation details, I don't think another team will have to look at this file

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
```diff
@@ -0,0 +1,74 @@
...
+}
+
+# validates data dict
+# data["hash"]: 7 char string with short commit hash
+# data["clean"]: "0" if dirty, "1" is clean
+def validateData(data):
+  validLength = len(data["hash"]) == 7
+  validHex = all([c in string.hexdigits for c in data["hash"]])
+  validClean = data["clean"] == "false" or data["clean"] == "true"
+  return validLength and validHex and validClean
```
gtaharaedmonds, line 20: nit: Python convention is to use `snake_case` for variables and function names, and `PascalCase` for classes

can_bus/json/BMS/BMS_tx.json
```diff
@@ -323,5 +323,18 @@
                 "unit": "degC"
             }
         }
+    },
+    "BMS_CommitInfo": {
```
gtaharaedmonds, line 327: You can review the board name prefixes, they're added automatically now. 
This change just went in today so I still need to make a post about it

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
liam-ilan: Tried to make things a bit clearer in commit e55adf4249ba968141137c430e981a249d7c51a1.

Previously, one try catch would handle both GitPython and data validation. Now data validation is not in the try except, and the try except handles GitPython errors only.

Because commit_info_gen is non-critical, we want to stop it from blocking compilation as much as possible. To do this, if any error is released by GitPython, dummy data is returned instead, with a warning logged to the terminal. We don't care what type of error is thrown, all errors should be caught so the process can continue.

scripts/code_generation/commit_info_gen/src/generate_commit_info.py
liam-ilan: good catch, fixed!

can_bus/json/BMS/BMS_tx.json
liam-ilan: Fixed!

Lucien950 (approved):

PR #1025, opened 2023-10-13
Faults cause exit DCM drive state (not opening contactors)

gtaharaedmonds:
### Summary

The current fault handling behavior is to open contactors. This is super annoying (and would mean the end of competition) since the BMS latches whenever contactors open, and so needs a power cycle.

### Changelist 

Modified state machines so:

- If the BMS detects a fault, it doesn't do anything
- When the DCM detects a fault during drive state, it goes back to init state
- Init -> drive transition on the DCM is blocked if there is a fault

By faults here I mean the faults defined in the `*_alerts.json`.

### Testing Done

TBD, I need to think about this

### Checklist


firmware/thruna/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -85,17 +81,13 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
             App_SharedStateMachine_SetNextState(state_machine, next_state);
         }
 
-        if (external_shutdown_occurred)
+        if (external_shutdown_occurred && is_charger_connected)
         {
+            // TODO: Why do we need this transition?
```
will-chaba, line 86: I'm not 100% on this, but I think this allows for an external shutdown condition to transition to fault state BEFORE enabling the charger as part of ChargeStateRunOnEntry. This could be re-worked, but maybe save that for a separate PR.

firmware/thruna/BMS/Src/App/states/App_AllStates.c
```diff
@@ -157,21 +152,13 @@ bool App_AllStatesRunOnTick100Hz(struct StateMachine *const state_machine)
...
-    bool fault_from_other_board = (dcm_fault || fsm_fault || pdm_fault || dim_fault) && !ignore_other_boards;
-
     // Wait for cell voltage and temperature measurements to settle. We expect to read back valid values from the
     // monitoring chips within 3 cycles
     if (acc_meas_settle_count < NUM_CYCLES_TO_SETTLE)
     {
         acc_meas_settle_count++;
     }
-    else if (acc_fault || ts_fault || missing_hb || fault_from_other_board)
+    else if (acc_fault || ts_fault)
```
will-chaba, line 161: Finally no more missing_hb ❤️

firmware/thruna/DCM/Src/App/states/App_DriveState.c
```diff
@@ -63,14 +63,20 @@ static void DriveStateRunOnTick1Hz(struct StateMachine *const state_machine)
...
+    // All states module checks for faults, and returns whether or not a fault was detected.
+    const bool all_states_ok    = App_AllStatesRunOnTick100Hz(state_machine);
+    const bool start_switch_off = !App_IsStartSwitchOn();
+    const bool bms_not_in_drive = !App_IsBmsInDriveState();
+    bool       exit_drive       = !all_states_ok || start_switch_off || bms_not_in_drive;
+
+    if (all_states_ok)
     {
         App_SetPeriodicCanSignals_TorqueRequests();
+    }
```
will-chaba, line 75: Should we explicitly set torque request to zero? or is the absence of a torque request CAN message sufficient?

will-chaba (approved):
Looks good to me

firmware/thruna/BMS/Src/App/states/App_PreChargeState.c
```diff
@@ -51,10 +51,6 @@ static void PreChargeStateRunOnTick100Hz(struct StateMachine *const state_machin
             &precharge_fault_limit_exceeded);
         const bool external_shutdown_occurred = !App_Airs_IsAirNegativeClosed(airs);
 
-        struct HeartbeatMonitor *hb_monitor = App_BmsWorld_GetHeartbeatMonitor(world);
-        const bool               missing_hb = !App_SharedHeartbeatMonitor_Tick(hb_monitor);
-        App_CanAlerts_BMS_MissingHeartbeatFault_Set(missing_hb);
```
will-chaba, line 56: My removing the missing hb error we seem to be entering pre-charge for driving when trying to charge is LV is powered up before the 240v charger plug is connected. (is_charge_connected signal not received until 240v plugged in).

Could re-implement missing hb as a condition to not enter pre-charge for driving in init state (not a fault, but would prevent from entering pre-charge for driving conditions if other boards disconnected)

firmware/thruna/DCM/Src/App/states/App_DriveState.c
gtaharaedmonds: The run on exit function and the run on entry to init both set torque to zero

firmware/thruna/BMS/Src/App/states/App_PreChargeState.c
gtaharaedmonds: Ok sounds good, let's leave it for a follow up.

firmware/thruna/BMS/Src/App/states/App_PreChargeState.c
gtaharaedmonds: talked on Slack, this is OK to leave for a later date. @will-chaba will make a Jira ticket

PR #1027, opened 2023-10-13
Jugveer-Sandher/power_manager

Jugveer-Sandher:
### Summary
Created a PowerManger file for the PDM that can change the state of the efuses to shutdown or drive.

### Testing Done
I did some testing checking if the channel is open or closed based on the state passed into the PowerManager_SetState() func. Also checked that the PowerManager_Init() func sets the efuses to closed.

### Checklist


.vscode/settings.json
```diff
@@ -6,5 +6,9 @@
 			],
 			"url": "/can_bus/json/schema_tx.json"
 		}
-	]
+	],
+	"files.associations": {
+		"app_powermanager.h": "c"
+	},
+	"cmake.configureOnOpen": false
```
gtaharaedmonds, line 13: Can you revert these changes?

firmware/thruna/DIM/src/app/states/app_driveState.c
```diff
@@ -99,7 +99,7 @@ static void driveStateRunOnTick100Hz(struct StateMachine *const state_machine)
 
     const float min_cell_voltage = App_CanRx_BMS_CellVoltages_MinCellVoltage_Get();
 
-    if (missing_hb)
+    if (false)
```
gtaharaedmonds, line 102: revert this too

firmware/thruna/PDM/Inc/App/App_PowerManager.h
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Initalizes to the drive state (closed)
+ */
+void PowerManager_Init();
+
+/**
+ * Sets the Power Manager state
+ * @param state to open/close
+ */
+void PowerManager_SetState(PowerManagerState state);
```
gtaharaedmonds, line 20: Prefix app-level functions with `App_`

firmware/thruna/PDM/Src/Io/tempCodeRunnerFile.c
```diff
@@ -0,0 +1 @@
+#include "Io_Hal.h"
```
gtaharaedmonds, line 1: Delete this file

firmware/thruna/PDM/Src/App/App_powerManager.c
```diff
@@ -0,0 +1,46 @@
...
+            [EFUSE_CHANNEL_AIR] = true,
+            [EFUSE_CHANNEL_LVPWR] = true,
+            [EFUSE_CHANNEL_EMETER] = true,
+            [EFUSE_CHANNEL_AUX] = true,
+            [EFUSE_CHANNEL_DRS] = true,
+            [EFUSE_CHANNEL_FAN] = true,
+            [EFUSE_CHANNEL_DI_LHS] = true,
+            [EFUSE_CHANNEL_DI_RHS] = true,
+        }
+    }
```
gtaharaedmonds, line 32: Enabled channels from Joe:

- EMETER, AUX, and DRS should be always off
- AIR and LVPWR should be always on
- DI_LHS and DI_RHS should be on only in the drive state

firmware/thruna/PDM/Inc/App/App_PowerManager.h
```diff
@@ -0,0 +1,20 @@
+#include <stdbool.h>
+#include "Io_Efuse.h"
+
+typedef enum
+{
+    POWER_MANAGER_CONTRACTOR_OPEN,
+    POWER_MANAGER_CONTRACTOR_CLOSED,
```
gtaharaedmonds, line 7: Maybe rename `POWER_MANAGER_CONTRACTOR_OPEN` -> `POWER_MANAGER_STATE_SHUTDOWN`, and -> `POWER_MANAGER_CONTRACTOR_CLOSED` -> `POWER_MANAGER_STATE_DRIVE`. This is because there will be a few seconds before we close contactors where we'll enable high power to turn on the inverters.

firmware/thruna/PDM/Src/App/App_powerManager.c
```diff
@@ -0,0 +1,46 @@
...
+};
+
+void PowerManager_Init()
```
gtaharaedmonds, line 35: Make sure you call this init function somewhere in `main.c` for it to actually get run.

gtaharaedmonds (changes requested):
Nice, looks great!

lehroatic (commented):
Looks good, do address Gus's comments.

firmware/thruna/PDM/Test/Src/Test_PowerManager.cpp
gtaharaedmonds, line 1: I think you can delete this file. In the tests, all we do is test that the fake function returns the value we set it to (true or false). So we're basically testing the fake function library itself, rather than our app-level code. 

There's currently no app code here to test - let's delete this file now and write some tests after you've written some more stuff in future tickets.

But I do like that you're thinking about tests :)

gtaharaedmonds (changes requested):

firmware/thruna/PDM/Test/Src/Test_PowerManager.cpp
gtaharaedmonds: Other than that this looks good to go in

PR #1037, opened 2023-10-18
Shayan accelerator pedal

shayana18:
### Summary
Changed the accelerator pedal to provide a percentage value of the maximum torque of the motors to use based on the angle of the pedal. 

### Changelist 
- changed SAPPS and PAPPS math to use cosine law and output percent based on the extremity of the angle relative to the test position 
- created more robust calculation for offset value 
- added dead zone to ensure noise close to rest position does not cause undesired motor rotation 
- added remapping feature outside of dead zone to ensure acceleration does not jump when leaving the dead zone and scales from 0 to 100 
-
- 

### Testing Done
While no screenshots were taken, the following testing was performed:
- used debugger to ensure that both PAPPS and SAPPS display correct percentage value based on angles (tested in rest position which should output 0, fully pressed which should output 100 and gradual increases from unpressed to pressed which should gradually increase from 0 to 100 
- did edge case testing as due to a mechanical fault the pedal can get stuck at a position that is very slightly ahead of its rest position when being gently placed in rest position (this then would output a small percentage value) for which a dead zone was developed 
- used de-bugger to ensure that PAPPS and SAPPS output relatively the same percentage value (+/- 2 percentage) by checked outputted percentage values by both at different pedal positions (fine-tuned constants for this functionality) 

### Resolved Issues
- fixed incorrect math used previously (no angle calculation for PAPPS and incorrect cosine low usage for SAPPS)
- corrected constants (unpressed and pressed lengths of both PAPPS and SAPPS along with their constants and offset values)
- ensured that pedal percentage grew in-relation to angle
  
### Checklist
-  the math used to calculate the pedal angle 
- code efficiency 
- correct use of macros 
- correct offset calculation
- no redundant constants/ functions  
- piecewise function written for re-mapping

### DIFF
--- a/firmware/CMakeLists.txt
+++ b/firmware/CMakeLists.txt
@@ -245,6 +245,7 @@ function(create_arm_binary
         -Wl,-T ${ARM_LINKER_SCRIPT}
         --specs=nano.specs
     )
+    target_link_libraries(${BOARD_NAME}.elf m) # link math library
 
     set(HEX_FILE "${BOARD_NAME}.hex")
     set(BIN_FILE "${BOARD_NAME}.bin")
--- a/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
+++ b/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
@@ -3,12 +3,8 @@
 #include <stdbool.h>
 
 /**
- * @return returns absolute angle based on pedal position using cosine law
- */
-float calcAppsAngle(float cos_law_coefficent, float cos_law_denominator, float pot_len);
-
-/**
- * @return Percentage of press of accelerator pedal
+ * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle 
+ * @return pedal percentage based on pqrimary accelerator pedal sensor  
  */
 float Io_AcceleratorPedals_GetPapps(void);
 /**
@@ -17,10 +13,11 @@ float Io_AcceleratorPedals_GetPapps(void);
 bool Io_AcceleratorPedals_PappsOCSC(void);
 
 /**
- * Gets the percentage of the accelerator pedal
- * @return The reported percentage travel of the secondary accelerator pedal sensor
+ * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle 
+ * @return pedal percentage based on secondary accelerator pedal sensor  
  */
 float Io_AcceleratorPedals_GetSapps(void);
+
 /**
  * @return whether or not the secondary accelerator pedal sensor is open or short circuit
  */
--- a/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
+++ b/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
@@ -3,6 +3,11 @@
 #include <stdbool.h>
 
 /**
+ * @return returns absolute angle based on pedal position using cosine law
+ */
+float calcAppsAngle(float cos_law_coefficent, float cos_law_denominator, float pot_len);
+
+/**
  * @return Percentage of press of accelerator pedal
  */
 float Io_AcceleratorPedals_GetPapps(void);
--- a/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
+++ b/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
@@ -3,13 +3,13 @@
 #include <stdbool.h>
 
 /**
- * Performs initial calculations for the rest and max angle of PAPPS and SAPPS 
- * @return 
+ * Performs initial calculations for the rest and max angle of PAPPS and SAPPS
+ * @return
  */
 void Io_AcceleratorPedals_Init(void);
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle 
+ * Calculates pedal percentage by comparing the current angle to the max angle
  * @return pedal percentage based on primary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetPapps(void);
@@ -19,7 +19,7 @@ float Io_AcceleratorPedals_GetPapps(void);
 bool Io_AcceleratorPedals_PappsOCSC(void);
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle 
+ * Calculates pedal percentage by comparing the current angle to the max angle
  * @return pedal percentage based on secondary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetSapps(void);
--- a/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
+++ b/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
@@ -3,8 +3,8 @@
 #include <stdbool.h>
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle 
- * @return pedal percentage based on pqrimary accelerator pedal sensor  
+ * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle
+ * @return pedal percentage based on primary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetPapps(void);
 /**
@@ -13,8 +13,8 @@ float Io_AcceleratorPedals_GetPapps(void);
 bool Io_AcceleratorPedals_PappsOCSC(void);
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle 
- * @return pedal percentage based on secondary accelerator pedal sensor  
+ * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle
+ * @return pedal percentage based on secondary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetSapps(void);
 
--- a/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
+++ b/firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
@@ -3,7 +3,13 @@
 #include <stdbool.h>
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle
+ * Performs initial calculations for the rest and max angle of PAPPS and SAPPS 
+ * @return 
+ */
+void Io_AcceleratorPedals_Init(void);
+
+/**
+ * Calculates pedal percentage by comparing the current angle to the max angle 
  * @return pedal percentage based on primary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetPapps(void);
@@ -13,7 +19,7 @@ float Io_AcceleratorPedals_GetPapps(void);
 bool Io_AcceleratorPedals_PappsOCSC(void);
 
 /**
- * Calculates pedal percentage by comparing the current angle to the max angle relative to the rest angle
+ * Calculates pedal percentage by comparing the current angle to the max angle 
  * @return pedal percentage based on secondary accelerator pedal sensor
  */
 float Io_AcceleratorPedals_GetSapps(void);
--- a/firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
+++ b/firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
@@ -52,22 +52,26 @@
 #define DEAD_ZONE_PERCENT (10.0f)
 
 static float PAPPS_rest_angle;
-static float PAPPS_max_angle; 
+static float PAPPS_max_angle;
 
 static float SAPPS_rest_angle;
-static float SAPPS_max_angle; 
+static float SAPPS_max_angle;
 
 // max and min angle calculation for PAPPS/SAPPS
-void Io_AcceleratorPedals_Init(void){
+void Io_AcceleratorPedals_Init(void)
+{
+    PAPPS_rest_angle =
+        (acosf(PAPPS_COS_LAW_COEFFICIENT - (powf(PAPPS_LENGTH_UNPRESSED_MM, 2) / PAPPS_COS_LAW_DENOMINATOR)));
+    PAPPS_max_angle =
+        PAPPS_rest_angle -
+        (acosf(PAPPS_COS_LAW_COEFFICIENT - (powf(PAPPS_LENGTH_FULLY_PRESSED_MM, 2) / PAPPS_COS_LAW_DENOMINATOR)));
 
-    PAPPS_rest_angle = (acosf(PAPPS_COS_LAW_COEFFICIENT - (powf(PAPPS_LENGTH_UNPRESSED_MM, 2) / PAPPS_COS_LAW_DENOMINATOR)));
-    PAPPS_max_angle = PAPPS_rest_angle - (acosf(PAPPS_COS_LAW_COEFFICIENT - (powf(PAPPS_LENGTH_FULLY_PRESSED_MM, 2) / PAPPS_COS_LAW_DENOMINATOR)));
-
-    SAPPS_rest_angle = (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_UNPRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
-    SAPPS_max_angle = SAPPS_rest_angle - (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_FULLY_PRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
-
+    SAPPS_rest_angle =
+        (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_UNPRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
+    SAPPS_max_angle =
+        SAPPS_rest_angle -
+        (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_FULLY_PRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
 }
-
 
 static float calcAppsAngle(float cos_law_coefficent, float cos_law_denominator, float pot_len)
 {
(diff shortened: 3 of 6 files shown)


scripts/utilities/generate_cube_code.py
```diff
@@ -86,7 +86,7 @@ def generate_cubemx_code(board, ioc, codegen_dir, cubemx, log4j_properties):
 if __name__ == '__main__':
     # Parse arguments
     parser = argparse.ArgumentParser()
-    supported_boards = get_board_names()
+    supported_boards = [] #get_board_names()
```
Lucien950, line 89: I think this was supposed to be temporary? Please allow supported board check

Lucien950: I believe if you rebase get_board_names works properly

firmware/CMakeLists.txt
```diff
@@ -220,6 +220,7 @@ function(create_arm_binary
         -Wl,-T ${ARM_LINKER_SCRIPT}
         --specs=nano.specs
     )
+    target_link_libraries(${BOARD_NAME}.elf m) # link math library
```
Lucien950, line 223: I did not know that [m was the math library](https://stackoverflow.com/questions/34625627/how-to-link-to-the-c-math-library-with-cmake), very interesting namespace choice.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
 
-    return CLAMP(pedal_percentage, 0.0f, 100.0f);
-}
+    float pedal_percentage_raw = (pedal_angle / max_angle) * 100.0f;
+
+    if(pedal_percentage_raw <= DEAD_ZONE_PERCENT){
+        return 0;
+    }
+
+    float pedal_percentage = CLAMP(((100 / (100 - DEAD_ZONE_PERCENT)) * (pedal_percentage_raw - DEAD_ZONE_PERCENT)), 0.0f, 100.0f);
```
gtaharaedmonds, line 84: I'd do this in 2 lines to make it more readable
```c
pedal_percent_with_deadzone = ((100 / (100 - DEAD_ZONE_PERCENT)) * (pedal_percentage_raw - DEAD_ZONE_PERCENT);
return CLAMP(pedal_percent_with_deadzone, 0.0f, 100.0f); 
```

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
+    
+    return pedal_percentage;
+ }
 
-bool Io_AcceleratorPedals_PappsOCSC(void)
+
+bool Io_AcceleratorPedals_PappsOCSC(
+    void) // returns false if within range (no OC or SC) and true if it is outside the range (OS or SC is present)
```
gtaharaedmonds, line 91: Add comments like this to the `.h` file. Look through other `.h` files for an example of how we document functions.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
 
-    float secondary_angle = pedal_travel_angle * 180 / (float)M_PI + 16.3f;
-    return secondary_angle / 30.0f * 100.0f - SAPPS_DEADZONE_PERCENT;
+    float pedal_percentage_raw = (pedal_angle / max_angle) * 100.0f;
+
+    if(pedal_percentage_raw <= DEAD_ZONE_PERCENT){
+        return 0;
+    }
+
+    float pedal_percentage = CLAMP(((100 / (100 - DEAD_ZONE_PERCENT)) * (pedal_percentage_raw - DEAD_ZONE_PERCENT)), 0.0f, 100.0f);
```
gtaharaedmonds, line 116: optional suggestion: This logic could be moved to another help function `applyDeadzone` or something like that

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
-    return !(MIN_PEDAL_VOLTAGE <= pedal_voltage && pedal_voltage <= MAX_PEDAL_VOLTAGE);
+    float pedal_voltage = Io_Adc_GetChannelVoltage(ADC1_CHANNEL_2);
+    return !(PAPPS_MIN_V <= pedal_voltage && pedal_voltage <= PAPPS_MAX_V);
 }
 
+
 float Io_AcceleratorPedals_GetSapps(void)
 {
     // TODO implement IO functionality
     // Length calculated via voltage
```
gtaharaedmonds, line 101: Can remove the TODOs now

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
-#define SAPPS_DEADZONE_PERCENT (8.0f)
+
+// function created to calculate angles relative to rest angle (since we dont want a negative angle we will do rest pos - current pos)
+float Io_AcceleratorPedals_Angle_calc(float cos_law_coefficent, float cos_law_denominator, float pot_len, float max_len)
+{
+    float angle = (acosf(cos_law_coefficent - (powf(max_len, 2) / cos_law_denominator)))  -
+                  (acosf(cos_law_coefficent - (powf(pot_len, 2) / cos_law_denominator)));
+
+    return angle;
+}
```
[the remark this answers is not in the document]
gtaharaedmonds: Yeah renaming to `calcAppsAngle` or something would be better
This is used for APPs and SAPPs though which use different constants, don't you have to pass them in to use the same func for both pedals?

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
+
 // Raw voltage seen at the pin
-#define PAPPS_FULL_PRESSED_POT_V (2.86f)
-#define PAPPS_UNPRESSED_POT_V (1.082f)
+
+#define PAPPS_FULL_PRESSED_POT_V (2.61f)
+#define PAPPS_UNPRESSED_POT_V (0.68f)
+
+#define SAPPS_FULL_PRESSED_POT_V (3.20f)
+#define SAPPS_UNPRESSED_POT_V (1.18f)
```
gtaharaedmonds, line 28: Did you look into using the sensor datasheet to convert from voltage -> length?

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
     const float pot_len_mm    = PAPPS_RAW_VOLTAGE_TO_LEN_MM(pedal_voltage);
 
-    // Calculate pedal percentage
-    float pedal_percentage = (pot_len_mm - FULLY_UNPRESSED_POT_LENGTH_MM) /
-                                 (FULLY_PRESSED_POT_LENGTH_MM - FULLY_UNPRESSED_POT_LENGTH_MM) * 100.0f -
-                             PAPPS_DEADZONE_PERCENT;
+    float pedal_angle = Io_AcceleratorPedals_Angle_calc(
+        PAPPS_COS_LAW_COEFFICIENT, PAPPS_COS_LAW_DENOMINATOR, pot_len_mm, PAPPS_LENGTH_UNPRESSED_MM);
+    const float max_angle = Io_AcceleratorPedals_Angle_calc(
+        PAPPS_COS_LAW_COEFFICIENT, PAPPS_COS_LAW_DENOMINATOR, PAPPS_LENGTH_FULL_PRESSED_MM, PAPPS_LENGTH_UNPRESSED_MM);
```
gtaharaedmonds, line 76: The code in `Io_AcceleratorPedals_Angle_calc` is a pretty CPU-intensive operation (that's why we were seeing watchdog timeouts before we enabled the high-speed clock).  You're passing `PAPPS_LENGTH_UNPRESSED_MM` in twice, and so doing the angle calc on `PAPPS_LENGTH_UNPRESSED_MM` twice. Ideally you refactor this so you only do it once.

Maybe make it so `Io_AcceleratorPedals_Angle_calc` only does the absolute angle conversion, and then you can make them relative here in `Io_AcceleratorPedals_GetPapps`.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,127 @@
...
+#define DEAD_ZONE_PERCENT (10.0f)
 
-#define PAPPS_DEADZONE_PERCENT (10.0f)
-#define SAPPS_DEADZONE_PERCENT (8.0f)
+
+// function created to calculate angles relative to rest angle (since we dont want a negative angle we will do rest pos - current pos)
+float Io_AcceleratorPedals_Angle_calc(float cos_law_coefficent, float cos_law_denominator, float pot_len, float max_len)
+{
+    float angle = (acosf(cos_law_coefficent - (powf(max_len, 2) / cos_law_denominator)))  -
+                  (acosf(cos_law_coefficent - (powf(pot_len, 2) / cos_law_denominator)));
```
gtaharaedmonds, line 62: Hmm I would've thought that the `cos_law_denominator` should divide everything inside the `acosf` function

gtaharaedmonds (commented):
Nice!

gtaharaedmonds:
> This is a very quick review (I have a midterm tomorrow 😭). I will trust you on the measurements, and that you completed testing properly, and I will only review the code quality/logic. Defer to @gtaharaedmonds on the new autogenerated code and IOC config.

Yeah empirically the high speed clock just magically works now
We should do some driving with it enabled to make sure it doesn't break randomly but it seemed fine on the jacks

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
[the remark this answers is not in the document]
Lucien950: oh yes I agree. I think gus's review is good on this one.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
shayana18: I did, I wasn't exactly sure if the independent linearity is the change in V/m so I ended up sticking to my experimental values.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
shayana18: it does, the way the coefficient macros works is that it performs the division on what would be a and b of the cosine law within its calculation.

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
shayana18: I changed the function to just calculate the angle and then created a rest_angle variable inside PAPPS and SAPPS to make max and pot angles relative.

scripts/utilities/supported_boards.py
```diff
@@ -22,6 +22,7 @@ def get_board_names() -> Tuple[str]:
         filter(lambda x: path_is_folder_and_has_ioc_file(os.path.join(path_to_boards, x)), listdir(path_to_boards)))
     return supported_board_names
 
+def get_board_names() -> Tuple[str]:
```
Lucien950, line 25: rebase pls

Lucien950 (changes requested):

scripts/utilities/generate_cube_code.py
```diff
@@ -95,6 +95,6 @@ def generate_cubemx_code(board, ioc, codegen_dir, cubemx, log4j_properties):
     args = parser.parse_args()
     if args.board not in supported_boards:
         print('Error: Invalid board name. Valid options: ' + ' '.join(supported_boards))
-        sys.exit(1)
+        # sys.exit(1)
```
Lucien950, line 98: this needs to also be reverted. Honestly, just revert the entire file.

Lucien950 (changes requested):

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,132 @@
...
 
 float Io_AcceleratorPedals_GetPapps(void)
 {
-    // Length calculated via voltage
-    float       pedal_voltage = Io_Adc_GetChannelVoltage(ADC1_CHANNEL_2);
-    const float pot_len_mm    = PAPPS_RAW_VOLTAGE_TO_LEN_MM(pedal_voltage);
+    float pedal_voltage = Io_Adc_GetChannelVoltage(ADC1_CHANNEL_2);
+
+    // Length calculated via voltage reading from PAPPS
+
```
gtaharaedmonds, line 71: nit: Extra lines like these under comments (here and elsewhere) aren't necessary

firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
```diff
@@ -2,6 +2,11 @@
 
 #include <stdbool.h>
 
+/**
+ * @return returns absolute angle based on pedal position using cosine law
+ */
+float calcAppsAngle(float cos_law_coefficent, float cos_law_denominator, float pot_len);
+
```
gtaharaedmonds, line 9: It's best practice to only add the functions that you need in other files to the header. This keeps the header file clean (you only expose the stuff that other files care about, so its easier to read) and avoids name conflicts (if defined non-statically in the header file, no other function can have this name).

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,132 @@
...
+#define SAPPS_MAX_V (SAPPS_FULL_PRESSED_POT_V + 0.5f)
+
+#define DEAD_ZONE_PERCENT (10.3f)
+
+float calcAppsAngle(float cos_law_coefficent, float cos_law_denominator, float pot_len)
+{
+    float angle = (acosf(cos_law_coefficent - (powf(pot_len, 2) / cos_law_denominator)));
+
+    return angle;
+}
```
gtaharaedmonds, line 64: Mark this function as static

gtaharaedmonds (commented):
Looks great, just left a few minor nitpicks now. 
Once you've taken care of those + tested on hardware, let's merge.

firmware/thruna/FSM/Inc/Io/Io_AcceleratorPedals.h
shayana18: Ahh, I see good to know

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,125 @@
...
+        calcAppsAngle(PAPPS_COS_LAW_COEFFICIENT, PAPPS_COS_LAW_DENOMINATOR, PAPPS_LENGTH_UNPRESSED_MM);
 
-    // Calculate pedal percentage
-    float pedal_percentage = (pot_len_mm - FULLY_UNPRESSED_POT_LENGTH_MM) /
-                                 (FULLY_PRESSED_POT_LENGTH_MM - FULLY_UNPRESSED_POT_LENGTH_MM) * 100.0f -
-                             PAPPS_DEADZONE_PERCENT;
+    float pedal_angle = rest_angle - calcAppsAngle(PAPPS_COS_LAW_COEFFICIENT, PAPPS_COS_LAW_DENOMINATOR, pot_len_mm);
+
+    const float max_angle =
+        rest_angle - calcAppsAngle(PAPPS_COS_LAW_COEFFICIENT, PAPPS_COS_LAW_DENOMINATOR, PAPPS_LENGTH_FULLY_PRESSED_MM);
```
gtaharaedmonds, line 73: `rest_angle` and `max_angle` are constants since they're just a pure function of `#define`s, so we theoretically only need to calculate them once, instead of every time this function is called. If you define them as static variables in the global scope, you can make an `Io_AcceleratorPedals_Init` function that calculates their values. Then just remember to call this init function from `main.c`. This is probably worth it since we know this is an expensive operation (was causing watchdog timeouts before we increased the clock speed)

gtaharaedmonds (changes requested):
Looks great, 99% ready to go in. Just left one comment.

Lucien950 (approved):
once gus's changes go through I approve of this

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,139 @@
...
+        (acosf(PAPPS_COS_LAW_COEFFICIENT - (powf(PAPPS_LENGTH_FULLY_PRESSED_MM, 2) / PAPPS_COS_LAW_DENOMINATOR)));
 
-#define MIN_PEDAL_VOLTAGE (0.5f)
-#define MAX_PEDAL_VOLTAGE (3.1f)
+    SAPPS_rest_angle =
+        (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_UNPRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
+    SAPPS_max_angle =
+        SAPPS_rest_angle -
+        (acosf(SAPPS_COS_LAW_COEFFICIENT - (powf(SAPPS_LENGTH_FULLY_PRESSED_MM, 2) / SAPPS_COS_LAW_DENOMINATOR)));
+}
```
gtaharaedmonds, line 74: Use your help function for this, no need to copy paste something that we already have a helper for

firmware/thruna/FSM/Src/Io/Io_AcceleratorPedals.c
```diff
@@ -1,76 +1,139 @@
...
+#define SAPPS_MIN_V (SAPPS_UNPRESSED_POT_V - 0.5f)
+#define SAPPS_MAX_V (SAPPS_FULL_PRESSED_POT_V + 0.5f)
+
+#define DEAD_ZONE_PERCENT (10.0f)
+
+static float PAPPS_rest_angle;
+static float PAPPS_max_angle;
+
+static float SAPPS_rest_angle;
+static float SAPPS_max_angle;
```
gtaharaedmonds, line 58: variables are `snake_case`, even for acronyms

gtaharaedmonds (changes requested):

gtaharaedmonds (approved):
LGTM

PR #1040, opened 2023-10-21
Untrack generated STM32Cube drivers and middleware

gtaharaedmonds:
### Summary

We currently use STM32CubeMX to generate our HAL drivers and FreeRTOS independently for every board. This is dumb because we track hundreds of thousands of lines per board that we don't write ourselves. These HAL and FreeRTOS libraries are available as a GitHub repo, so we should subrepo them instead.

### Changelist 

- Subrepo STM32CubeMX F4 firmware package
- Significantly refactor CMake to build libraries instead of passing around massive lists of source files
- Also just deleted the GSM since its currently an F3 project but will be made with an F4, so there's really nothing there worth keeping around

### Testing Done

- [ ] Get some driving time (this is a massive change, so we should take in on track)

### Resolved Issues

SW-43

### Checklist


Lucien950 (commented):
RIP GSM
RIP my computer loading this diff

Lucien950 (commented):
Please investigate sparse checkout rather than forked submodule in order to be able to stay up to date with the latest repository. (Because if we need to update we need to update how we delete, rather than how we keep)

gtaharaedmonds:
> Please investigate sparse checkout rather than forked submodule in order to be able to stay up to date with the latest repository. (Because if we need to update we need to update how we delete, rather than how we keep)

Yeah sparse checkout is way better, didn't know this was a thing. Updated.

firmware/cmake/embedded.cmake
```diff
@@ -0,0 +1,285 @@
+include("${CMAKE_SOURCE_DIR}/firmware/cmake/shared.cmake")
```
Lucien950, line 1: I would argue that if you want shared code, just separately import it to prevent double running the code? There's only one other spot I see you using embedded without shared, and it seems like it shouldn't be too much head bandwidth to import both shared and embedded, especially since most of the meat is in shared, and embedded if you need embedded stuff

Lucien950 (approved):
- I don't exactly see how the submodule sparse checkout is working for `STM32CubeF4`
- GSM support looks good, post refactor we definately should have matrix refer to a global boards list, and make-release autodetect which artifacts have been uploaded
- I don't see any usages of consts defined in `shared/src/hw`, and `shared/src/io`, but I also have no idea how any of those work
- idk if I agree with `TARGET_EMBEDDED`, like it's nice to have if we are in a pinch, but in the way it is currently used I am very wary of comprimising the type system. If there are things that are in app code that would disconnect if we cut off Io implementation, I would argue that it should just stay in Io. (I realize you didn't make this change, but it caught my attention since it is the only thing left in firmware cmakelists.txt)
- Looks like more or less the same files are deleted from all of the boards, so I'll pass it

## CMake
- I def agree with the CMake changes, and I appriciate the compactness of the new functions
- I would also include GTest in the refactor you are doing with CMake (idk if we're dropping it or something? at least just move it now to keep it consistent and clear for the next system)
- I've added a few changes to conform the dependency management with the new system
- How will we handle JSONCAN for Dimos? I think I might need to rewrite jsoncan cmake functions? should we move some of it to the jsoncan folder under `scripts/code_generation`?
- New UMLs for CMake are very needed with all the functions floating around
- **__CPM DEPENDENCY PROBLEM__** I believe this is before @Jiajun-Huang added littlefs, which is destroyed in these changes. I would suggest adding a function which can fetch embedded libraries with CPM, or passing in the PACKAGE_SOURCE_DIR in as an argument to a library function.

> the changes look alright, I'll approve pending the one cmake include dependency loop change, and perhaps documentation.

PR #1055, opened 2023-10-30
Enable HardFault handler on all boards

gtaharaedmonds:
### Summary

On ARM cores, when you do something very very bad an interrupt is fired which is handled by each board's `HardFault_Handler`. When doing so, the core's registers are pushed to the stack. The previous leads have written some black magic to get these values viewable via a hooked up debugger. I enabled this handler callback for all boards, because it was previously only enabled on the BMS and DCM.

### Changelist 

Add hard fault handler to board `stm32f4xx_it.c` files.

### Testing Done

N/A, but this is only changing the PDM, FSM, and DIM with code that has been on the BMS and DCM for months. Plus, this only caused a change in behavior after a critical fault, so the risk is low.

### Checklist


will-chaba (approved):
Looks good, for testing at least check that wheels spin before merging (but I see your point about being low risk)

PR #1057, opened 2023-10-30
Add support for STM32H733xx

gtaharaedmonds:
### Summary

Support binaries for STMH733xx microcontrollers, and add a project for the H7 dev board.

### Changelist 

- Support ARM Cortex-M7
- Support STM33H7xx microcontrollers with SEGGER SystemView, FreeRTOS, and H7 HAL drivers
- Add project for H7 dev board

Note: CAN is not supported on H7 yet, since it ST's FD CAN HAL driver rather than ST's CAN HAL driver (which we use on the rest of our microcontrollers).

### Testing Done

- None

### Checklist


.vscode/launch.json
```diff
@@ -71,6 +71,20 @@
       "showDevDebugOutput": "none",
       "preLaunchTask": "Build: DIM.elf"
     },
+    {
+      "name": "Debug Embedded: h7dev",
```
will-chaba, line 75: Good for now, may need to re-think how to clearly list which boards are h7 vs F4 in the future

firmware/dev/h7dev/src/app/configs/App_SharedSignalConfig.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+struct H7DevWorld;
+#define World H7DevWorld
```
will-chaba, line 4: should this have the exact same contents as firmware/dev/h7dev/src/app/configs/App_SharedStateMachineConfig.h?

will-chaba (approved):
Looks good, small concerns notes

firmware/CMakeLists.txt
```diff
@@ -10,7 +10,14 @@ if("${TARGET}" STREQUAL "deploy")
     CPMAddPackage(
         NAME STM32CUBEF4
         GITHUB_REPOSITORY UBCFormulaElectric/STM32CubeF4
-        GIT_TAG 3e907f8
+        GIT_TAG master
```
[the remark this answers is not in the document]
gtaharaedmonds: yep good call

firmware/dev/h7dev/src/app/configs/App_SharedSignalConfig.h
gtaharaedmonds: just removed the shared code that needed these via cmake instead

PR #1058, opened 2023-11-01
Platform Independent JSONCAN

gtaharaedmonds:
### Summary

Make `jsoncan` platform independent. This is required so @Lucien950 can use it on the dashboard, without our STM32 firmware.

### Changelist 

To use `jsoncan` on any platform, you need to:
- Define the flag `THREAD_SAFE_CAN_PACKING` if you don't want to use the FreeRTOS thread safe guards
- Define a file called `Io_SharedCanMsg.h` that defines the struct: 
```c
struct CanMsg
{
    uint32_t std_id;
    uint32_t dlc;
    uint8_t  data[8];
};
```
- Call `Io_CanTx_Init` and pass the transmit function that actually sends the message (or pushes it to a queue to be sent later). The function returns void and takes a pointer to a `const struct CanMsg`. 

### Testing Done

None yet.

### Checklist


Lucien950 (approved):
lets go

PR #1082, opened 2023-11-11
Shayan/heartbeat fault fix

shayana18:
### Summary
Removed heartbeat delays, fixed start up settle fault count and added some safety measures for pedal angle calculation 

### Changelist 
- Removed Hal delays from BMS, PDM and FSM as they were causing missing heart beat faults at start up
- moved accumulator and traction system fault counting to after the settle period (this way faults in the settle period will not be recorded)
- added a clamping function to the input of acos in the calc angle function of Io_AcceleratorPedal.C as a safety measure 

### Resolved Issues
Extensive testing was done, if you would like to see the results please contact me and I will send you the CAN result screenshots (Will not include here because there are more than 10)

### Checklist
- syntax
- Clamp implementation


will-chaba (approved):
Make sure to test this on car, otherwise looks good

gtaharaedmonds (approved):
Nice!

PR #1083, opened 2023-11-11
CAN Bootloader

gtaharaedmonds:
### Summary

Create a CAN bootloader for STM32F412 targets so we can update their firmware over the CAN bus. A bootloader is a separate program running on the MCU that (unsurprisingly) handles:
- Booting: Can start the main program (usually called the "app", not to be confused with our app-level abstraction layer)
- Loading: Can update its app code

When you reset the microcontroller, instead of running the usual program it now runs the bootloader. Its then up to the bootloader to decide whether or not to start the app or update its code. A microcontroller is able to overwrite its own flash: So its only a problem of telling the microcontroller what data to write at what location.

### Changelist 

- Create STM32F4 CAN bootloader in `firmware/boot` directory. The bootloader handles updating the application code, and uses STM's cyclic redundancy check peripheral to verify that the application was programmed correctly. Also, it broadcasts its application state at 10Hz over CAN while in the bootloader.
- Created linker scripts for the app and bootloader in `firmware/linker` to place the programs in their respective locations in memory.
- Add Python script to parse and send the bootloader a binary over CAN. Unfortunately this has to be a separate repo, since WSL has no CAN support whatsoever. This can be used to update or erase boards' applications.
- Add shared `hw` drivers for FLASH memory and cyclic redundancy check peripherals

### Testing Done

- [x] Testing on all boards

### Checklist


Lucien950:
image
what's up with all these new targets?

Lucien950 (changes requested):
I couldn't highlight it in the review section cause it's a submodule, but can you also add the requirements.txt to the requiremnets.txt in the environment folder, and Pipenv?

Lucien950:
Also python command documentation pls because I want to have a Clion config for Python I can add here.

will-chaba (approved):
Fix tests and maybe add some newlines at end of files

can_bus/json/shared_enum.json
```diff
@@ -58,5 +58,10 @@
 		"RANGE_CHECK_OK": 0,
 		"RANGE_CHECK_UNDERFLOW": 1,
 		"RANGE_CHECK_OVERFLOW": 2
-	}
+	},
+    "BootStatus": {
+        "BOOT_STATUS_NO_APP": 0,
+        "BOOT_STATUS_APP_INVALID": 1,
+        "BOOT_STATUS_APP_VALID": 2
```
Lucien950, line 65: probably better to make happy state 0, and natural numbers error codes.

PR #1087, opened 2023-11-17
Track .ioc checksums to remove code generation from CI

gtaharaedmonds:
### Summary

We currently re-generate STM32CubeMX code during CI, which is flaky and slow. This is because the binary builds depend on a checksum of the `.ioc` files, which are generated during STM32CubeMX code generation. These are currently git ignored, forcing code regeneration in CI.

However, we can instead track the checksums, and add a new job to verify the `.ioc` matches the tracked checksum. This gives us the best of both worlds:
- We can verify that the `.ioc` file is up to date with the code generated by STM32CubeMX by comparing the checksums. If they don't match, you will need to re-generate code locally.
- CI is fast.

### Changelist 

- Track `.ioc` MD5 checksums
- Add H7 dev board (I forgot to do this before whoops)

### Checklist


.github/workflows/ci.yml
```diff
@@ -98,6 +95,7 @@ jobs:
           - DCM
           - DIM
           - f4dev
+          - h7dev
```
Lucien950, line 98: sneaky

.github/workflows/ci.yml
```diff
@@ -133,7 +131,7 @@ jobs:
         password: ${{ secrets.DOCKER_TOKEN }}
 
     env:
-      BUILD_DIR: build_x86
+      BUILD_DIR: build_fw_test
```
Lucien950, line 134: my b

firmware/dev/h7dev/src/cubemx/Src/main.c
```diff
@@ -23,6 +23,7 @@
 /* Private includes ----------------------------------------------------------*/
 /* USER CODE BEGIN Includes */
 #include "string.h"
+#include "hw_hardFaultHandler.h"
```
Lucien950, line 26: wait is this a separate?

firmware/dev/h7dev/src/cubemx/Src/stm32h7xx_it.c
```diff
@@ -22,7 +22,7 @@
 #include "stm32h7xx_it.h"
 /* Private includes ----------------------------------------------------------*/
 /* USER CODE BEGIN Includes */
-#include "Io_SharedHardFaultHandler.h"
+#include "hw_hardFaultHandler.h"
```
Lucien950, line 25: this one too is this part of the same change or different change?

.github/workflows/ci.yml
```diff
@@ -66,20 +68,15 @@ jobs:
...
-          echo "test: $GITHUB_WORKSPACE"
-          xvfb-run python3 scripts/utilities/generate_cube_code.py \
-            --board ${{ matrix.board }} \
-            --log4j_properties_output ~/.stm32cubemx/log4j.properties \
-            --ioc $GITHUB_WORKSPACE/${{ matrix.ioc_dir }}/${{ matrix.board }}.ioc \
-            --codegen_output_dir $GITHUB_WORKSPACE/${{ matrix.ioc_dir }} \
-            --cube_bin /usr/local/STM32CubeMX/STM32CubeMX
-
-      - name: Fix formatting
-        run: python3 scripts/clang_format/fix_formatting.py
```
Lucien950, line 82: no more fix formatting? does this mean that "Check for Differences" below is also not required?

Lucien950 (changes requested):

.github/workflows/ci.yml
Lucien950: Do we remove?

Lucien950: code format enforcement?

Lucien950 (approved):

PR #1099, opened 2023-12-01
Power Manager for PDM

Jugveer-Sandher:
### Summary
Created a power manager for the PDM sets the efuses.

### Testing Done
Tested on car

### Checklist
- [X] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [X] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).


firmware/dev/h7dev/src/cubemx/h7dev.ioc.md5
```diff
@@ -0,0 +1 @@
+a1f3069ca5520fefcdbf19ab412ed788
```
gtaharaedmonds, line 1: I believe the changes to the .ioc checksums were unintentional, can you revert?

firmware/thruna/PDM/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ file(GLOB_RECURSE APP_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/src/app/*.c")
 list(APPEND APP_SRCS 
     "${SHARED_APP_INCLUDE_DIR}/App_SharedStateMachine.c"
     "${SHARED_APP_INCLUDE_DIR}/App_SharedHeartbeatMonitor.c"
+    "${SHARED_APP_INCLUDE_DIR}/App_Timer.c"
```
gtaharaedmonds, line 9: Why was this added?

firmware/thruna/PDM/src/app/states/app_driveState.c
```diff
@@ -0,0 +1,49 @@
...
+
+static void driveStateRunOnTick100Hz(struct StateMachine *const state_machine)
+{
+    app_allStates_runOnTick100Hz(state_machine);
+    bool isCorrectState =
+        App_CanRx_BMS_State_Get() != BMS_DRIVE_STATE || App_CanRx_BMS_State_Get() != BMS_INVERTER_ON_STATE;
+
+    if (isCorrectState)
+    {
+        App_SharedStateMachine_SetNextState(state_machine, app_initState_get());
```
gtaharaedmonds, line 29: I thought the behavior we wanted was we just stay in drive state indefinitely? 
I don't think we took notes on this discussion, but we probably should have, my bad. How are you implementing it here?

firmware/thruna/PDM/src/app/states/app_initState.c
```diff
@@ -17,6 +21,11 @@ static void initStateRunOnTick1Hz(struct StateMachine *const state_machine)
 static void initStateRunOnTick100Hz(struct StateMachine *const state_machine)
 {
     app_allStates_runOnTick100Hz(state_machine);
+
+    if (App_CanRx_BMS_State_Get() == BMS_INVERTER_ON_STATE)
+    {
+        App_SharedStateMachine_SetNextState(state_machine, app_driveState_get());
+    }
```
gtaharaedmonds, line 28: To make this more robust, maybe also go to drive state if the BMS is PRECHARGE or DRIVE

firmware/thruna/PDM/test/test_stateMachine.cpp
```diff
@@ -0,0 +1,20 @@
...
+    EXPECT_EQ(PDM_INIT_STATE, App_CanTx_PDM_State_Get());
+}
+
+TEST_F(StateMachineTest, check_state_transition_from_init_to_drive)
+{
+    SetInitialState(app_initState_get());
+    App_CanRx_BMS_State_Update(BMS_INVERTER_ON_STATE);
+    LetTimePass(state_machine, 10);
+    EXPECT_EQ(PDM_DRIVE_STATE, App_CanTx_PDM_State_Get());
+}
```
gtaharaedmonds, line 20: These tests could be a bit more robust. 

1. There is no check that the PDM remains in its intended state, it could immediately transition to another state and the tests wouldn't catch that. A simple way to check this is just wait a large amount of time (ex. 1s) and then check that it is still in the intended state.
2. For the second test, you can't say for sure that it was your action of updating the BMS RX message that caused the state transition. Theoretically, it could have done that by itself. A way to make sure is again to just wait a large amount of time before setting your test condition, and then you can have more confidence it was actually you causing the transition.
3. You don't test that the power manager also changes its state. Maybe add a `app_powerManager_getState` function and use that to check the PDM is in the correct power state at all times.
4. We don't test the DRIVE -> INIT condition.

I'm being pedantic in this case, but in general I think these are good things to think about when writing tests so I thought I'd mention them.

gtaharaedmonds (changes requested):
Overall looks good, just a couple comments on the DRIVE->INIT transition and the tests. 
Also I think you accidentally updated + committed the `canup` submodule, pls revert.

firmware/thruna/PDM/CMakeLists.txt
Jugveer-Sandher: I thought I needed this for testing, for the LetTimePass() function???

firmware/thruna/PDM/src/app/states/app_initState.c
```diff
@@ -17,6 +21,14 @@ static void initStateRunOnTick1Hz(struct StateMachine *const state_machine)
...
 {
     app_allStates_runOnTick100Hz(state_machine);
+
+    bool isInverterOnOrDriveState =
+        App_CanRx_BMS_State_Get() == BMS_INVERTER_ON_STATE || App_CanRx_BMS_State_Get() == BMS_DRIVE_STATE;
+
+    if (isInverterOnOrDriveState)
+    {
+        App_SharedStateMachine_SetNextState(state_machine, app_driveState_get());
+    }
```
gtaharaedmonds, line 31: Can you also get precharge state into this list?

nit: snake case pls `inverter_on_or_drive_state`

gtaharaedmonds (commented):
One minor change and then LGTM

will-chaba (approved):
Gus's Comments resolved, Looks good to me

PR #1101, opened 2023-12-03
Mighty199/fault display code

Mighty199:
### Testing Done

No testing on Car yet but the code has been tested on the example generated files

### Checklist


Lucien950:
Idk if this is too much of an ask, but is it possible write a simple script to generate a text file (or even a PDF) from JSONCAN information about which error codes map to which faults?

Lucien950 (changes requested):
a few changes requested on DIM mostly, JsonCAN code generates properly on Windows ✅.  I have also requested a few JsonCAN changes, but it may be due to a misunderstanding with me and the requirements laid out by @gtaharaedmonds.

PR #1105, opened 2023-12-21
Bootloader on H7

gtaharaedmonds:
### Summary

Port our CAN bootloader to the STM32H733xx microcontroller, so it can be used with the BMS and VC on Quadruna.

### Changelist 

- STM32H7 bootloader
- Python changes to `canup` subrepo
- Rework CMake to make shared files "opt-in" vs. "opt-out" across the entire repo

### Testing Done

- Can update the binary on the H7 dev board on the bench
- CI

### Checklist


firmware/shared/src/app/App_SharedMacros.h
```diff
@@ -64,3 +64,5 @@ typedef struct \
 {\
     unsigned char _unused;\
 } name;
+
+#define FLOOR_TO_MULTIPLE(x, multiple) (x / multiple) * multiple
```
will-chaba, line 68: nit: macro function should be wrapped in brackets ()

will-chaba (approved):
Looks good

firmware/shared/src/app/App_SharedMacros.h
gtaharaedmonds: whoops looked like i never ended up using, deleted

PR #1106, opened 2023-12-28
Jugveer-Sandher/vc_setup

Jugveer-Sandher:
### Summary
Setup the VC.
This PR is pretty fat (longer than 500 lines) but I think it would be easier for the reviewer keeping everything inside one PR.

### Changelist 
Implemented: 
- fdcan and jsoncan
- freeRTOS
- watchdog
- testing
- hw_tasks
- basic states (allStates/init/drive)
- heartbeat and state machine

### Testing Done
No testing done but once back from break will test fdcan and jsoncan.

### Checklist
- [X] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [X] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*)


can_bus/json/VC/VC_alerts.json
```diff
@@ -0,0 +1,14 @@
...
+    "warnings_counts_id": 1307,
+    "faults_counts_id": 1308,
+    "warnings": [
+        "WatchdogTimeout",
+        "TxOverflow",
+        "RxOverflow"
+    ],
+    "faults": [
+        "DummyFault"
+    ]
```
gtaharaedmonds, line 13: I believe we want to make Quadruna JSONCAN its own separate JSONCAN subproject in its own subfolder, since otherwise there will be lots of conflicts with boards like BMS and FSM. also it means that when we're done with Thruna we can just delete its JSON without affecting Quadruna.

So the structure would look something like:
```md
can_bus/
    dbcs/
    thruna/
        BMS/
        ...
    quadruna/
        VC/
        ...
```

gtaharaedmonds: also right now we reserve a group of 100 CAN IDs per board. so for example the inverters use 0-99, BMS uses 100-199, DCM uses 200-299, etc. In CAN, lower message means higher priority, so the boards are rougly ordered in "importance".

I think we should probably stick with this for Quadruna for now, if only for organization, so put the VC in the 200-299 range. (0-99 are still inverters, 100-199 is still BMS, and then VC).

firmware/quadruna/VC/src/hw/hw_tasks.h
```diff
@@ -0,0 +1,27 @@
...
+typedef struct
+{
+    ADC_HandleTypeDef *      hadc1;
+    ADC_HandleTypeDef *      hadc3;
+    FDCAN_HandleTypeDef *    hfdcan1;
+    IWDG_HandleTypeDef *     hiwdg1;
+    const CanConfig *        can_config;
+    struct StateMachine *    state_machine;
+    struct HeartbeatMonitor *heartbeat_monitor;
+} HwTasksConfig;
```
gtaharaedmonds, line 19: It might be easier to just `extern` these in the `hw_tasks.c` file. The reason we inject the dependencies into shared modules is so they can be reused across different boards, but the tasks file is always going to be specific to a single board. So just using `extern` to get them from the `main.c` file will probably be less boilerplate.

firmware/quadruna/VC/src/hw/hw_tasks.h
```diff
@@ -0,0 +1,27 @@
...
+
+void hw_tasks_preinit();
+void hw_tasks_init(HwTasksConfig *hw_tasks_config);
+void hw_tasks_100hz(void *argument);
+void hw_tasks_cantx(void *argument);
+void hw_tasks_canrx(void *argument);
+void hw_tasks_1khz(void *argument);
+void hw_tasks_1hz(void *argument);
```
gtaharaedmonds, line 27: super nit: make the `pascalCase`, i.e. `_preInit`, `_canTx`, `_run100Hz`, etc.

gtaharaedmonds (changes requested):
a couple minor nitpicky things but the structure of this looks great, nice. 99% ready to merge.

gtaharaedmonds:
> a couple minor nitpicky things but the structure of this looks great, nice. 99% ready to merge.

oh also, can you add building the embedded and test targets into the GitHub Actions CI?

gtaharaedmonds (approved):
LGTM!

PR #1107, opened 2023-12-31
Bootloader Debug for CLion

Lucien950:
### Summary

Change up the Clion run configs.

Major TODO is to add support for the new symbol files. There does not seem to be support for GDB config settings in Clion, which means that we manually add symbol files?? Or maybe it auto detects it. I am not sure. @gtaharaedmonds 

### Changelist 

- Clion runConfigurations

### Testing Done

- [ ] Use this with real board

### Checklist


firmware/thruna/BMS/Src/App/App_Eeprom.c
```diff
@@ -129,7 +129,7 @@ EEPROM_StatusTypeDef App_Eeprom_UpdateSavedSocAddress(struct Eeprom *eeprom, uin
     }
     else
     {
-        *address = *address + 1;
+        (*address)++;
```
gtaharaedmonds, line 132: cheeky

gtaharaedmonds (approved):
ok !

firmware/thruna/BMS/Src/App/App_Eeprom.c
Lucien950: it wouldn't compile otherwise because 1 is not a uint16_t

Lucien950: idk how it got past CI

gtaharaedmonds: that's weird it might just be a difference in compiler version

Lucien950:
fixed in 774488f08e79941a485c2b735576e5f369e1ffde on #1147

PR #1110, opened 2024-01-07
Data persistence functionality with Firebase 

myung03:
### Summary
IMPORTANT -> remember to `npm i` to ensure all dependencies are updated (firebase was added). You can now save, load, and delete your live dashboards :)))

### Changelist 
- added Firebase
- added delete, save, and load functionality (ONLY FOR LIVE GRAPHS CURRENTLY) 
- added error handling within dashboard.tsx and visualize.tsx components for data persistence functionality  
- changed "livegraph.tsx"  to be able to accept an optional "signals" component to support loading dashboards 
- changed "websocket.tsx" to reflect this change in livegraph.tsx
- typed all new functions and files

### Footnotes 
- currently experiencing a bug in dashboard.tsx (see comment in file). Don't think this will be a persisting issue after putting some thought into it as live data will not need to be loaded (i.e we'll just have the signals and the dashboards will start spitting out live data automatically). If data does not show up in loaded dashboards just press submit and it will load <3 

### Checklist


software/tracksight/frontend/firebase.ts
forbesashli, line 1: it seems to work on my device, i suggest we move to a generalized key, using the contact or electrical email (maybe a software email) in the future

software/tracksight/frontend/src/app/components/dashboard.tsx
forbesashli, line 1: I agree that this can be addressed later, i did notice that my live data was 'not' working in that it did not trace live, however i could have been using it wrong as it has been a while since i have used the front end.

software/tracksight/frontend/src/app/components/live/livegraph.tsx
```diff
@@ -33,18 +33,20 @@ const DEFAULT_LAYOUT = {
     yaxis: {},
     legend: { "orientation": "h" },
 };
-const UPDATE_INTERVAL_MS = 100; // changes update speed
```
forbesashli, line 36: do we not need an update interval anymore?

forbesashli (commented):
UI is great, left a couple of comments, if you could address those, some we may not need to address right away. I think it is worth chatting in person first!!

software/tracksight/frontend/src/app/components/dashboard.tsx
myung03: live data feature is currently commented out (i think I commented it out because it was messing with retrieving the signals using websocket, don't remember entirely). Should work if you go into the code and uncomment

software/tracksight/frontend/src/app/components/live/livegraph.tsx
myung03: noticed it was not being used in the previous versions so I removed it but can probably readd later once we have an actual live data stream

software/tracksight/frontend/firebase.ts
myung03: yes maybe next time we meetup

Evanyl:
Are we adding the ability to edit dashboards later?

PR #1111, opened 2024-01-13
Stack monitoring for tasks on all boards

lehroatic:
### Summary
New CAN messages displaying percentage of stack remaining for different tasks on each board. There is now a flag that is set to true if stack usage is above 70 percent. This flag was previously always true.

### Changelist 

- Io_SharedStackWaterMark.c/.h: reused stack size calculation from Io_IsStackAboveWatermarkThreshold and added stack_remaining to the stack_watermark struct
- Io_stackWaterMark.c (for all boards): set stack_remaining for different flavors of tasks (i.e 1Hz, 100Hz, 1kHz, CanRx, etc...) to functions for setting the proper CAN message displaying percentages
- Io_stackWaterMark.c (for all boards): changed the warnings to be dependent on stack overflow conditions 

### Testing Done
- Tested and flashed code onto the car, all messages are visible

### Checklist


Lucien950 (approved):
looks pretty good

PR #1112, opened 2024-01-13
Shayan/Gps_and_RingQueue

shayana18:
### Summary
Enabled GPS and velocity measurement using the SBG Ellipse-n sensor. Refactored ring queue for sensor to use FreeRTOS  CMSIS v2 API.

### Changelist 
- JSON Can: New message for the sensor was added
- EllipseImu files: enabled position and velocity capability of the sensor.  Refactored ring queue to not really be a ring queue but is now instead an FreeRTOS Queue. 
- DCM Alerts: Added Queue overflow warning. 

### Testing Done
- Testing was done on the sensor to make sure it working as expected. No such testing was done on the code being committed as the sensor is not yet mounted on the car. 

### Checklist
- did something slightly sus where I call the warning set function in the IO file (I know by standard we want call anything communicating with CAN in APP but since the queue is initialized in the IO, I just called it here). Please review and let me know if its all good. 
- Ensure Queue set up is correct and is a proper implementation of the API 
-


cmake/CPM.cmake
Lucien950, line 1: i don't think this should exist

cpm-package-lock.cmake
Lucien950, line 1: neither should this

Lucien950 (changes requested):
few cmake things before real review

firmware/thruna/DCM/Inc/App/App_EllipseImu.h
```diff
@@ -20,6 +20,24 @@ typedef enum
...
+    GPS_ALT_ACC,
+
+    // Velocity Info
+    GPS_VEL_STATUS,
+    GPS_VEL_N,
+    GPS_VEL_N_ACC,
+    GPS_VEL_E,
+    GPS_VEL_E_ACC,
+    GPS_VEL_D,
+    GPS_VEL_D_ACC,
```
gtaharaedmonds, line 39: Can you add the `ELLIPSE_OUTPUT_` prefix? 

Since there is no namespacing in C, and to avoid naming conflicts, its a pretty popular convention to prefix enums with their type name.

firmware/thruna/DCM/Src/Io/Io_EllipseImu.c
```diff
@@ -245,6 +328,44 @@ static void Io_EllipseImu_ProcessMsg_Status(const SbgBinaryLogData *log_data)
     sensor_data.status_data.com_status     = log_data->statusData.comStatus;
 }
 
+/*
+ * Process and save a relevant GPS Velocity information.
+ */
+static void Io_EllipseImu_ProcessMsg_GpsVel(const SbgBinaryLogData *log_data)
+{
+    // 0 means solution computed, 1-3 means an error occured (see binry log for more detail)
+    sensor_data.gps_data.gps1_velocity.status = sbgEComLogGpsVelGetStatus(log_data->gpsVelData.status);
```
gtaharaedmonds, line 337: We should parse this in a new function, and then return it as an enum, so we can then broadcast it over CAN. If we can see something like `GPS_SOLUTION_COMPUTED` ON PCAN vs. just remembering 0 is the good state, that'd be great.

firmware/thruna/DCM/Src/Io/Io_EllipseImu.c
```diff
@@ -291,6 +416,12 @@ float Io_EllipseImu_GetSensorOutput(EllipseImuOutput output)
     return *(sensor_output_map[output]);
 }
 
+float Io_EllipseGps_GetSensorOutput(EllipseImuOutput output)
+{
+    assert(output < NUM_SBG_OUTPUTS);
+    return *(sensor_output_map[output]);
+}
```
gtaharaedmonds, line 423: What is this function for?

gtaharaedmonds (commented):
This looks solid, I'm happy to approve once its working on hardware.

firmware/thruna/DCM/Src/Io/Io_EllipseImu.c
shayana18: This just checks that you are trying too access a valid member of the enum (since NUM_SBG_OUTPUTS is the last index in  the enum)

gtaharaedmonds:
Merged in another PR, closed.

PR #1117, opened 2024-01-20
Logging message on CAN bus to SD card through file system

Jiajun-Huang:
### Summary
Implement the functionality of logging messages on the CAN bus to SD card through the file system.

### Changelist 
Add `io_io_canLogging` with an additional queue inside
Refactor `hw_can` for a more flexible receive callback function.

### Testing Done
Tested on the board. Will not overflow with 500ms period per message

### Checklist
- [ x ] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [ x ] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).


firmware/shared/src/hw/hw_can.c
gtaharaedmonds, line 1: I'm thinking that instead of passing in 2 separate callback functions, one for FIFO0 and one for FIFO1, we just pass in a single callback function that takes a `CanMsg*` as a parameter.

Users of this driver shouldn't ever need to care which FIFO a message comes from, so might as well simplify the configuration by only needing to pass in a single callback. It would also be more portable, for example if we ever want to support an MCU with a single FIFO or with 3 FIFOs for example.

What this means is that each interrupt would call `HAL_CAN_GetRxMessage(handle->can, rx_fifo, &header, msg->data)`, construct a `CanMsg`, and then pass it to single callback.

firmware/dev/h7dev/src/io/lfs_config.c
gtaharaedmonds, line 1: nitpick: I would recommend eventually adding the `io` prefix to these functions.
1. We do this in all other functions that we write, so might as well keep it up.
2. It's nice to distinguish between functions from `lfs` and functions that we wrote.

gtaharaedmonds:
No rush on my comments btw since this is mostly a proof-of-concept for Quadruna, but just wanted to get you a review earlier rather than later.

firmware/dev/h7dev/CMakeLists.txt
```diff
@@ -49,20 +55,20 @@ if("${TARGET}" STREQUAL "deploy")
...
-        "h7dev_boot.hex"
-        "${CMAKE_BINARY_DIR}/firmware/boot/h7boot/h7dev_boot.hex"
-    )
+    # embedded_image(
+    #     "h7dev"
+    #     "h7dev_app.hex"
+    #     "${CMAKE_CURRENT_BINARY_DIR}/h7dev_app.hex"
+    #     "h7dev_boot.hex"
+    #     "${CMAKE_BINARY_DIR}/firmware/boot/h7boot/h7dev_boot.hex"
+    # )
```
gtaharaedmonds, line 72: Now that the board is stable, could you revert disabling the bootloader? I think we should keep it on here for testing if necessary.

firmware/shared/src/hw/hw_can.h
```diff
@@ -9,16 +9,26 @@
...
 
+typedef void (*MsgReceivedCallback)(CanMsg *rx_msg);
+
 #ifdef CANFD
 // STM32 HAL CAN FD handle.
-typedef FDCAN_HandleTypeDef CanHandle;
+typedef struct
+{
+    FDCAN_HandleTypeDef *can;
+    MsgReceivedCallback  canMsgRecievecallback;
```
gtaharaedmonds, line 23: nitpick: variables should be snake case (`can_msg_receive_callback`).

firmware/shared/src/hw/hw_sd.h
```diff
@@ -47,6 +48,7 @@ SdCardStatus hw_sd_read(SdCard *sd, uint8_t *pdata, uint32_t block_addr, uint32_
  *
  */
 SdCardStatus hw_sd_write(SdCard *sd, uint8_t *pdata, uint32_t block_addr, uint32_t num_blocks);
+SdCardStatus hw_sd_write_dma(SdCard *sd, uint8_t *pdata, uint32_t block_addr, uint32_t num_blocks);
```
gtaharaedmonds, line 51: nitpick: parts of functions should be camel case (`hw_sd_writeDma`)

firmware/shared/src/io/io_canLogging.c
gtaharaedmonds, line 1: This file has a lot of commented out stuff, can you clean it up before we merge to master?

scripts/canup
gtaharaedmonds, line 1: Did you mean to change this?

firmware/dev/h7dev/src/io/io_lfs_config.h
```diff
@@ -0,0 +1,24 @@
...
+#define IO_LFS_READ_SIZE HW_DEVICE_SECTOR_SIZE *IO_LFS_READ_SIZE_FACTOR
+#define IO_LFS_PROG_SIZE HW_DEVICE_SECTOR_SIZE *IO_LFS_PROG_SIZE_FACTOR
+#define IO_LFS_LOOKAHEAD_SIZE HW_DEVICE_SECTOR_SIZE *IO_LFS_LOOKAHEAD_SIZE_FACTOR
+#define IO_LFS_CACHE_SIZE HW_DEVICE_SECTOR_SIZE *IO_LFS_CACHE_SIZE_FACTOR
+
+int io_lfs_read(const struct lfs_config *c, lfs_block_t block, lfs_off_t off, void *buffer, lfs_size_t size);
+int io_lfs_prog(const struct lfs_config *c, lfs_block_t block, lfs_off_t off, const void *buffer, lfs_size_t size);
+int io_lfs_erase(const struct lfs_config *c, lfs_block_t block);
+int io_lfs_sync(const struct lfs_config *c);
+int io_lfs_config(uint32_t block_size, uint32_t block_number, struct lfs_config *cfg);
```
gtaharaedmonds, line 24: nitpick: Function name prefix and file name should match. So this should be called `io_lfs.h/.c` to match.

gtaharaedmonds (changes requested):
A few nitpick-y comments, but otherwise LGTM.

PR #1123, opened 2024-01-27
IMD/PWM signal

KelmLelm:
### Summary

Got BMS to interpret IMD pwm signal that described its state.
**I will most likely just rewrite this on the new port instead of rebasing after we have tested**
### Changelist 

Changed cubemx config for TIM Channel 1 for capture and compare timer interrupt in NVIC settings. Then added some functionality for broadcasting frequencies on CAN. 
### Testing Done

Going to solder connection to IMD 3V3 on the BMS side and using a function generator to mock the PWM signals the IMD gives. For IMD_Normal pwm and IMD_SST on startup, the BMS can just be booted up to check for SST (30Hz) and just operating normally (10Hz)
### Checklist


.vscode/settings.json
```diff
@@ -10,4 +10,8 @@
 	"cmake.buildDirectory": "${workspaceFolder}/${env:BUILD_DIR}",
 	"cmake.configureOnEdit": false,
 	"cmake.configureOnOpen": true,
+	"files.associations": {
+		"typeinfo": "c",
+		"app_setperiodiccansignals.h": "c"
+	},
```
will-chaba, line 16: I don't think you meant to commit this

firmware/thruna/BMS/Src/App/App_SetPeriodicCanSignals.c
```diff
@@ -7,11 +8,16 @@ void App_SetPeriodicCanSignals_Imd(struct Imd *imd)
     App_CanTx_BMS_ImdDutyCycle_Set(App_Imd_GetPwmDutyCycle(imd));
 
     const struct Imd_Condition condition = App_Imd_GetCondition(imd);
-    App_CanTx_BMS_ImdCondition_Set((uint8_t)condition.name);
+    App_CanTx_BMS_ImdCondition_Set((ImdCondition)condition.name);
```
will-chaba, line 11: Maybe we should change the type of condition.name to ImdCondition so we don't need this cast? (and so there's only one enum as opposed to two currently)

will-chaba (changes requested):
Looks good! But let's try and use swap to only having one ImdCondition Enum

PR #1127, opened 2024-01-28
Serial Debugger Initial System

Steven141:
### Summary
Initial pipeline of the serial debugger implemented. Protocol buffer file (.proto) created for debug modules and corresponding python and c library files generated. Nanopb used to auto generate these library files so it was added as a submodule.

### Testing Done
Added test_debug_modules.py to test the python library. Successfully built BMS binaries after defining a variable of type GPIO that is defined in the generated c library. Further testing planned down the line.

### Checklist


.gitmodules
```diff
@@ -9,4 +9,7 @@
 	url = https://github.com/UBCFormulaElectric/dimos-assets.git
 [submodule "scripts/canup"]
 	path = scripts/canup
-	url = https://github.com/UBCFormulaElectric/canup.git
\ No newline at end of file
+	url = https://github.com/UBCFormulaElectric/canup.git
```
Steven141, line 12: not sure what's going on in this url diff

firmware/serial_debugger/debug_modules.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+message GPIO {
+  uint32 pin = 1;
+  bool value = 2;
+}
+
+message ATDC {
```
Steven141, line 8: Before this was named ADC. Got the following error when adding the following changes to the current PR. 

1. target_link_libraries("BMS_app.elf" "BMS_stm32cube" "BMS_can" "BMS_commit_info" "debug_modules") in firmware/thruna/BMS/CMakeLists.txt
2. #include "debug_modules.pb.h" in firmware/thruna/BMS/src/cubemx/Src/main.c
3. GPIO gpio = GPIO_init_zero; in main(void) of firmware/thruna/BMS/src/cubemx/Src/main.c

image

Seems like this is the problem: https://stackoverflow.com/questions/3720034/error-expected-before-token

Changing the name fixed it. Do we care to fix with previous naming?

Steven141: Changed to ADC_Channel

firmware/serial_debugger/debug_modules.proto
1mackenziekyle, line 1: Is this the file that decodes the protos on the C side?

firmware/serial_debugger/README.txt
```diff
@@ -0,0 +1,12 @@
+Serial Debugger (Using Nanopb)
+==============================
+
+See ../third_party/nanopb for more information on how to use nanopb.
```
gtaharaedmonds, line 4: Maybe drop a link here.

firmware/serial_debugger/debug_modules.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+message GPIO {
+  uint32 pin = 1;
+  bool value = 2;
+}
+
+message ADC_Channel {
+  float value = 1;
+}
```
gtaharaedmonds, line 10: nitpick: We use `PascalCase` for type names, i.e. `Gpio` and `AdcChannel`
Maybe also distinguish between `GpioWrite` and `GpioRead`, because eventually they're gonna be different, but fine for now since this is really just settting up the build system

firmware/serial_debugger/test_debug_modules.py
```diff
@@ -0,0 +1,31 @@
+from google.protobuf.json_format import MessageToDict
+try:
+    import debug_modules_pb2 as debug_modules
+except ModuleNotFoundError:
+    print("\n*** ERROR: Some python protobuff files are missing. Run generate_proto_py.sh and try again.\n")
```
gtaharaedmonds, line 5: Maybe drop note that this is just for testing

gtaharaedmonds (changes requested):
minor nitpick things but LGTM

gtaharaedmonds (approved):
yep

jpatch22 (approved):
LGTM!

Lucien950 (approved):
Gus and I have come to an agreement

firmware/serial_debugger/debug_modules.proto
Steven141: will add gpio write and read when adding the enums in next PR

.gitmodules
Evanyl, line 12: can we bring the new line back

firmware/serial_debugger/README.txt
```diff
@@ -0,0 +1,12 @@
+Serial Debugger (Using Nanopb)
+==============================
+
+See ../third_party/nanopb for more information on how to use nanopb or visit https://github.com/nanopb/nanopb.
+
```
Evanyl, line 5: can we make this markdown

firmware/serial_debugger/test_debug_modules.py
```diff
@@ -0,0 +1,34 @@
...
+
+"""
+This file is for testting the python encoding and decoding for the serial debugger.
+"""
+
+def encode_gpio(pin, write_value):
+    gpio = debug_modules.GPIO()
+    gpio.pin = pin
+    gpio.value = write_value
+    gpio_dic = MessageToDict(gpio) # example dic transformation
```
Evanyl, line 15: i don't like inline comments

Evanyl (changes requested):
a few nits!

Evanyl (approved):

PR #1133, opened 2024-01-31
Kevin/bms/imd pwm signal(new port)

KelmLelm:
### Summary

Got BMS to interpret IMD pwm signal that described its state.
### Changelist 

Changed cubemx config for TIM Channel 1 for capture and compare timer interrupt in NVIC settings. Also added some functionality for broadcasting frequencies on CAN. In particular, added the ability to display 0Hz signal from the IMD. Since the timer interrupts only triggers from a PWM signal, added a counter to keep track of how long it's been since the IMD sent a PWM signal to interpret. Other frequencies were broadcasted the same way as 10Hz and 20Hz. Cleaned up and synced some enums from JSON and the imd.h file.
### Testing Done

Soldered connection to IMD_HS 3V3 on the IMD and then set function generator to mock the frequencies and duty cycles for the bms to interpret. Tried at every frequency for the valid duty cycles and both stepping through the code and PCAN explorer showed the correct readings.
### Checklist


firmware/thruna/BMS/src/app/app_imd.c
```diff
@@ -64,20 +66,29 @@ ImdCondition app_imd_getCondition()
     ImdCondition condition;
     memset(&condition, 0, sizeof(condition));
 
-    condition.name = estimateConditionName(pwm_frequency);
+    uint8_t ticks_since_pwm_update = pwm_counter_tick();
+
+    if (ticks_since_pwm_update == PWM_TICKS_MAX)
```
will-chaba, line 71: I think adding a brief comment to explain the timeout would be a good idea

will-chaba (approved):
Looks good!

firmware/shared/src/hw/hw_pwmInput.c
```diff
@@ -2,13 +2,35 @@
...
+/*
+This is a software counter that is incremented everytime the status of the IMD
+is broadcasted to CAN and the counter resets everytime the pwm signal from the
+IMD is received and turned into an IMD state to broadcast. When the counter reaches
+255 it will result in a 0 hz reading from the pwm signal. Usually if the IMD was not
+sending a frequency, it would not trigger the interrupt that updates IMD state
+*/
+static uint8_t pwm_counter = 0;
+
+uint8_t hw_pwm_counter_tick(void)
```
will-chaba, line 14: By convention this should be called hw_pwmInput_counter_tick(void);
```suggestion
uint8_t hw_pwmInput_counter_tick(void);
```

firmware/thruna/BMS/src/app/app_imd.c
```diff
@@ -7,7 +7,9 @@
 #include "app_canTx.h"
 #include "app_utils.h"
 
+// PWM_TICKS_MAX also defined and explained in "hw_pwmInputs.c"
 #define IMD_FREQUENCY_TOLERANCE 2.0f
+#define PWM_TICKS_MAX 255
```
will-chaba, line 12: Maybe insteadjust use #define MAX_8_BITS_VALUE (uint32_t)(255) which already exists in app_utils.h?

firmware/thruna/BMS/src/app/app_imd.c
```diff
@@ -148,6 +160,11 @@ ImdCondition app_imd_getCondition()
     return condition;
 }
 
+uint8_t app_imd_pwmCounterTick()
```
will-chaba, line 163: There is no need for this function anymore, you can replace all isntances of app_imd_pwmCounterTick() with io_imd_pwmCounterTick()

firmware/thruna/BMS/src/app/app_imd.c
```diff
@@ -148,6 +160,11 @@ ImdCondition app_imd_getCondition()
...
+{
+    return io_pwm_counter_tick();
+}
+
 float app_imd_getPwmFrequency()
```
will-chaba, line 168: Same with this function, call Io functions directly instead

firmware/thruna/BMS/src/io/io_imd.c
```diff
@@ -4,28 +4,33 @@
...
 
 float io_imd_getDutyCycle(void)
 {
-    return hw_pwmInput_getDutyCycle(pwm_input);
+    return hw_pwmInput_getDutyCycle(&pwm_input);
 }
 
 void io_imd_inputCaptureCallback(TIM_HandleTypeDef *htim)
 {
-    if (htim == pwm_input->config->htim)
```
will-chaba, line 26: Perhaps move the timeout counter to here so that you don't have the HW not working issue. Have it reset anytime this input capture callback is triggered.

will-chaba (changes requested):
Looks good, but let's move that IMD specific timer out of a shared HW file and move it to io_imd.c

will-chaba (approved):
Looks Good!

PR #1137, opened 2024-02-05
Add bootloader to VC

gtaharaedmonds:
### Summary

Add bootloader to the VC. Had to enable a second CAN peripheral since the VC uses FDCAN1, whereas the H7 dev board used for testing uses FDCAN2.

### Changelist 

- Add bootloader to the VC
- Add blinky to VC
- Fix outdated commit info on FSM

### Testing Done

- Testing on bench, with real VC

### Checklist


can_bus/quadruna/Boot/Boot_tx.json
```diff
@@ -0,0 +1,20 @@
+{
+    "VC": {
+        "msg_id": 1219,
+        "cycle_time": 100,
+        "signals": {
+            "BMS_Status": {
```
will-chaba, line 6: Should be VC

gtaharaedmonds: fixed

will-chaba (approved):
Looks good

PR #1143, opened 2024-02-06
Serial Debugger pt2

Steven141:
### Summary
Serial Debugger is at the state where you can read/write gpio's and read from adc's from the VC. Currently only one gpio and adc has been supported and tested with the VC.

Example usage in test_serial_debugger.py

### Checklist


gtaharaedmonds (approved):
mildly sus but joe wants to test
LGTM

will-chaba:
> mildly sus but joe wants to test LGTM

Mildly sus indeed, seems to have broken load cmake: 
image

PR #1144, opened 2024-02-06
Shayan/GPS&RTOSQueue

shayana18:
…TOS queue implementation

### Summary
Same as prev PR just rebased on DCM refactor

### Testing Done
CAN Testing was done on the queue to see if it was receiving  and transmitting data. GPS still needs to be tested

### Checklist


firmware/thruna/DCM/src/io/io_sbgEllipse.h
```diff
@@ -7,23 +7,41 @@
 #include "hw_uart.h"
 #endif
 
-// Enum of all the values read from the SBG Ellipse N sensor
+// Enum of all the values read from the SBG Ellipse N sensortypedef enum
```
gtaharaedmonds, line 10: looks like typo here

firmware/thruna/DCM/src/io/io_sbgEllipse.c
```diff
@@ -89,9 +153,12 @@ static SbgErrorCode io_sbgEllipse_logReceivedCallback(
 static void io_sbgEllipse_processMsg_imu(const SbgBinaryLogData *log_data);
 static void io_sbgEllipse_processMsg_eulerAngles(const SbgBinaryLogData *log_data);
 static void io_sbgEllipse_processMsg_status(const SbgBinaryLogData *log_data);
+static void io_sbgEllipse_processMsg_gpsVel(const SbgBinaryLogData *log_data);
+static void io_sbgEllipse_processMsg_gpsPos(const SbgBinaryLogData *log_data);
 
 /* ------------------------- Static Function Definitions -------------------------- */
 
+static const sensor_msg_config msg_config = { .sensor_rx_overflow_callback = app_canTx_DCM_Warning_SbgRxOverflow_set };
```
gtaharaedmonds, line 161: It is a bit suspicious to define a new type for the config, but then hardcode it to be a specific function - why not just use the CAN function where you want the RX overflow callback?
Also it's best practice to keep all the CAN stuff in the app level so it can be tested. Maybe a better way of doing this is making a public getter that will return the overflow count, and then using that with all the other CAN stuff in the app file.

gtaharaedmonds (approved):
Sorry, forgot about this and thought it already went in - just a few minor comments, looks solid feel free to merge when addressed.

PR #1145, opened 2024-02-07
Rename Thruna targets

gtaharaedmonds:
### Summary

Rename Thruna targets to avoid conflicts with Quadruna boards with the same name.

### Changelist 

- Rename Thruna targets in VS Code config, CMake, and CI

### Testing Done

- Basic testing in VS Code + CI

### Checklist
*Please change `[ ]` to `[x]` when you are ready.*
- [x] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [x] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).

PR #1147, opened 2024-02-09
Fix CI/CD

Lucien950:
### Summary

Change up the Clion run configs.
Major TODO is to add support for the new symbol files. There does not seem to be support for GDB config settings in Clion, which means that we manually add symbol files?? Or maybe it auto detects it. I am not sure. @gtaharaedmonds

### Changelist 

- [x] Fix damage from #1145 
    - [x] Change name of deployed binaries
    - [x] Change name of CLion embedded build targets
- [x] Run DBC CD only when requested
   - [x] Delete Release task now needs to check if there was a release made in the first place
- [x] Single Line Release Deploy now Multiline 
- [x] Check for Protoc to Build System (Mandatory)
- [x] Delete all old tags

### Resolved Issues

- Fixes [SW-54](https://ubcformulaelectric.atlassian.net/jira/software/projects/SW/boards/20?selectedIssue=SW-54) and [SW-60](https://ubcformulaelectric.atlassian.net/jira/software/projects/SW/boards/20?selectedIssue=SW-60)

### Checklist

[SW-54]: https://ubcformulaelectric.atlassian.net/browse/SW-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SW-60]: https://ubcformulaelectric.atlassian.net/browse/SW-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/push_release/action.yml
```diff
@@ -21,45 +21,47 @@ inputs:
...
-      env:
-        GH_TOKEN: ${{ github.token }}
       run: |
-        gh release delete ${{ inputs.tag_name }} --repo UBCFormulaElectric/Consolidated-Firmware --cleanup-tag --yes
-      continue-on-error: true
-    - name: debug env
+        echo "BASE_REF: ${{ github.base_ref }} <- HEAD_REF: ${{github.head_ref}}"
+        echo "SHA: ${{github.SHA}}"
+        ls -l -a
+    - name: debug git
```
gtaharaedmonds, line 30: `debug env` and `debug name` aren't very descriptive, could pick a better name

firmware/boot/config.h
```diff
@@ -61,6 +61,15 @@
...
 
+#elif defined(quadruna_BMS)
+
+#define START_UPDATE_ID 1210
+#define UPDATE_ACK_ID 1211
+#define STATUS_10HZ_ID 1219
+#define BOOT_GPIO_PORT GPIOB
+#define BOOT_GPIO_PIN GPIO_PIN_5
+#define BOOT_FDCAN1
+
```
gtaharaedmonds, line 72: These CAN IDs need to be different than the VCs, this is how each bootloader is uniquely identified on the bus. 
The convention I've been running with is 11xx IDs are for Thruna, 12xx IDs are for Quadruna, and 13xx IDs are for dev boards. And then a range of 10 is reserved for each board in the correct 100 range. Also fixed the pin by looking at Altium schematics.

```
#define START_UPDATE_ID 1200
#define UPDATE_ACK_ID 1201
#define STATUS_10HZ_ID 1209
#define BOOT_GPIO_PORT GPIOC
#define BOOT_GPIO_PIN GPIO_PIN_13
```

gtaharaedmonds: Can you also add this to in `can_bus/quadruna/Boot` so this shows up in the DBC? (reference the VC for how to do it)

firmware/quadruna/BMS/CMakeLists.txt
```diff
@@ -116,22 +116,22 @@ if("${TARGET}" STREQUAL "deploy")
...
-    #     "quadruna_BMS_boot.hex"
-    #     "${CMAKE_BINARY_DIR}/firmware/boot/h7boot/quadruna_BMS_boot.hex"
-    # )
+     embedded_image(
+         "quadruna_BMS"
+         "quadruna_BMS_app.hex"
+         "${CMAKE_CURRENT_BINARY_DIR}/quadruna_BMS_app.hex"
+         "quadruna_BMS_boot.hex"
+         "${CMAKE_BINARY_DIR}/firmware/boot/h7boot/quadruna_BMS_boot.hex"
+     )
```
gtaharaedmonds, line 134: Looks like you need to change the linker script from `*app_only.ld` to `*app.ld`
You should test this on the BMS before merging.

scripts/utilities/generate_log4j_properties.py
```diff
@@ -0,0 +1,30 @@
...
+
+    log4j_properties = args.log4j_properties_output
+    log4j_properties_dir = os.path.dirname(log4j_properties)
+    if not os.path.exists(log4j_properties_dir):
+        os.makedirs(log4j_properties_dir)
+
+    # Generate a config file to make STM32CubeMX logging output less verbose
+    log4j_properties_f = open(log4j_properties, 'w')
+    log4j_properties_f.write(LOG4J_PROPERTIES)
+    log4j_properties_f.close()
```
gtaharaedmonds, line 30: Now that we're using this again, how do we use it?

gtaharaedmonds: I can test it on Mac

gtaharaedmonds (changes requested):
Various comments

scripts/utilities/generate_log4j_properties.py
Lucien950: this just runs at generate time

gtaharaedmonds:
lots of good cleanup tests here, thanks

scripts/utilities/generate_log4j_properties.py
Lucien950: it will create a specific config file that stm looks for in your directory structure

firmware/boot/config.h
Lucien950: is it boot_fdcan 1 or 2?

firmware/quadruna/BMS/CMakeLists.txt
Lucien950: ok test.

firmware/boot/config.h
gtaharaedmonds: FDCAN1
Whats the error?

Lucien950: https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/e011f4572e47f1699bf69ea79c56c96212fe5d86/firmware/boot/h7boot/cubemx/Src/main.c#L154

gtaharaedmonds (approved):
ok lgtm now feel free to merge when my comments are all addressed

can_bus/quadruna/Boot/Boot_tx.json
Lucien950, line 1: jk, just checked the blame i ain't never done that

PR #1151, opened 2024-02-10
Serial port demo

gtaharaedmonds:
### Summary

A quick example demoing serial communication with the VC. Something like this could be used for the serial debugger project for the hardware team.

### Changelist 

- Added a program to the VC that will echo any data received, in 12 byte chunks.
- Simple Python notebook to demo this.

### Testing Done

- Bench testing with VC, works reliably

### Checklist


gtaharaedmonds:
Demo has served its purpose

PR #1171, opened 2024-03-02
Kevin/bms/charging over can

KelmLelm:
### Summary

Implemented charging over CAN for THRUNA ONLY. Added BRUSA "board" in JSON files to send and receive messages. Took out code that sends signals to the BRUSA directly.

### Changelist 

- JSON files for BRUSA
- App states in thruna for init, charge and all states

### Testing Done

- Will test to see if we can receive signals from charger over CAN first
- Then we can try to send signals to the charger through the BMS
- Then we can try to start charging with a charging profile.

### Checklist


firmware/thruna/BMS/src/app/states/app_initState.c
```diff
@@ -45,17 +45,16 @@ static void initStateRunOnTick100Hz(void)
     {
         const bool air_negative_closed = io_airs_isNegativeClosed();
         const bool ts_discharged       = app_tractiveSystem_getVoltage() < TS_DISCHARGED_THRESHOLD_V;
+        const bool missing_hb          = app_heartbeatMonitor_checkFaults();
 
         if (air_negative_closed && ts_discharged)
         {
-            const bool charger_connected         = io_charger_isConnected();
+            const bool charger_connected         = app_canRx_BRUSA_OnStatus_get();
```
will-chaba, line 52: If the charger requires messages in order to exit it's idle state and send can messages, this won't work as intended. You'd have to send messages to the charger to see if it's alive. Stick to using io_charger_isConnected() until we can test this further

will-chaba (commented):
Looks good for now, will need to test further once battery is ready for charging

PR #1182, opened 2024-03-16
logfs

gtaharaedmonds:
### Changelist 

Added a custom filesystem designed for high bandwidth data logging to SD cards. The filesystem is power-loss resilient (won't corrupt itself due to random power loss) and wear-levelling (the number of writes to a single block is bounded). Also added Python bindings so filesystem images can be read from a laptop. See `firmware/logfs`. 

### Testing Done

- Software tests on a laptop using Pytest, using both a fake disk (i.e. disk in RAM) and a real SD card (pass `--disk` to `pytest` invocation).
- Tested CAN message logging on the VC. Could record on the order of 11k CAN messages per second.


Lucien950 (approved):
image

PR #1185, opened 2024-03-16
Efuse Retry Protocol

Jugveer-Sandher:
### Summary
Implemented a efuse retry protocol for when a efuse's current drop.

### Testing Done
Created unit testing for both successful and unsuccessful protocol.

### Checklist


firmware/thruna/PDM/src/app/app_powerManager.c
```diff
@@ -1,49 +1,131 @@
...
-            [EFUSE_CHANNEL_DI_RHS] = false,
+        .retry_configs = {
+            [EFUSE_CHANNEL_AIR] = {true, RETRY_PROTOCOL_AIR, 1},
+            [EFUSE_CHANNEL_LVPWR] = {true, RETRY_PROTOCOL_LV, 1},
+            [EFUSE_CHANNEL_EMETER] = {false, RETRY_PROTOCOL_EMETER, 3},
+            [EFUSE_CHANNEL_AUX] = {false, RETRY_PROTOCOL_AUX, 3},
+            [EFUSE_CHANNEL_DRS] = {false, RETRY_PROTOCOL_DRS, 3},
+            [EFUSE_CHANNEL_FAN] = {false, RETRY_PROTOCOL_FANS, 3},
+            [EFUSE_CHANNEL_DI_LHS] = {false, RETRY_PROTOCOL_DI_LHS, 1},
+            [EFUSE_CHANNEL_DI_RHS] = {false, RETRY_PROTOCOL_DI_RHS, 1},
```
gtaharaedmonds, line 24: Consider using designated initializers (label each variable when you initialize a struct) since it's not clear what each parameter means unless you manually open `app_retryHandler.h`. 

For example, this snippet for the efuses is a lot more readable:
```c
static const EfuseConfig efuse_configs[NUM_EFUSE_CHANNELS] = {
    [EFUSE_CHANNEL_AIR] = {
        .enable_gpio = {
            .port = EN_AIR_GPIO_Port,
            .pin = EN_AIR_Pin
        },
        .stby_reset_gpio = {
            .port = FR_STBY_AIR_LVPWR_GPIO_Port,
            .pin = FR_STBY_AIR_LVPWR_Pin,
        },
        .cur_sns_adc_channel = ADC_1_CHANNEL_9,
    },
```

firmware/thruna/PDM/src/app/app_powerManager.c
```diff
@@ -1,49 +1,131 @@
...
-        io_efuse_setChannel(efuse, power_states_config[state].efuses[efuse]);
+        io_efuse_setChannel(efuse, power_states_config[state].retry_configs[efuse].efuse_state);
+    }
+}
+
+void app_powerManager_check_efuses(PowerManagerState state)
+{
+    int efuse = 0;
+
+    while (efuse < NUM_EFUSE_CHANNELS)
```
gtaharaedmonds, line 67: for loop is probably a better choice here

firmware/thruna/PDM/src/app/app_powerManager.c
```diff
@@ -1,49 +1,131 @@
...
+                bool is_efuse_faulting =
+                    efuse_retry_config->efuse_state && io_efuse_getChannelCurrent(efuse) < FAULT_CURRENT_THRESHOLD;
+                if (is_efuse_faulting)
+                {
+                    efuse_retry_data->retry_state = RETRY_STATE_RUNNING;
+                    efuse_retry_data->current_sum += io_efuse_getChannelCurrent(efuse);
+                    efuse_retry_data->timer_attempts++;
+                }
+                break;
+            }
```
gtaharaedmonds, line 120: Some short comments explaining what each state means would be much appreciated.

firmware/thruna/PDM/src/app/app_powerManager.c
```diff
@@ -1,49 +1,131 @@
...
 };
 
+void app_powerManager_init()
+{
+    for (int efuse = 0; efuse < NUM_EFUSE_CHANNELS; efuse++)
+    {
+        retry_data[efuse].retry_state          = RETRY_STATE_OFF;
+        retry_data[efuse].retry_attempts       = 0;
+        retry_data[efuse].timer_attempts       = 0;
+        retry_data[efuse].timer_attempts_limit = 3;
```
gtaharaedmonds, line 48: I think it'd be nice for this to be a config parameter too. 
(If I understand correctly, this is the number of cycles over which you average the current?)

Also, I think this should be a lot higher, I believe this value is 3 x 10ms = 30ms. We want on the order of a few seconds to be certain.

firmware/thruna/PDM/src/app/app_powerManager.c
```diff
@@ -1,49 +1,131 @@
...
+            case RETRY_STATE_EXPIRED:
+            {
+                float avg = efuse_retry_data->current_sum / (float)efuse_retry_data->timer_attempts_limit;
+                if (avg <= FAULT_CURRENT_THRESHOLD)
+                {
+                    if (efuse_retry_data->retry_attempts == efuse_retry_config->retry_attempts_limit)
+                    {
+                        app_canTx_PDM_EfuseFault_set(true);
+                        return;
+                    }
```
gtaharaedmonds, line 97: You can probably have this check at the top of the for loop, since there's really no need to stay do the rest of the checks if its already expired.

firmware/thruna/PDM/src/app/app_powerManager.h
```diff
@@ -1,5 +1,11 @@
 #include <stdbool.h>
 #include "io_efuse.h"
+#include "app_timer.h"
+#include "app_retryHandler.h"
+
+#pragma once
+
+#define FAULT_CURRENT_THRESHOLD 0.5f
```
gtaharaedmonds, line 8: This should also be a config parameter, since it'll be difference for every efuse.

firmware/thruna/PDM/src/app/app_retryHandler.c
```diff
@@ -0,0 +1,147 @@
...
+    {
+        // turn off lv and inverters
+        case RETRY_PROTOCOL_LV:
+            io_efuse_setChannel(EFUSE_CHANNEL_LVPWR, false);
+            io_efuse_setChannel(EFUSE_CHANNEL_DI_LHS, false);
+            io_efuse_setChannel(EFUSE_CHANNEL_DI_RHS, false);
+
+            retry_data[EFUSE_CHANNEL_LVPWR].retry_state  = RETRY_STATE_RUNNING;
+            retry_data[EFUSE_CHANNEL_DI_LHS].retry_state = RETRY_STATE_WAITING;
+            retry_data[EFUSE_CHANNEL_DI_RHS].retry_state = RETRY_STATE_WAITING;
```
gtaharaedmonds, line 27: Why do you need to update the states here?

firmware/thruna/PDM/src/app/app_retryHandler.c
```diff
@@ -0,0 +1,147 @@
...
+            io_efuse_standbyReset(EFUSE_CHANNEL_AIR);
+
+            io_efuse_setChannel(EFUSE_CHANNEL_DI_LHS, true);
+            io_efuse_setChannel(EFUSE_CHANNEL_DI_RHS, true);
+            io_efuse_setChannel(EFUSE_CHANNEL_AIR, true);
+
+            retry_data[EFUSE_CHANNEL_DI_LHS].retry_state = RETRY_STATE_OFF;
+            retry_data[EFUSE_CHANNEL_DI_RHS].retry_state = RETRY_STATE_OFF;
+            retry_data[EFUSE_CHANNEL_AIR].retry_state    = RETRY_STATE_OFF;
+            break;
```
gtaharaedmonds: The channels should already be enabled if we're calling standby reset on them right, so re-enabling their channel should be unecessary?

firmware/thruna/PDM/src/app/app_retryHandler.h
```diff
@@ -0,0 +1,43 @@
...
+    RetryState retry_state;
+    int        retry_attempts;
+    int        timer_attempts;
+    int        timer_attempts_limit;
+    float      current_sum;
+} RetryData;
+
+void retry_handler_start(RetryProtocol protocol, const RetryConfig retry_configs[], RetryData retry_data[]);
+
+void retry_handler_recover(RetryProtocol protocol, const RetryConfig retry_configs[], RetryData retry_data[]);
```
gtaharaedmonds, line 43: Not sure I understand how these 2 functions work. It seems like when we've decided an efuse is faulted, we call `retry_handler_start`, which looks like in a lot of cases it just disables the efuses.

Then there is `retry_handler_recover`, which we call if we decide the efuses aren't faulted, which is the standby reset we've talked about. However, I would think that if we decide the efuses aren't faulted, we shouldn't do anything?

I think I'm probably missing something, maybe we should talk about this on Monday.

gtaharaedmonds: Also nitpick: `app_retryHandler_start` and `app_retryHandler_recover`

gtaharaedmonds (changes requested):
Nice progress! Had a few questions, maybe we should sync up about this on Monday.

firmware/thruna/PDM/src/app/app_powerManager.c
Jugveer-Sandher: I made `timer_attempts_limit = 200` so 200 * 10ms = 2s is that enough time?

Jugveer-Sandher:
Confluence Page: https://ubcformulaelectric.atlassian.net/wiki/spaces/UFE/pages/277643278/Software+-+Thruna+FW+-+Efuse+Protocol

PR #1186, opened 2024-03-17
Port Thruna PDM to VC

liam-ilan:
### Summary
Ported Thruna PDM to VC, and added shutdown loop checks.

### Changelist 
- `efuse`, `lowVoltageBattery`, and `shutdown` io layers added to VC.
- `lowVoltageBattery`, and `shutdown` app layers added to VC.
- Transferred tests from PDM to VC.

### Testing Done
- Software tests pass.
- To be flashed on VC along with Amir's DCM port.

### Checklist
**This PR only includes PDM ported code, should be kept in one PR**


can_bus/quadruna/VC/VC_alerts.json
```diff
@@ -1,8 +1,8 @@
 {
-    "warnings_id": 205,
-    "faults_id": 206,
-    "warnings_counts_id": 207,
-    "faults_counts_id": 208,
+    "warnings_id": 296,
+    "faults_id": 297,
+    "warnings_counts_id": 298,
+    "faults_counts_id": 299,
```
gtaharaedmonds, line 5: What motivated changing all of the CAN message IDs?

gtaharaedmonds: (Here and in `VC_tx.json`)

firmware/boot/f4boot/cubemx/f4boot.ioc.md5
gtaharaedmonds, line 1: You should revert delelting all these `.md5` files.

can_bus/quadruna/VC/VC_alerts.json
```diff
@@ -20,6 +20,10 @@
         "MissingFSMHeartbeat",
         "MissingDIMHeartbeat",
         "LeftInverterFault",
-        "RightInverterFault"
+        "RightInverterFault",
+        "LEStopFault", 
+        "REStopFault",
+        "TSMSFault",
+        "PCMInterlockFault"
```
gtaharaedmonds, line 27: Can you make these warnings instead? In theory this doesn't matter because if any of these faults are set, the shutdown loop is open, and so we don't have HV and obviously can't drive.

However, with any sensor, there is always the possibility of sensor error, transients, noise, etc. which throw off your measurement. If that happened while driving, we wouldn't want to unnecessarily throw a fault and shutdown the car. 

On the other hand, if there is a true shutdown loop fault, the code doesn't need to do anything to make the car safe, because the shutdown loop will have already opened. So, there's really no downside of just making these warnings.

gtaharaedmonds: (For anyone reading this who doesn't know, faults cause the vehicle to exit drive state (no torque from motors), and warnings don't do anything (just for logging))

can_bus/quadruna/VC/VC_tx.json
```diff
@@ -13,7 +13,7 @@
         }
     },
     "LeftInverterCommand": {
-        "msg_id": 32,
+        "msg_id": 201,
```
gtaharaedmonds, line 16: These message IDs are actually special, which is why they're not in the usual 200s range. The inverter needs to know the specific ID of its command message, which is currently programmed to be 32/82. 

This is also why we're so specific in the signal config, because we have to conform exactly to our inverter's spec.

firmware/quadruna/VC/src/app/states/app_allStates.c
```diff
@@ -1,5 +1,21 @@
...
-void app_allStates_runOnTick100Hz(void) {}
+void app_allStates_runOnTick100Hz(void)
+{
+    app_lowVoltageBattery_broadcast();
+    app_shutdown_broadcast();
+
+    // TODO: add heartbeat monitor
+    // app_heartbeatMonitor_checkIn();
+    // app_heartbeatMonitor_tick();
+    // app_heartbeatMonitor_broadcastFaults();
```
gtaharaedmonds, line 20: Is this still a TODO, or can we do it now?

firmware/quadruna/VC/src/io/io_shutdown.h
```diff
@@ -0,0 +1,47 @@
...
+ * Get the status of the REStop shutdown pin
+ * @return the status of the REStop pin set in the shutdown config
+ */
+bool io_shutdown_hasREStopFault(void);
+
+/*
+ * Initialize config for shutdown gpio pins
+ * @param config with shutdown gpio pins
+ */
+void io_shutdown_init(const ShutdownConfig *shutdown_config);
```
gtaharaedmonds, line 47: nitpick: put `_init` function as first function.

firmware/quadruna/VC/src/hw/hw_adc.h
```diff
@@ -38,7 +38,8 @@ typedef enum
     ADC1_IN19_VBAT_SENSE,
     ADC3_IN0_AUX_PWR_I_SNS,
     ADC3_IN1_PUMP_PWR_I_SNS,
-    NUM_ADC_CHANNELS
+    NUM_ADC_CHANNELS,
+    NO_ADC // for cases where configuration requires an ADC pin, but there is none (TELEM/BUZZER efuses)
```
gtaharaedmonds, line 42: nitpick: since C has no namespaces I think it's good practice to keep the name of the enum in each entry. Like `NO_ADC_CHANNEL` or `ADC_CHANNEL_UNAVAILABLE` for example.

gtaharaedmonds (changes requested):
Looks good overall, mostly nitpicks + a few CAN-related concerns. 

Let's get Amir to merge into master first, and then we'll merge this after.

can_bus/quadruna/VC/VC_alerts.json
liam-ilan: Originally ordered them to try to find a blank id, reverted them back!

can_bus/quadruna/VC/VC_alerts.json
liam-ilan: Makes sense, moved to warnings.

can_bus/quadruna/VC/VC_tx.json
liam-ilan: Fixed.

firmware/quadruna/VC/src/hw/hw_adc.h
liam-ilan: Fixed!

firmware/quadruna/VC/src/io/io_shutdown.h
liam-ilan: done!

firmware/boot/f4boot/cubemx/f4boot.ioc.md5
liam-ilan: md5s back!

liam-ilan:
A note on the md5 checksums, these are also broken on the base branch, to be fixed when base branch updated.

Lucien950 (changes requested):
I didn't look too deep, just things we have already talked about. It seems that these pins are currently unused. Please implement the corresponding behaviour. I'm putting the buck on you to implement the pin behaviour because DCM port is more concerned about torque generation.

LV Battery Things
- bat_i_sns, bat_i_sns_nflt (battery current sensing, battery fault)
- PGood (I believe indicates if the power from the LV batteries is boost converted to 22V)
- VBAT_SNS (Battery Voltage Sensing)
- 22V_boost_sense from battery

Power Distribution Things
- acc_i_sense, acc_i_sense_nflt (accumulator current sensing, for 24V from accumulator when driving)
- 24V_acc_sense
- nchrg and nchrg_nflt (LV Power Charging, external power)
- Current sensing lines for all load switches (LV power, shutdown power, aux power, pump power, inv l/r power, buzzer power, telem power)
- SB_ILCK_SHDN_SNS -> Splitter Box Interlock Shutdown Sensor
    - (should be easy to extend as there is currently lots of shutdown sensors)

Random Things
- led (random LED, idk not labelled)
- _900k_gpio (but it seems like it's for telemetry)
- Setup I2C connection to pumps (DPOT_SCL/SDA)
- INV_R/L_PROGRAM

firmware/thruna/DIM/src/cubemx/DIM.ioc.md5
```diff
@@ -1 +1 @@
-d41ad53b247e4e1d6391ac3ac07808ba
```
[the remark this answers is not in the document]
liam-ilan: Looking at my branch right now, I think I made some branch errors, will fix for tomorrow. I'll try to get a clean PR based on master rather than Amir's branch, sorry for the md5 mess.

Same thing applies to vehicle dynamics.

liam-ilan:
@Lucien950 thanks for the list! I'll run through it and make sure everything is implemented. I'm closing this PR, and opening a new one based on master, which should fix the md5/Vehicle dynamics issues.

PR #1192, opened 2024-03-23
Chimera Startup Button Implementation

will-chaba:
### Summary

Enable functionality of putting threads to sleep if chimera button held on reset

### Changelist 

- All tasks sleep after reset if chimera button pressed
- Added required pull-up to button on all boards

### Testing Done

Tested with basic Chimera script on BMS.

### Checklist


firmware/shared/src/io/io_chimera.c
```diff
@@ -24,6 +23,14 @@ static uint8_t  rx_packet_size;
 static uint32_t net_name_gpio;
 static uint32_t net_name_adc;
 
+static const Gpio *n_chimera_gpio;
+
+static bool chimera_button_pressed(void)
```
gtaharaedmonds, line 28: nitpick: `chimeraButtonPressed`

firmware/shared/src/io/io_chimera.c
```diff
@@ -88,15 +95,28 @@ static AdcChannel io_chimera_parseNetLabelAdc(const AdcNetName *net_name)
...
+        hw_uart_receiveIt(uart, data, DEBUG_SIZE_MSG_BUF_SIZE);
+    }
+}
+
+void io_chimera_sleepTaskIfEnabled(void)
+{
+    if (chimera_button_pressed())
+    {
+        osDelay(osWaitForever);
+    }
```
gtaharaedmonds, line 119: You should only check the button once, and then save the value as a static variable. I feel like there's a chance of confusing behavior happening if the button is pressed/released right as the board wakes up.

gtaharaedmonds: Like its theoretically possible you press the button as a task starts, but not when it checks for the UART RX call.

gtaharaedmonds (approved):
Minor change requested but LGTM. Can merge once addressed

PR #1205, opened 2024-03-28
Improve error handler

gtaharaedmonds:
### Summary

Added some macro magic (from the old leads) such that `Error_Handler` in `main.c` (what STM32CubeMX generates as the error handling function if anything goes wrong during peripheral initialization) calls our own custom `hw_error` function instead. This function gets the line number, file name, and function call and prints it over over RTT. It also sets a breakpoint.

### Testing Done

Tested on H7 dev board.

### Checklist


Lucien950 (approved):
GREAT SUCCESS

PR #1207, opened 2024-03-28
IMU Linear Acceleration

Jugveer-Sandher:
### Summary
VC can broadcast the linear acceleration from the IMU. I got some of the code for hw_i2c.h/c from `soroush/I2C_driver` branch, so I can wait for him to merge before I do.

### Testing Done
Tested on the VC and seems to be working fine.

### Checklist


firmware/quadruna/VC/src/io/io_imu.c
```diff
@@ -0,0 +1,76 @@
...
+    hw_i2c_mem_read(imu, 0x2C, 1, z_data, 1);
+    hw_i2c_mem_read(imu, 0x2D, 1, z_data, 1);
+    int16_t z_raw = (int16_t)(z_data[1] << 8 | z_data[0]);
+
+    // Convert raw value to acceleration in m/s^2
+    float z_acceleration = z_raw * SENSITIVITY * 9.81f / 1000.0f;
+
+    app_canTx_VC_Z_set(z_acceleration);
+    return z_acceleration;
+}
```
Lucien950, line 76: is there a way to take the shared parts of these three functions and bring it into one function?

Lucien950 (changes requested):
one last thing

Lucien950 (approved):

firmware/quadruna/VC/src/io/io_imu.c
```diff
@@ -0,0 +1,73 @@
...
+const float SENSITIVITY = 0.061f;
+
+bool io_imu_init()
+{
+    imu->i2c_handle     = &hi2c2;
+    imu->target_address = 0x6B;
+    imu->timeout_ms     = 100;
+
+    return hw_i2c_is_target_ready(imu);
+}
```
Lucien950, line 20:
```suggestion
static I2cInterface imu {
    &hi2c2, 0x6B, 100
};

// Default sensitivity for LSM6DSM is 0.061 mg/digit
const float SENSITIVITY = 0.061f;

bool io_imu_init()
{
    return hw_i2c_is_target_ready(imu);
}
```

firmware/quadruna/VC/src/tasks.c
```diff
@@ -268,6 +269,8 @@ bool (*heartbeatFaultGetters[HEARTBEAT_BOARD_COUNT])() = {
     [CRIT_HEARTBEAT_BOARD] = app_canAlerts_VC_Fault_MissingCRITHeartbeat_get
 };
 
+static I2cInterface *imu;
```
Lucien950, line 272:
```suggestion
```

firmware/quadruna/VC/src/tasks.c
```diff
@@ -310,6 +313,9 @@ void tasks_init(void)
         Error_Handler();
     }
 
+    io_imu_init();
```
Lucien950, line 316: do something with the return?

Lucien950 (changes requested):

firmware/quadruna/VC/src/io/io_imu.c
Lucien950: and make all the rest of the functions take the address of this imu.

firmware/quadruna/VC/src/io/io_imu.c
```diff
@@ -0,0 +1,73 @@
...
+
+bool io_imu_init_accelerometer()
+{
+    uint8_t buffer[1];
+    buffer[0] = 0x40; // turn on accelerometer to normal mode
+
+    return hw_i2c_mem_write(imu, 0x10, 1, buffer, sizeof(buffer));
```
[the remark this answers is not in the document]
gtaharaedmonds: Also maybe group this into `io_imu_init()`?

firmware/quadruna/VC/src/tasks.c
gtaharaedmonds: Logging a warning over CAN is probably the best bet

firmware/quadruna/VC/src/io/io_imu.h
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @brief gets the linear acceleration in the y direction
+ */
+double io_imu_acceleration_y();
+
+/**
+ * @brief gets the linear acceleration in the y direction
+ */
+double io_imu_acceleration_z();
```
gtaharaedmonds, line 28: nitpicks: `camelCase` for functions here.

so: `io_imu_accelerationX` for example.

firmware/quadruna/VC/src/io/io_imu.c
```diff
@@ -0,0 +1,73 @@
...
+
+    app_canTx_VC_X_set(x_acceleration);
+    return x_acceleration;
+}
+
+double io_imu_acceleration_y()
+{
+    uint8_t y_data[2];
+    read_register(y_data, 0x2A);
+    int16_t y_raw = (int16_t)((int16_t)y_data[1] << 8 | y_data[0]);
```
gtaharaedmonds, line 53: Instead of declaring a byte array, you can pass in a `uint16_t`. ARM microcontrollers (the ones we use!) are little endian, which means that least significant bytes have lower addresses, and more significant bytes have higher memory addresses. 

So, for example, if a `uint16_t` is stored at address 0x2000, then the LSB lives at 0x2000 and the MSB lives at 0x2001. Similarly, a `uint8_t` array of length 2 stored at 0x2000 will have element 0 at 0x2000 and element 1 at 0x2001. 

Notice that a regular `uint16_t`, and a `uint8_t[2]` with its LSB/MSB representing a 16-bit int, will have equivalent representations in memory (LSB at 0x2000, MSB at 0x2000 + 1).

firmware/quadruna/VC/src/io/io_imu.c
```diff
@@ -0,0 +1,73 @@
...
+    buffer[0] = 0x40; // turn on accelerometer to normal mode
+
+    return hw_i2c_mem_write(imu, 0x10, 1, buffer, sizeof(buffer));
+}
+
+void read_register(uint8_t *rx_buffer, uint16_t mem_addr)
+{
+    hw_i2c_mem_read(imu, mem_addr, 1, &rx_buffer[0], 1);         // LSB
+    hw_i2c_mem_read(imu, (mem_addr + 0x1), 1, &rx_buffer[1], 1); // MSB
+}
```
gtaharaedmonds, line 34: I believe you can just invoke `hw_i2c_mem_read` once with a size of 2 bytes, and it'll do the same as you're doing here.

gtaharaedmonds, line 34: Maybe think about error handling here - if any of these calls fail, the `rx_buffer` will be filled with garbage, and we'll transmit nonsensical random values over the CAN bus.

gtaharaedmonds: You should only return values if the I2C transactions succeed, and ideally we set a warning if I2C transactions are failing.

firmware/shared/src/hw/hw_i2c.h
```diff
@@ -0,0 +1,76 @@
...
+    uint16_t           target_address;
+    uint32_t           timeout_ms;
+} I2cInterface;
+
+/**
+ * @brief Check if device connected to the given I2C interface is ready for communication.
+ * @param i2c The given I2C interface.
+ * @return True if connected device is ready to communicate over I2C.
+ */
+bool hw_i2c_is_target_ready(const I2cInterface *i2c);
```
gtaharaedmonds, line 23: nitpick: functions should be `camelCase`. 

So `hw_i2c_isTargetReady` (here and below).

gtaharaedmonds (changes requested):
nice

firmware/shared/src/hw/hw_i2c.c
```diff
@@ -0,0 +1,63 @@
...
+    uint16_t            tx_buffer_size)
+{
+    if (hw_i2c_is_target_ready(i2c))
+    {
+        return HAL_I2C_Mem_Write(
+                   i2c->i2c_handle, (i2c->target_address << 1), mem_addr, mem_size, tx_buffer, tx_buffer_size,
+                   i2c->timeout_ms) == HAL_OK;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
gtaharaedmonds: The `I2C_MEMADD_SIZE_8BIT` macro is defined to 1, and looks like you're always reading memory of size 1, which is maybe why it was working on the bench despite this bug. :)

firmware/quadruna/VC/src/io/io_imu.c
Jugveer-Sandher: I removed the `read_register` function since it became redundant and I can just call `hw_i2c_mem_read` directly, but `hw_i2c_mem_read` requires a `uint8_t* rx_buffer` since it is easier to work with byte size data and I think most registers in `HAL_I2C` have one byte registers. So wouldn't it just be best to keep the `uint8_t y_data[2]` array instead?

firmware/quadruna/VC/src/io/io_imu.c
Jugveer-Sandher: I did add this functionality but I'm setting a CAN warning in io level. Is that okay?

firmware/quadruna/VC/src/app/states/app_allStates.c
```diff
@@ -39,6 +40,24 @@ bool app_allStates_runOnTick100Hz(void)
     app_heartbeatMonitor_checkIn();
     app_heartbeatMonitor_tick();
 
+    float *x_acceleration               = 0;
+    float *y_acceleration               = 0;
+    float *z_acceleration               = 0;
+    bool   successful_acceleration_read = io_imu_getLinearAccelerationX(x_acceleration) ||
+                                        io_imu_getLinearAccelerationY(y_acceleration) ||
+                                        io_imu_getLinearAccelerationZ(z_acceleration);
```
Lucien950, line 48:
```suggestion
    float x_acceleration;
    float y_acceleration;
    float z_acceleration;
    bool   successful_acceleration_read = io_imu_getLinearAccelerationX(&x_acceleration) ||
                                        io_imu_getLinearAccelerationY(&y_acceleration) ||
                                        io_imu_getLinearAccelerationZ(&z_acceleration);
```

Lucien950 (changes requested):

Lucien950 (approved):

gtaharaedmonds (approved):

PR #1210, opened 2024-03-30
Integrate file system logging to VC

Jiajun-Huang:
### Summary
Integrate file system logging to VC
Add custom assert to lfs library 

### Testing Done
Not test done yet. 

### Checklist


will-chaba (changes requested):
Looks good, couple small things

firmware/quadruna/VC/src/tasks.c
```diff
@@ -80,6 +90,29 @@ static const CanConfig can_config = {
...
+}
+
+static uint32_t overflow_count = 0;
+static uint32_t read_count     = 0;
+static uint32_t write_count    = 0;
+
+static void dummyOverflow(uint32_t f)
+{
+    overflow_count++;
+}
```
Lucien950, line 105: dummy functions?

firmware/quadruna/VC/src/tasks.c
```diff
@@ -475,6 +515,44 @@ _Noreturn void tasks_runCanRx(void)
...
+_Noreturn void tasks_runLogging(void)
+{
+    static uint32_t count = 0;
+    for (;;)
+    {
+        if (!sd_functional())
+        {
+            osDelay(1000);
+            continue;
+        }
```
Lucien950, line 527: dynamic sd card init? maybe move it from the init function to in here?

firmware/quadruna/VC/src/tasks.c
```diff
@@ -475,6 +515,44 @@ _Noreturn void tasks_runCanRx(void)
     }
 }
 
+_Noreturn void tasks_runLogging(void)
+{
+    static uint32_t count = 0;
```
Lucien950, line 520: what count? more specific variable naming please

firmware/quadruna/VC/src/tasks.c
```diff
@@ -475,6 +515,44 @@ _Noreturn void tasks_runCanRx(void)
...
+    {
+        io_canLogging_msgReceivedCallback(rx_msg); // push to logging queue
+    }
+    read_count++;
+}
+
+static bool sd_functional(void)
+{
+    return !hw_gpio_readPin(&sd_present);
+}
```
Lucien950, line 554: please just inline this function.

firmware/quadruna/VC/src/tasks.c
```diff
@@ -80,6 +90,29 @@ static const CanConfig can_config = {
     .rx_overflow_clear_callback = canRxQueueOverflowClearCallback,
 };
 
+static bool dummyFilter(uint32_t msg_id)
```
Lucien950, line 93: it seems most things in this block can be moved to another app file

firmware/quadruna/VC/src/tasks.c
Lucien950: or move it to io/app function. Keep tasks.c as minimal as possible.

Lucien950 (approved):
There is a lot of seperation of concerns things happening here. I know there are refactors that we need to make, so I would like to discuss the decisions you have made here before merging them in?

Lucien950 (changes requested):
oops clicked the wrong button

firmware/quadruna/VC/src/tasks.c
```diff
@@ -475,6 +515,44 @@ _Noreturn void tasks_runCanRx(void)
...
+        write_count++;
+        if (count > 256)
+        {
+            io_canLogging_sync();
+            count = 0;
+        }
+    }
+}
+
+static void can_msg_received_callback(CanMsg *rx_msg)
```
Lucien950, line 540: this is an io_can construct??

Lucien950:
also if there is any functions that are exported from a header file, then you need to prefix it (for example with `io_canlogging`)

firmware/shared/src/hw/hw_sd.c
```diff
@@ -7,17 +7,17 @@
...
 {
     sd = sd_config;
-    if (sd_inited)
-    {
-        sd->sd_init_complete = true;
-    }
+    // if (sd_inited)
+    // {
+    //     sd->sd_init_complete = true;
+    // }
```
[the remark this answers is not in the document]
Jiajun-Huang: I missed this part when I reviewed the BMS code. I think it is a bad assumption that we have an extern variable in shared driver code. I think I check if the sd card inited in task.c .

firmware/quadruna/VC/src/tasks.c
Jiajun-Huang: I can change the function name, but I think the behavior will not change, since I don't what should we filter and what to do when overflow.

firmware/quadruna/VC/src/tasks.c
Jiajun-Huang: This callback function gets called when rx can interrupt fire. Originally, we only called `io_can_msgReceivedCallback` which push the message to a queue, but now we need to push the message to another queue for logging. So I need to replace the behavior of original callback function to this new callback function.

firmware/quadruna/VC/src/tasks.c
Jiajun-Huang: I don't want the io function coupled with this gpio pin. This is the reason I put it here.

firmware/quadruna/VC/src/tasks.c
Jiajun-Huang: This means that if the sd card fall off from the sd card slot, we should deschedule the task, and try again later.

firmware/quadruna/VC/src/tasks.c
```diff
@@ -475,6 +515,44 @@ _Noreturn void tasks_runCanRx(void)
...
+{
+    // TODO: check gpio present
```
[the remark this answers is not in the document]
Jiajun-Huang: already checked

Lucien950 (approved):
Screenshot 2023-11-02 014238

firmware/quadruna/VC/src/io/io_lfsConfig.c
```diff
@@ -0,0 +1,109 @@
...
+    // cfg->block_count    = block_number / IO_LFS_BLOCK_SIZE_FACTOR;
+    cfg->block_count      = 1000000;
+    cfg->block_cycles     = IO_LFS_BLOCK_CYCLES;
+    cfg->cache_size       = IO_LFS_CACHE_SIZE;
+    cfg->read_buffer      = lfs_read_buffer;
+    cfg->prog_buffer      = lfs_prog_buffer;
+    cfg->lookahead_buffer = lfs_lookahead_buffer;
+    // cfg->attr_max         = 0;
+
+    return 0;
```
forbesashli, line 108: nit but if we are gonna do ints for returns, should be defined like LFS_ERR_OK etc are

firmware/shared/src/io/io_canLogging.c
```diff
@@ -18,9 +21,7 @@ static uint8_t            queue_buf[QUEUE_BYTES];
 static uint32_t current_bootcount;
 static int      log_fd; // fd for the log file
 
-extern SdCard sd;
-extern Gpio   sd_present;
-extern bool   sd_inited;
+static uint8_t logging_error_remaining = 5; // number of times to error before stopping logging
```
forbesashli, line 24: again nit with the defining of numbers

firmware/shared/src/io/io_canLogging.c
```diff
@@ -33,63 +34,92 @@ static const osMessageQueueAttr_t queue_attr = {
...
+    CanMsgLog log;
+    convertCanMsgToLog(msg, &log);
+
+    if (osMessageQueuePut(message_queue_id, &log, 0, 0) != osOK && config->tx_overflow_callback != NULL)
     {
         // If pushing to the queue failed, the queue is full. Discard the msg and invoke the TX overflow callback.
         config->tx_overflow_callback(++tx_overflow_count);
+        return 1;
     }
+    return 0;
```
forbesashli, line 96: define returns

firmware/shared/src/io/io_canLogging.c
```diff
@@ -33,63 +34,92 @@ static const osMessageQueueAttr_t queue_attr = {
...
+    CanMsgLog tx_msg;
     osMessageQueueGet(message_queue_id, &tx_msg, NULL, osWaitForever);
 
-    io_fileSystem_write(log_fd, &tx_msg, sizeof(tx_msg));
-    // assert(size = sizeof(tx_msg));
+    int err = io_fileSystem_write(log_fd, &tx_msg, sizeof(tx_msg));
+    if (err < 0)
+    {
+        logging_error_remaining--;
+        return 1;
```
forbesashli, line 112: define returns

firmware/shared/src/io/io_canLogging.c
```diff
@@ -99,16 +129,23 @@ void io_canLogging_msgReceivedCallback(CanMsg *rx_msg)
...
-    io_fileSystem_sync(log_fd);
-}
\ No newline at end of file
+    int err = io_fileSystem_sync(log_fd);
+    if (err < 0)
+    {
+        logging_error_remaining--;
+        return 1;
+    }
+    return 0;
```
forbesashli, line 150: define returns

forbesashli (commented):
looks good, a personal preference at the return values being defined (makes it easier to read)

PR #1213, opened 2024-03-30
VC GPS Fix

philam2001:
### Summary

- Add new GPS functionality, wanting to use EKF in mode ` SBG_ECOM_SOL_MODE_NAV_POSITION ` to measure vehicle speed
- Fix GPS callback on Quadruna
- Tested and generally works but triggers watchdog sometimes, haven't had time to test it before 2024 Comp so will comment out

### TODO (For the future)
- test ekf while in motion, values at the moment are suspicious
- figure out watchdog issue
### Checklist


Lucien950:
@1mackenziekyle after you left we discovered that the pins that you guys made was backwards, cause you need the RX to connect to the TX (because UART pin naming is relative to the broadcasting element). From there we didn't get much further, but we have electrically validated that the GPS is sending the messages to the MCU.

can_bus/quadruna/VC/VC_alerts.json
```diff
@@ -105,6 +105,10 @@
         "RightInverterFault": {
             "id": 556,
             "description": "Right inverter entered fault state."
-        }
+        },
+        "SBGModeFault":  {
+            "id" :561,
+            "description" : "Ensure mode is Nominal mode for EKF data used"
+        }
```
gtaharaedmonds, line 112: make disabled

gtaharaedmonds (approved):
Approved once you make that fault disabled too.

can_bus/quadruna/VC/VC_alerts.json
philam2001: yup

PR #1214, opened 2024-03-30
LV State of Charge

liam-ilan:
### Summary
Implement state of charge for our 18650 cells, and throw it over CAN.

### Changelist 
- Added app soc layer, and new can message to broadcast soc info.

### Testing Done
- None.

### Checklist


firmware/quadruna/VC/src/app/app_soc.c
```diff
@@ -0,0 +1,84 @@
...
+#include "app_canTx.h"
+#include "io_lowVoltageBattery.h"
+
+static SocStats stats;
+
+void app_soc_init(void)
+{
+    stats.prev_current_A = 0.0f;
+
+    stats.charge_c = app_soc_getChargeFromOcv();
```
will-chaba, line 13: Ideally, this value would be stored in persistent memory (SD), as the voltage of the cells only mimic "Open-circuit" conditions after long periods of settling with no current draw (~1 hour).

Resets based on OCV are still required, but should be done so manually if possible.

firmware/quadruna/VC/src/app/app_soc.h
```diff
@@ -0,0 +1,43 @@
...
+static const float ocv_charge_spline_anchors[OCV_CHARGE_SPLINE_ANCHORS_SIZE][2] = {
+    { 2.49f, 158.4f }, { 2.9994f, 679.7f }, { 3.178f, 1180.8f }, { 3.275f, 1800.0f },  { 3.403f, 3333.6f },
+    { 3.5f, 5162.4f }, { 3.66f, 10152.0f }, { 4.09f, 17870.4f }, { 4.1674f, 18702.0f }
+};
+
+void  app_soc_init(void);
+void  app_soc_update(void);
+float app_soc_getPercent(void);
+void  app_soc_broadcast(void);
+float app_soc_getChargeFromOcv(void);
```
will-chaba, line 43: I know this is in progress, but add function descriptions

firmware/quadruna/VC/src/app/states/app_allStates.c
```diff
@@ -33,6 +34,9 @@ bool app_allStates_runOnTick100Hz(void)
     app_heartbeatMonitor_tick();
     app_heartbeatMonitor_broadcastFaults();
 
+    app_soc_update();
+    app_soc_broadcast();
```
will-chaba, line 38: broadcast could likely be moved to 1Hz to save on bus load

firmware/quadruna/VC/src/app/app_soc.c
```diff
@@ -0,0 +1,84 @@
...
+        right_anchor_index = OCV_CHARGE_SPLINE_ANCHORS_SIZE - 1;
+    }
+
+    float left_ocv_anchor     = ocv_charge_spline_anchors[left_anchor_index][0];
+    float left_charge_anchor  = ocv_charge_spline_anchors[left_anchor_index][1];
+    float right_ocv_anchor    = ocv_charge_spline_anchors[right_anchor_index][0];
+    float right_charge_anchor = ocv_charge_spline_anchors[right_anchor_index][1];
+
+    // compute slope in sector
+    float m = (right_charge_anchor - left_charge_anchor) / (right_ocv_anchor - left_ocv_anchor);
```
will-chaba, line 79: nit: use more descriptive variable name

can_bus/quadruna/VC/VC_tx.json
```diff
@@ -717,5 +717,18 @@
                 "unit": "L/H"
             }
         }
+    },
+    "StateOfCharge": {
+        "msg_id": 226,
+        "cycle_time": 100,
```
will-chaba, line 723: See the other comment about moving broadcast to 1Hz, maybe up cycle time to 1000 if that change is performed

firmware/quadruna/VC/src/app/app_soc.c
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void app_soc_broadcast()
+{
+    app_canTx_VC_Soc_set(app_soc_getPercent());
+}
+
+float app_soc_getChargeFromOcv(void)
+{
+    float battery_ocv = io_lowVoltageBattery_getBatVoltage();
```
will-chaba, line 44: Does this return the individual cell voltage? or the voltage of the two cells in series? I think it may be the latter unless a change was made from last year

will-chaba (changes requested):
Looks good so far! Left some nitpicks.

From an accuracy standpoint, I have a few major concerns:
1. No individual cell voltage sense. To my knowledge, we only measure the voltage of the two cells in parallel, SOC is based on the capacity of the lowest charge single cell, and therefore because we can't isolate between the two cells, our starting point will always be inaccurate. 
2. Heavy voltage sag due to resetting LV while power actively being drawn from cells. In the accumulator, we have the benefit of not drawing any current from the cells in init state, the VC does not have this luxury. Voltage will realistically always be sagged when the ocv->charge lookup is performed, resulting in an inaccurate value.
3. No persistent saving of charge value to non-volatile memory. Without doing this, we will always be resetting based on voltage, which is incredibly inaccurate. If we are not going to save to persistent memory, we may as well just do a ocv->soc lookup table and omit coulomb counting altogether. I'd like to see the system store it's last known value to the SD card and recover based on that at startup if possible. Hell. If it'd interfere with datalogging, might be better off having the BMS store it to its SD card and broadcast the value back to the VC to reset upon at startup. There are definitely options, but storing the value in memory should absolutely be done.

firmware/quadruna/VC/src/app/app_soc.h
liam-ilan: done!

firmware/quadruna/VC/src/app/states/app_allStates.c
liam-ilan: agreed, fixed

can_bus/quadruna/VC/VC_tx.json
liam-ilan: fixed!

firmware/quadruna/VC/src/app/app_soc.c
liam-ilan: yep! also fixed a bug there, good call

firmware/quadruna/VC/src/app/app_soc.c
liam-ilan: Returns two cells in series, will be inaccurate, the ocv charge spline anchors reflect this doubling, though it will be inaccurate.

firmware/quadruna/VC/src/app/app_soc.c
liam-ilan: Will implement, currently looks like the SD card io interface isn't complete yet, will add todo note

PR #1217, opened 2024-03-31
VC State Machine Transitions

Jugveer-Sandher:
### Summary
Made some minor changes to the state machine transitions on the VC. Since we don't want the inverters to turn off if the start switch is off or the BMS is not in `BMS_Drive_State` but still may be in `BMS_Inverter_On_State` or `BMS_Precharge_State`.

### Testing Done
Created some new unit tests

### Checklist

### DIFF
--- a/firmware/quadruna/VC/src/app/states/app_inverterOnState.c
+++ b/firmware/quadruna/VC/src/app/states/app_inverterOnState.c
@@ -29,6 +29,8 @@ static void inverterOnStateRunOnTick100Hz(void)
     prev_start_switch_pos                 = curr_start_switch_pos;
     const bool is_brake_actuated          = app_canRx_FSM_BrakeActuated_get();
     const bool bms_in_drive_state         = app_canRx_BMS_State_get() == BMS_DRIVE_STATE;
+    const bool bms_in_inverterOn_state    = app_canRx_BMS_State_get() == BMS_INVERTER_ON_STATE;
+    const bool bms_in_precharge_state     = app_canRx_BMS_State_get() == BMS_PRECHARGE_STATE;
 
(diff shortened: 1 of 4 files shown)


firmware/quadruna/VC/src/app/states/app_inverterOnState.c
```diff
@@ -26,6 +26,8 @@ static void inverterOnStateRunOnTick100Hz(void)
     prev_start_switch_pos                 = curr_start_switch_pos;
     const bool is_brake_actuated          = app_canRx_FSM_BrakeActuated_get();
     const bool bms_in_drive_state         = app_canRx_BMS_State_get() == BMS_DRIVE_STATE;
+    const bool exit = !all_states_ok || (!bms_in_drive_state && app_canRx_BMS_State_get() != BMS_INVERTER_ON_STATE &&
```
Lucien950, line 29: Make a bit more clear what the BMS condition is with a variable name

firmware/quadruna/VC/test/test_stateMachine.cpp
```diff
@@ -223,3 +230,50 @@ TEST_F(VCStateMachineTest, drive_to_init_state_on_CRIT_fault)
 
     TestFaultBlocksDrive(set_fault, clear_fault);
 }
+
+TEST_F(VCStateMachineTest, drive_to_init_inverter_fault)
+{
+    app_canRx_CRIT_StartSwitch_update(SWITCH_ON);
```
Lucien950, line 236: Can we make going to drive state a single reusable function, and make a test to make sure that function is working

Lucien950 (changes requested):

Lucien950 (approved):

PR #1218, opened 2024-04-03
Re+Moving faultchecks to new file

AmirTajaddodi:
### Summary

I moved all the fault checking done in the all states to its own individual file and created a boardFaultCheck function as well as 
an inverterFaultCheck function for VC

### Testing Done

Not much testing required ran the old test already there and it passed all of them

### Checklist


Lucien950 (approved):
spectacular ✨. Please move app_faultcheck with the rest of the app files (outside of the state folder) and I will be the most happy.

Lucien950 (changes requested):
wait i'm wrong isn't the whole point that you move fault checking outside of the allStates function? I think the purpose of the removal was because we should realize that allStates functions should just go in `tasks.c` @gtaharaedmonds

firmware/quadruna/VC/src/app/states/app_allStates.c
```diff
@@ -2,6 +2,7 @@
 #include "states/app_allStates.h"
 
 #include "app_sbgEllipse.h"
+#include "app_faultCheck.h"
```
Lucien950, line 5:
```suggestion
```

firmware/quadruna/VC/src/app/states/app_driveState.c
```diff
@@ -85,7 +86,9 @@ static void driveStateRunOnTick1Hz(void)
 static void driveStateRunOnTick100Hz(void)
 {
     // All states module checks for faults, and returns whether or not a fault was detected.
-    const bool all_states_ok = app_allStates_runOnTick100Hz();
+    const bool any_board_has_fault = app_boardFaultCheck();
+    const bool inverter_has_fault  = app_inverterFaultCheck();
+    const bool all_states_ok       = !(any_board_has_fault || inverter_has_fault);
```
Lucien950, line 91: this sort of pattern is used a lot, perhaps think about making it also a function? not entirely mandatory.

Lucien950 (approved):
i'm happy. Little change that you could do that I recommended, but not strictly required.

PR #1222, opened 2024-04-06
Seperate Voltage Readings from Cell Discharging for Balancing

will-chaba:
### Changelist 

Create a new state in iso_spi state machine to separate cell monitoring from balancing
- When balancing enabled:
    1. Take Cell measurements for 100 Cycles (1s)
    1. Run Open Wire Check
    1. Run Cell Balancing for 1000 Cycles (10s)

- When balancing disabled
    1. Take Cell measurements for 100 Cycles (1s)
    1. Run Open Wire Check

### Testing Done

- Tested on Thruna pack
- Tested on Quadruna segment


firmware/quadruna/BMS/src/app/states/app_allStates.c
```diff
@@ -54,25 +58,62 @@ bool app_allStates_runOnTick100Hz(void)
     app_heartbeatMonitor_tick();
     app_heartbeatMonitor_broadcastFaults();
 
+    const bool balancing_enabled = true; // app_canRx_Debug_CellBalancingRequest_get();
```
DJ90864, line 61: I'm assuming you did it for testing and probably forgot to revert to CAN Rx flag.

DJ90864 (changes requested):
just one change needed before merging

firmware/quadruna/BMS/src/app/states/app_allStates.c
will-chaba: Ah yes good catch thank you

DJ90864 (approved):

PR #1229, opened 2024-04-09
Crit switchleds

leahmariefernandez:
### Changelist 
- Switch LEDs are being read from VC
- Removed hardware function calls from the application layer and created a ShutdownSensor module
- Added tests for drive mode, shutdown sensor, LED status and CAN messages

### Testing Done
Unit tests and integration test. Did not test on board.


firmware/quadruna/CRIT/test/test_shutdownSensor.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    // Set the function to return a specific value when called with a specific sensor.
+    fake_io_shutdownSensor_readPin_returnsForArgs(&sensor, true);
+
+    // Call the function with the sensor.
+    io_shutdownSensor_readPin(&sensor);
+
+    // Verify that the function was called with the specific arguments.
+    ASSERT_EQ(1, fake_io_shutdownSensor_readPin_callCountForArgs(&sensor));
+}
```
Lucien950, line 48: usually you want to assert on the app level effects of the code. This code is just testing if the fake io generation would work. For example, when a shutdown io pin is enabled, what happens? (usually the answer is something on CAN, or another peripheral, especially so for the CRIT).

firmware/quadruna/CRIT/src/app/app_mainState.c
```diff
@@ -22,20 +23,24 @@ static void mainStateRunOnTick100Hz(void)
...
 
     const bool regen_switch_on = io_switch_isClosed(globals->config->regen_switch);
     app_canTx_CRIT_RegenSwitch_set(regen_switch_on ? SWITCH_ON : SWITCH_OFF);
-    io_led_enable(globals->config->regen_led, regen_switch_on);
 
     const bool torquevec_switch_on = io_switch_isClosed(globals->config->torquevec_switch);
     app_canTx_CRIT_TorqueVecSwitch_set(torquevec_switch_on ? SWITCH_ON : SWITCH_OFF);
-    io_led_enable(globals->config->torquevec_led, torquevec_switch_on);
+
+    if (app_canRx_VC_State_get() == 2)
```
Lucien950, line 38: use the enum to compare

firmware/quadruna/CRIT/src/app/app_mainState.c
```diff
@@ -22,20 +23,24 @@ static void mainStateRunOnTick100Hz(void)
...
+    {
+        io_led_enable(globals->config->start_led, true);
+        io_led_enable(globals->config->regen_led, true);
+        io_led_enable(globals->config->torquevec_led, true);
```
Lucien950, line 42: idk about the logic here, i don't think this is correct. state drives the start led but not the regen/torque vectoring LEDs.

firmware/quadruna/CRIT/test/test_driveMode.cpp
```diff
@@ -0,0 +1,49 @@
...
+    // Verifying that the drive mode value is consistent across multiple reads
+    uint16_t consistentValue = 0b1010;
+    fake_io_driveMode_readPins_returns(consistentValue);
+
+    ASSERT_EQ(consistentValue, io_driveMode_readPins());
+    ASSERT_EQ(consistentValue, io_driveMode_readPins());
+
+    // The drive mode should be read consistently, and the fake should be called each time
+    ASSERT_EQ(2u, fake_io_driveMode_readPins_callCount());
+}
```
Lucien950, line 49: same thing here. Consider what the **last** effect of triggering the IO pin is.

firmware/quadruna/CRIT/test/test_ledStatus.cpp
```diff
@@ -0,0 +1,159 @@
...
+    ASSERT_EQ(1, fake_io_rgbLed_enable_callCountForArgs(&crit_status_led, false, false, true));
+}
+
+TEST_F(LedStatusTest, crit_board_status_led_control_with_no_error)
+{
+    // Don't set any error and check that the CRIT LED turns green
+    LetTimePass(10);
+    ASSERT_EQ(1, fake_io_rgbLed_enable_callCountForArgs(&crit_status_led, false, true, false));
+}
+
```
Lucien950, line 75: CRIT test for LED turning red?

firmware/quadruna/CRIT/test/test_ledStatus.cpp
```diff
@@ -0,0 +1,159 @@
...
+    app_canRx_BMS_ImdOk_update(true);
+    app_canRx_BMS_BspdOk_update(true);
+    app_canRx_BMS_BmsOk_update(true);
+
+    // Set any critical error and check that the BMS LED turns red
+    app_canRx_BMS_Fault_CellOvertemp_update(true);
+    LetTimePass(10);
+
+    ASSERT_EQ(1, fake_io_rgbLed_enable_callCountForArgs(&bms_status_led, true, false, false));
+}
```
Lucien950, line 159: test all the LEDs too, very good for regression and very important for safety.
SHDN especially.

possibly a very good idea to write a generic test, and pass in function pointers for leds and canerrors as arguments.

very strange that only VC/FSM/2 of CRIT is tested only

make sure that test ordering is also organized and logical.

Lucien950 (changes requested):
solid work overall, just a few test things to fix. this is almost ready to drive.

firmware/quadruna/CRIT/src/app/app_mainState.c
leahmariefernandez: ah ok i see. ok i moved the regen/torque vec LEDs to being read from switches.

firmware/quadruna/CRIT/src/io/io_driveMode.h
```diff
@@ -2,6 +2,7 @@
 
 #include <stdbool.h>
 #include "app_utils.h"
+#include <stdint.h>
```
Lucien950, line 5:
```suggestion
```

Lucien950 (approved):

PR #1231, opened 2024-04-09
Quadruna Efuse Retry

Jugveer-Sandher:
### Changelist 
Efuse retry protocol:
- Implemented retry handler for efuse faults.
- The retry handler works as follows: If an efuse faults, we wait for 1 second (100 100Hz cycles) to obtain a current sum. We then check if the efuse is still faulting. If so, we disable this efuse and any dependent efuses (ex. all efuses depend on if LV is working). After a 1 second debounce time, we re-enable the efuse and wait another 1 second while getting the current sum. If the efuse is now working, we re-enable all dependent efuses. If the efuse is still faulting, we throw a fault.

- Still pending: Getting current values for the efuses.

### Testing Done
- Implemented unit testing

### DIFF
--- a/can_bus/quadruna/VC/VC_enum.json
+++ b/can_bus/quadruna/VC/VC_enum.json
@@ -10,5 +10,16 @@
     "INSUFFICIENT_READINGS": 1, 
     "INTERNAL_ERROR": 2, 
     "LIMIT_EXCEEDED": 3
+  },
+
+  "Efuse": {
+    "SHDN": 0,
+    "LV": 1,
+    "PUMP": 2,
+    "AUX": 3,
+    "INV_R": 4,
+    "INV_L": 5,
+    "TELEM": 6,
+    "BUZZER": 7
   }
 }
--- a/can_bus/quadruna/VC/VC_enum.json
+++ b/can_bus/quadruna/VC/VC_enum.json
@@ -13,13 +13,14 @@
   },
 
   "Efuse_Enum": {
-    "SHDN": 0,
-    "LV": 1,
-    "PUMP": 2,
-    "AUX": 3,
-    "INV_R": 4,
-    "INV_L": 5,
-    "TELEM": 6,
-    "BUZZER": 7
+    "NO_FAULT": 0,
+    "SHDN": 1,
+    "LV": 2,
+    "PUMP": 3,
+    "AUX": 4,
+    "INV_R": 5,
+    "INV_L": 6,
+    "TELEM": 7,
+    "BUZZER": 8
   }
 }
--- a/can_bus/quadruna/VC/VC_enum.json
+++ b/can_bus/quadruna/VC/VC_enum.json
@@ -12,7 +12,7 @@
     "LIMIT_EXCEEDED": 3
   },
 
-  "Efuse": {
+  "Efuse_Enum": {
     "SHDN": 0,
     "LV": 1,
     "PUMP": 2,
--- a/can_bus/quadruna/VC/VC_tx.json
+++ b/can_bus/quadruna/VC/VC_tx.json
@@ -771,5 +771,19 @@
                 "unit": "%"
             }
         }
+    },
+    "Efuses": {
+        "msg_id": 231,
+        "cycle_time": 100,
+        "description": "PDM efuses",
+        "signals": {
+            "Efuse": {
+                "enum": "Efuse"
+            },
+            "Fault_Count": {
+                "bits": 3,
+                "signed": false
+            }
+        }
     }
 }
--- a/can_bus/quadruna/VC/VC_tx.json
+++ b/can_bus/quadruna/VC/VC_tx.json
@@ -772,13 +772,13 @@
             }
         }
     },
-    "Efuses": {
+    "Efuses": { 
         "msg_id": 231,
         "cycle_time": 100,
         "description": "VC efuse retry protocol faults",
         "signals": {
             "Efuse": {
-                "enum": "Efuse"
+                "enum": "Efuse_Enum"
             },
             "Fault_Count": {
                 "bits": 3,
--- a/can_bus/quadruna/VC/VC_tx.json
+++ b/can_bus/quadruna/VC/VC_tx.json
@@ -772,15 +772,15 @@
             }
         }
     },
-    "Efuses": { 
+    "EfuseFaults": {
         "msg_id": 231,
         "cycle_time": 100,
         "description": "VC efuse retry protocol faults",
         "signals": {
-            "Efuse": {
+            "FaultedEfuse": {
                 "enum": "Efuse_Enum"
             },
-            "Fault_Count": {
+            "EFuseFaultCount": {
                 "bits": 3,
                 "signed": false
             }
--- a/can_bus/quadruna/VC/VC_tx.json
+++ b/can_bus/quadruna/VC/VC_tx.json
@@ -775,7 +775,7 @@
     "Efuses": {
         "msg_id": 231,
         "cycle_time": 100,
-        "description": "PDM efuses",
+        "description": "VC efuse retry protocol faults",
         "signals": {
             "Efuse": {
                 "enum": "Efuse"
--- a/firmware/quadruna/VC/src/app/app_powerManager.c
+++ b/firmware/quadruna/VC/src/app/app_powerManager.c
@@ -94,9 +94,8 @@ void app_powerManager_init()
     {
         retry_data[efuse].protocol_state          = PROTOCOL_STATE_OFF;
         retry_data[efuse].retry_attempts          = 0;
-        retry_data[efuse].current_timer_attempts  = 0;
+        retry_data[efuse].timer_attempts  = 0;
         retry_data[efuse].current_sum             = 0.0;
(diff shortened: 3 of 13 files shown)


firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+                    {
+                        go_to_init_state =
+                            app_retryHandler_start(efuse_retry_config->retry_protocol, efuse_retry_config, retry_data);
+                        efuse_retry_data->retry_attempts++;
+                    }
+                }
+                // Successful retry protocol after previous retry attempts
+                else if (efuse_retry_data->retry_attempts > 0)
+                {
+                    app_retryHandler_success(efuse_retry_config->retry_protocol, efuse_retry_config, retry_data);
```
gtaharaedmonds, line 175: I think we can reset the attempts counter if we successfully recovered.

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+}
+
 void app_powerManager_setState(PowerManagerState state)
 {
     current_power_state = state;
 
     for (int efuse = 0; efuse < NUM_EFUSE_CHANNELS; efuse++)
     {
-        io_efuse_setChannel(efuse, power_states_config[state].efuses[efuse]);
+        bool not_in_retry_protocol = retry_data[efuse].protocol_state != PROTOCOL_STATE_DEPENDENCY_WAITING;
```
gtaharaedmonds, line 101: We probably also don't want to change it if it's being debounced right?

can_bus/quadruna/VC/VC_enum.json
```diff
@@ -10,5 +10,16 @@
     "INSUFFICIENT_READINGS": 1, 
     "INTERNAL_ERROR": 2, 
     "LIMIT_EXCEEDED": 3
+  },
+
+  "Efuse_Enum": {
```
gtaharaedmonds, line 15: `EfuseEnum`

gtaharaedmonds: Also, maybe make another value called `NO_FAULT` or something like that. JSONCAN will transmit the zeroth value by default for everything, so at startup it's going to broadcast `SHDN` for `Efuse` and 0 for `Fault_Count` which is maybe confusing. Maybe manually set the former signal to `NO_FAULT` at startup to avoid this confusion.

can_bus/quadruna/VC/VC_tx.json
```diff
@@ -717,5 +717,19 @@
...
+        "signals": {
+            "Efuse": {
+                "enum": "Efuse_Enum"
+            },
+            "Fault_Count": {
+                "bits": 3,
+                "signed": false
+            }
+        }
     }
```
gtaharaedmonds, line 734: nitpicks:
`Efuse_Faults` -> `EfuseFaults` (Pascal case)
`Efuse` -> `FaultedEfuse` (EFuse is not a very descriptive name)
`Fault_Count` -> `EFuseFaultCount` (Fault count doesn't necessarily refer to the efuses)

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
-            [EFUSE_CHANNEL_LV] = true,
-            [EFUSE_CHANNEL_PUMP] = true,
-            [EFUSE_CHANNEL_AUX] = true,
-            [EFUSE_CHANNEL_INV_R] = true,
-            [EFUSE_CHANNEL_INV_L] = true,
-            [EFUSE_CHANNEL_TELEM] = true,
-            [EFUSE_CHANNEL_BUZZER] = true,
-        }
-    }
+    [POWER_MANAGER_SHUTDOWN] = { .retry_configs = { [EFUSE_CHANNEL_SHDN]   = { .efuse_state    = true,
```
gtaharaedmonds, line 16: Can you set the `min_needed_current` all to be zero for now? I think this'll be more convenient until the VC is in the car and we can test this properly. Otherwise it's gonna be constantly faulting while on the bench.

gtaharaedmonds (changes requested):
Looks great! Minor comments, then down to merge. I think we can/should test this on hardware first too.

firmware/quadruna/VC/src/app/app_powerManager.c
Jugveer-Sandher: I have it at 0.5 for unit testing, so I'll change it to 0 and comment out the tests.

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+            case PROTOCOL_STATE_CALC_AVG:
+            {
+                efuse_retry_data->current_sum += io_efuse_getChannelCurrent(efuse);
+                efuse_retry_data->timer_attempts++;
+                // Check if timer attempts reached the limit
+                if (efuse_retry_data->timer_attempts == TIMER_ATTEMPTS_LIMIT)
+                {
+                    efuse_retry_data->protocol_state = PROTOCOL_STATE_CALC_DONE;
+                }
+                break;
```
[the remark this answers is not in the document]
Jugveer-Sandher: For the switch statement

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+            }
+            // Handle finishing calculating current average
+            case PROTOCOL_STATE_CALC_DONE:
+            {
+                float avg = efuse_retry_data->current_sum / TIMER_ATTEMPTS_LIMIT;
+
+                // Calculate average and check if it's less than the minimum needed current to go into protocol
+                if (avg <= efuse_retry_config->min_needed_current)
```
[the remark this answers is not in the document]
Jugveer-Sandher: Still need to test on hardware and get a good min_needed_current reading that is a ~80% of the actual current.

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+                    io_efuse_setChannel(efuse, true);
+
+                    efuse_retry_data->protocol_state = PROTOCOL_STATE_CALC_AVG;
+                    efuse_retry_data->timer_attempts = 0;
+                }
+                else
+                {
+                    efuse_retry_data->timer_attempts++;
+                }
+            // Handle efuse dependant that is waiting for an efuse in protocol
```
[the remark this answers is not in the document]
Jugveer-Sandher: For example, if LV starts faulting and goes into the retry protocol. We also need to turn off the inverters to make sure that HV isn't on while LV may be faulting. So I call the inverters, in this case, a dependent efuse of LV.

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
...
+        if (not_in_retry_protocol)
+        {
+            io_efuse_setChannel(efuse, is_efuse_on);
+        }
+    }
+}
+
+bool app_powerManager_checkEfuses(PowerManagerState state)
+{
+    bool go_to_init_state = false;
```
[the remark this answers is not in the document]
Jugveer-Sandher: This function is checking if an efuse has faulted and then running the corresponding retry protocol and not all protocols make the car reset (go to init state).

firmware/quadruna/VC/src/app/app_retryHandler.c
```diff
@@ -0,0 +1,96 @@
...
+            retry_data[EFUSE_CHANNEL_INV_R].protocol_state = PROTOCOL_STATE_DEBOUNCE;
+            retry_data[EFUSE_CHANNEL_INV_L].protocol_state = PROTOCOL_STATE_DEPENDENCY_WAITING;
+            retry_data[EFUSE_CHANNEL_SHDN].protocol_state  = PROTOCOL_STATE_DEPENDENCY_WAITING;
+            break;
+        case RETRY_PROTOCOL_INV_L:
+            io_efuse_setChannel(EFUSE_CHANNEL_INV_R, false);
+            io_efuse_setChannel(EFUSE_CHANNEL_INV_L, false);
+            io_efuse_setChannel(EFUSE_CHANNEL_SHDN, false);
+
+            go_to_init = true;
```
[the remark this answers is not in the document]
Jugveer-Sandher: Depending on the retry protocol, we may need to reset the car (go to init state).

firmware/quadruna/VC/src/app/app_powerManager.c
```diff
@@ -1,52 +1,211 @@
 #include "io_efuse.h"
 #include "app_powerManager.h"
-#include "app_canRx.h"
+#include "app_canTx.h"
+
+const int   DEBOUNCE_TIMER_LIMIT = 100;
```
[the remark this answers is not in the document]
Jugveer-Sandher: After turning off an efuse we give it a debounce time to settle it down before turning back on again.

PR #1236, opened 2024-04-13
Fix Heartbeat Monitor for CRIT

liam-ilan:
### Changelist 
- Called heartbeat check in/broadcast faults functions in CRIT.

### Testing Done
- [ ] Flashed on boards w/ harness


Lucien950 (approved):
sexy

PR #1239, opened 2024-04-28
Calibrate Current Sensor

will-chaba:
### Changelist 

Current Sensor parameters updated to match the new sensor, calibrated using a DC power supply at 4A with 10 turns (40A relative to the sensor)

### Testing Done

Tested using method mentioned above

### Resolved Tickets

SW-68

### DIFF
--- a/firmware/quadruna/BMS/src/io/io_tractiveSystem.c
+++ b/firmware/quadruna/BMS/src/io/io_tractiveSystem.c
@@ -10,19 +10,19 @@
 // Voltage divider for the TS+ voltage sensing
 #define TS_VOLTAGE_DIV (20e+3f / (6 * 1e+6f + 30e3f + 20e3f))
 
-// Offset voltage of output 1. Found to be 2.45V through testing
-#define OUTPUT_1_OFFSET (2.45f)
-// Sensitivity of output 1: 40mV/A
-#define OUTPUT_1_SENSITIVITY (1.0f / 40e-3f)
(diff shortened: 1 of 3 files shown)


firmware/quadruna/BMS/src/io/io_tractiveSystem.c
```diff
@@ -120,26 +120,26 @@ float io_tractiveSystem_getCurrentLowResolution()
...
+    //                                                                  1
+    // Current = (DHAB S/124 Output 1 Voltage - Offset Voltage) x ---------------
     //                                                             Sensitivity
-    //                                              1k + 2.2k
-    // HSNBV-D06 Output 2 Voltage = ADC Voltage x --------------
-    //                                                 2.2k
+    //                                              33k + 60.4k
+    // DHAB S/124 Output 2 Voltage = ADC Voltage x --------------
+    //                                                 60.4k
     // Offset Voltage = 2.5V
```
KelmLelm, line 140: Should this be a flat offset of 2.558V? Sorry for nitpick

KelmLelm: Oh and same for output voltage 1

will-chaba: 2.5V is from the datasheet, updated the comment to make this clear

KelmLelm (approved):

PR #1242, opened 2024-05-03
Accumulator Integration

will-chaba:


will-chaba:
Remember to update the segment thermistor indexes

Lucien950:
@will-chaba there was an IOC merge conflict, please check that

PR #1244, opened 2024-05-04
Larak/lookup generate

larakawasme:
### Changelist 
Script to make json lookup for telem


forbesashli (approved):
Looks great to me. Good work

PR #1252, opened 2024-05-07
Quadruna CAN Charging

KelmLelm:
### Changelist 

- Added dedicated BRUSA "board" for sending and receiving can msgs from the charger. 
- Added logic to have BMS control starting and stopping charging over can.
- Added default values for charging currents, voltages and mains currents
- Added translation functions from big to little endian (and vice versa) and a function that uses it to convert floats for can msgs so that we can set values directly in the code without having to do a manual conversion to have the charger interpret the value we want.
- Added threshold of cycles we run after manual charging stop signal is received and telling the charger to stop charging to stop the BMS from immediately exiting charge state like we had seen testing with thruna
- Changed all (I think) io functions to do with charger to can functions.


can_bus/quadruna/BRUSA/BRUSA_tx.json
```diff
@@ -0,0 +1,21 @@
...
+                "start_bit": 6,
+                "bits": 1
+            },
+            "PowerLimitedWarning": {
+                "start_bit": 5,
+                "bits": 1
+            }
+        }
+    }
+}
```
will-chaba, line 21: We should eventually decode errors if time allows

firmware/quadruna/BMS/src/app/states/app_allStates.c
```diff
@@ -29,8 +29,9 @@ uint32_t    iso_spi_state_counter = 0;
 
 void app_allStates_runOnTick1Hz(void)
 {
-    bool charger_is_connected = io_charger_isConnected();
-    app_canTx_BMS_ChargerConnected_set(charger_is_connected);
+    bool charger_is_connected = app_canRx_BRUSA_IsConnected_get();
+    // Feel like this wouldn't be needed if the BRUSA broadcasts over CAN
```
will-chaba, line 33: I tend to agree. The behaviour we want is to remain in Init State if the BMS detects it is plugged in via the charging harness outside the car (Check for consistent missing heartbeats perhaps?), then transition to charging state if the debug message is sent

firmware/quadruna/BMS/src/app/states/app_chargeState.c
```diff
@@ -1,18 +1,46 @@
...
 #include "io_airs.h"
 #include "io_charger.h"
 
 // 0.05C is standard for a boundary to consider full charge
-#define CURRENT_AT_MAX_CHARGE (0.05f * C_RATE_TO_AMPS)
+#define C_RATE_FOR_MAX_CHARGE (0.05f)
 #define MAX_CELL_VOLTAGE_THRESHOLD (4.15f)
+#define CURRENT_AT_MAX_CHARGE (C_RATE_FOR_MAX_CHARGE * C_RATE_TO_AMPS)
+
+static uint16_t sixteenByteEndianSwap(uint16_t can_signal)
```
will-chaba, line 11: Technically it's two bytes, but maybe a function name more akin to the use case would be better eg:

```suggestion
static uint16_t canMsgEndianSwap(uint16_t can_signal)
```

firmware/quadruna/BMS/src/app/states/app_chargeState.c
```diff
@@ -26,12 +54,12 @@ static void chargeStateRunOnTick100Hz(void)
...
         const bool external_shutdown_occurred = !io_airs_isNegativeClosed();
         const bool charging_enabled           = app_canRx_Debug_StartCharging_get();
-        const bool is_charger_connected       = io_charger_isConnected();
+        const bool is_charger_connected       = app_canRx_BRUSA_IsConnected_get();
 
         bool has_charger_faulted = false;
         if (globals->ignore_charger_fault_counter >= CYCLES_TO_IGNORE_CHGR_FAULT)
         {
-            has_charger_faulted = io_charger_hasFaulted();
+            has_charger_faulted = app_canRx_BRUSA_Error_get();
```
will-chaba, line 62: Perhaps we should also find a way to detect if the brusa has dropped off the bus? Technically the brusa should disable itself, but always safe to have measures on both ends

firmware/quadruna/BMS/src/app/states/app_chargeState.h
```diff
@@ -8,6 +8,12 @@
 
 // Time to allow charger to disable before opening airs
 #define CHARGER_SHUTDOWN_TIMEOUT (100U)
+#define PREMATURE_EXIT_FILTER_THRESHOLD (20U)
+
+// Setting these to testing, will need to be changed later.
+#define INITIAL_CHARGING_VOLTAGE (330.0f)
+#define INITIAL_CHARGING_CURRENT (8.0f)
+#define INITIAL_MAX_MAINS_CURRENT (8.0f)
```
will-chaba, line 16: Fine for now, but once working let's try and do multi-stage charging profiles

firmware/quadruna/BMS/src/app/states/app_initState.c
```diff
@@ -45,17 +46,16 @@ static void initStateRunOnTick100Hz(void)
     {
         const bool air_negative_closed = io_airs_isNegativeClosed();
         const bool ts_discharged       = app_tractiveSystem_getVoltage() < TS_DISCHARGED_THRESHOLD_V;
+        const bool missing_hb          = app_heartbeatMonitor_checkFaults();
 
         if (air_negative_closed && ts_discharged)
         {
-            const bool charger_connected         = io_charger_isConnected();
+            const bool charger_connected         = app_canRx_BRUSA_IsConnected_get();
```
will-chaba, line 53: See comment in allStates.c about desired behaviour

firmware/quadruna/BMS/src/app/states/app_allStates.c
will-chaba: Also, maybe in case we accidentally go to drive state, have a condition to return to init if the charger is connected? IDK might be overkill

firmware/quadruna/CRIT/boot/cubemx/CRIT_boot.ioc.md5
```diff
@@ -1 +0,0 @@
-2d762bad1b1a74592e25cb8336ff2930
```
will-chaba, line 1: don't forget to build-all embedded and push the .md5's 😛

will-chaba (commented):
Looks good, couple of small things but mostly comments about future development

firmware/quadruna/BMS/src/app/app_tractiveSystem.c
```diff
@@ -27,7 +27,7 @@ void app_tractiveSystem_broadcast()
 {
     const float ts_voltage = app_tractiveSystem_getVoltage();
     const float ts_current = app_tractiveSystem_getCurrent();
-    const float ts_power   = ts_voltage * ts_current;
+    const float ts_power   = ts_voltage * ts_current / 1000.0f;
```
will-chaba, line 30: nit: lil somthin' like this
```suggestion
    #define W_TO_KW 1.0e-3f
    const float ts_power_kw   = ts_voltage * ts_current * W_TO_KW;
```

firmware/quadruna/BMS/src/app/states/app_initState.c
```diff
@@ -20,6 +20,7 @@ static void initStateRunOnEntry(void)
     // Should always be opened at this point from other states, this is only for redundancy since we really don't want
     // AIR+ closed in init
     io_airs_openPositive();
+    io_charger_enable(true);
```
will-chaba, line 23: This doesn't do anything anymore does it? Remove.

firmware/quadruna/RSM/boot/cubemx/Inc/main.h
```diff
@@ -58,6 +58,16 @@ extern "C"
...
+#define LED_Pin GPIO_PIN_13
+#define LED_GPIO_Port GPIOC
+#define BRAKE_LIGHT_EN_3V3_Pin GPIO_PIN_5
+#define BRAKE_LIGHT_EN_3V3_GPIO_Port GPIOA
+#define FR_STBY_Pin GPIO_PIN_6
+#define FR_STBY_GPIO_Port GPIOA
+#define RAD_FAN_EN_Pin GPIO_PIN_1
+#define RAD_FAN_EN_GPIO_Port GPIOB
+#define ACC_FAN_EN_Pin GPIO_PIN_10
+#define ACC_FAN_EN_GPIO_Port GPIOB
```
will-chaba, line 70: This was an accident, revert this

firmware/quadruna/RSM/boot/cubemx/Src/main.c
```diff
@@ -312,9 +312,40 @@ static void MX_GPIO_Init(void)
...
+    GPIO_InitStruct.Pull  = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /*Configure GPIO pins : RAD_FAN_EN_Pin ACC_FAN_EN_Pin */
+    GPIO_InitStruct.Pin   = RAD_FAN_EN_Pin | ACC_FAN_EN_Pin;
+    GPIO_InitStruct.Mode  = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Pull  = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
```
will-chaba, line 348: revert this too

will-chaba (approved):
Fix tests and small changes

can_bus/quadruna/BRUSA/BRUSA_tx.json
```diff
@@ -0,0 +1,21 @@
...
+            "Error": {
+                "start_bit": 6,
+                "bits": 1
+            },
+            "PowerLimitedWarning": {
+                "start_bit": 5,
+                "bits": 1
+            }
+        }
+    }
```
gtaharaedmonds, line 20: Not sure if this worked for you guys but this failed to decode the signals from the Brusa when we tried it yesterday. Also, there's a few other messages that aren't on here, maybe add a TODO to support them later?

firmware/quadruna/BMS/src/app/states/app_chargeState.c
```diff
@@ -1,18 +1,50 @@
...
 #define MAX_CELL_VOLTAGE_THRESHOLD (4.15f)
+#define CURRENT_AT_MAX_CHARGE (C_RATE_FOR_MAX_CHARGE * C_RATE_TO_AMPS)
+#define MAX_CHARGING_VOLTAGE (336.0f)
+// Each cell can handle 11.8A per the Datasheet, x3 in parallel = 35.4A, Setting as 15A for safety (limited by mains
+// current at this stage)
+#define MAX_CHARGING_CURRENT (15.0f)
+
+static uint16_t canMsgEndianSwap(uint16_t can_signal)
+{
+    uint16_t swapped_signal = (uint16_t)(can_signal >> 8) | (uint16_t)(can_signal << 8);
```
gtaharaedmonds, line 16: the one thing I didn't bother adding to jsoncan smh

scripts/canup
gtaharaedmonds, line 1: accidental change?

gtaharaedmonds (approved):
The Brusa vitals signals weren't decoding properly when we tested yesterday. But otherwise charging worked as intended.

firmware/quadruna/BMS/test/test_stateMachine.cpp
```diff
@@ -359,14 +359,29 @@ TEST_F(BmsStateMachineTest, charger_connected_can_msg_init_state)
...
+    SetInitialState(app_initState_get());
+    fake_io_airs_isNegativeClosed_returns(true);
+    app_heartbeatMonitor_blockFaults(false);
+    app_canRx_VC_Heartbeat_update(false);
+    app_canRx_RSM_Heartbeat_update(false);
+
+    LetTimePass(500U);
+
+    ASSERT_EQ(app_initState_get(), app_stateMachine_getCurrentState());
+}
```
gtaharaedmonds, line 383: Is this still not working or is this outdated?

KelmLelm: I fixed the test, forrgot about the comment. thanks for the reminder

scripts/canup
KelmLelm: not sure ewhy, but the  git submodule update --init --recursive wasn't updating to the latest version, so I went in to the submodule manually and pulled the latest version

can_bus/quadruna/BRUSA/BRUSA_tx.json
KelmLelm: Sounds good, it just involves decoding the entire can msg so that the signals can be properly displayed on pcan. There are quite a few signals in the msg though so will be done at a later time. Added note in chargestate.h

PR #1253, opened 2024-05-07
Bootloader Retransmit Packet Loss

Jugveer-Sandher:
### Changelist 
The bootloader now performs a handshake after processing each chunk (32 bytes) to verify packet integrity. This handshake involves comparing the address being programmed by the bootloader with the address being sent. If these addresses do not match, the last chunk is resent to the bootloader. 

### Testing Done
No testing done yet, will test on hardware.


firmware/boot/bootloader.c
```diff
@@ -213,6 +213,22 @@ _Noreturn void bootloader_runInterfaceTask(void)
...
+            io_can_pushTxMsgToQueue(&reply);
+        }
+        else if (command.std_id == LOST_CHUNK_ID && update_in_progress)
+        {
+            // Program the lost chunk to the address
+            uint32_t address = ((uint32_t)command.data[3] << 24) | ((uint32_t)command.data[2] << 16) |
+                               ((uint32_t)command.data[1] << 8) | (uint32_t)command.data[0];
+            uint32_t data = ((uint32_t)command.data[7] << 24) | ((uint32_t)command.data[6] << 16) |
+                            ((uint32_t)command.data[5] << 8) | (uint32_t)command.data[4];
+            bootloader_boardSpecific_program(address, data);
```
Jugveer-Sandher, line 230: Would programming 4-bytes of data, when retransmitting a lost chunk be too slow?

firmware/boot/bootloader.c
```diff
@@ -213,6 +213,22 @@ _Noreturn void bootloader_runInterfaceTask(void)
...
+            CanMsg   reply = { .std_id = CONFIRM_CHUNK_ID, .dlc = 1, .data = { (address == current_address) ? 1 : 0 } };
+            io_can_pushTxMsgToQueue(&reply);
+        }
+        else if (command.std_id == LOST_CHUNK_ID && update_in_progress)
+        {
+            // Program the lost chunk to the address
+            uint32_t address = ((uint32_t)command.data[3] << 24) | ((uint32_t)command.data[2] << 16) |
+                               ((uint32_t)command.data[1] << 8) | (uint32_t)command.data[0];
+            uint32_t data = ((uint32_t)command.data[7] << 24) | ((uint32_t)command.data[6] << 16) |
+                            ((uint32_t)command.data[5] << 8) | (uint32_t)command.data[4];
```
gtaharaedmonds, line 229: grabbing the least significant and most significant 32 bit words is easier than this. all you need is the pointer to the least significant byte and you can cast it to a `uint32_t` pointer. so:

```
uint32_t address = *((uint32_t*)(&command.data[0]))
uint32_t data = *((uint32_t*)(&command.data[4]))
```

Also note that arrays can basically just be treated as pointers to the first element, so:

```
uint32_t address = *((uint32_t*)command.data)
```

PR #1255, opened 2024-05-10
Dockerize Tracksight

Lucien950:
### Changelist 
moved the front and backend to be brought up via a docker container

### Testing Done
launched on different devices/OS systems 

### Resolved Tickets
n/a


forbesashli (approved):
thanks for the help and doing this, we really appreciate it. It looks great!

PR #1259, opened 2024-05-19
LittleFS Python Nav Support

Lucien950:
### Changelist 

- Added a dependency to Pipenv
- Added script from Jiajun from littlefs-python github repo


Jiajun-Huang (approved):
Looks good

PR #1264, opened 2024-05-19
Quick Heartbeat Rework

liam-ilan:
### Changelist 
- See second diagram on confluence on the heartbeat monitor docs; RSM will no longer shut down car, some connections turned into warnings.
image

### Testing Done
- None


can_bus/quadruna/BMS/BMS_alerts.json
```diff
@@ -35,17 +35,17 @@
         "RxOverflow" : {
             "id" :108,
             "description" : "Example"
+        },
+        "MissingRSMHeartbeat":{
+            "id" :152,
```
gtaharaedmonds, line 40: Should update the ID to match: 152 -> 108

can_bus/quadruna/BMS/BMS_alerts.json
```diff
@@ -35,17 +35,17 @@
...
+            "description" : "Example"
         }
     },
     "faults": {
         "MissingVCHeartbeat":{
             "id" :151,
             "description" : "Example"
         },
```
gtaharaedmonds, line 48: Doesn't look like the BMS needs to check the VC or the RSM heartbeat. It doesn't use any signals from either board anywhere, so I think we can just delete these checks. What do you think? @liam-ilan @Lucien950

can_bus/quadruna/CRIT/CRIT_alerts.json
```diff
@@ -54,5 +48,11 @@
...
             "description" : "Example"
         }
-}
+    },
+    "faults": {
+        "MissingBMSHeartbeat": {
+            "id" :251,
+            "description" : "Example"
+        }
+    }
```
gtaharaedmonds, line 57: What's the justification for leaving the BMS missing heartbeat a fault and moving the rest to warnings? Not against it, just wanna make sure there is a good reason

gtaharaedmonds (changes requested):
Had a few questions/concerns about specific transitions, but these are definitely steps in the right direction. Nice.

gtaharaedmonds:
Fixed in https://github.com/UBCFormulaElectric/Consolidated-Firmware/pull/1282
So closing this one.

PR #1268, opened 2024-05-21
CAN logging with logfs

gtaharaedmonds:
### Changelist 

Change the VC from using littlefs for CAN message logging to logfs. Tested on the VC and was seeing 5-10% CPU load while logging the entire CAN bus. Still needs more reliability testing but results looked promising so far.

### Testing Done

Basic testing with the car on jacks. Nothing broke obviously, but still needs more testing.


Jiajun-Huang (approved):
Looks good!!

PR #1269, opened 2024-05-21
Consolidate BMS CAN Messages

will-chaba:
### Changelist 

Consolidate BMS CAN Messages, including the deletion of un-used debugging messages to reduce bus load and ease burden on data logging

- Combined signals where possible
- Reduced cycle time where possible

### Testing Done

- Existing unit tests updated to match new signals


can_bus/quadruna/BMS/BMS_tx.json
```diff
@@ -259,7 +252,7 @@
         "description": "BMS Tractive System Data",
         "signals": {
             "TractiveSystemVoltage": {
-                "resolution": 0.0001,
+                "resolution": 0.01,
```
will-chaba, line 255: our voltage sense is not accurate enough to warrant this resolution

can_bus/quadruna/BMS/BMS_tx.json
```diff
@@ -271,82 +264,19 @@
...
             "AvailablePower": {
-                "resolution": 0.1,
+                "resolution": 1,
                 "min": 0,
                 "max": 100000,
                 "unit": "W"
             }
         }
     },
-    "AuxThermistors": {
```
will-chaba, line 307: these are not currently installed, can be added back later for future testing

firmware/quadruna/BMS/src/app/app_thermistors.c
```diff
@@ -28,10 +28,11 @@ void app_thermistors_updateAuxThermistorTemps()
...
-    app_canTx_BMS_ThermTemp3_set(therm_temps[3]);
-    app_canTx_BMS_ThermTemp4_set(therm_temps[4]);
-    app_canTx_BMS_ThermTemp5_set(therm_temps[5]);
+    // Re-enable if aux thermistors installed 
+    // app_canTx_BMS_ThermTemp0_set(therm_temps[0]);
+    // app_canTx_BMS_ThermTemp1_set(therm_temps[1]);
+    // app_canTx_BMS_ThermTemp2_set(therm_temps[2]);
+    // app_canTx_BMS_ThermTemp3_set(therm_temps[3]);
+    // app_canTx_BMS_ThermTemp4_set(therm_temps[4]);
+    // app_canTx_BMS_ThermTemp5_set(therm_temps[5]);
```
will-chaba, line 37: Aux Thermistors not currently installed, no sense loading the CAN bus with their messages

gtaharaedmonds: You should delete them from the JSON instead, commenting out the setters doesn't disable them from the bus. They'll just be transmitted with their start value (probably all zeros).

scripts/canup
gtaharaedmonds, line 1: Was this on purpose?

gtaharaedmonds (changes requested):
Yeep this is definitely an improvement

Lucien950 (approved):
i am happy when gus is happy

scripts/canup
will-chaba: No definitely not, will revert

firmware/quadruna/BMS/src/app/app_thermistors.c
will-chaba: Ah okay gotcha. Will do.

gtaharaedmonds (approved):
yeep

PR #1063, opened 2023-11-02
Refactor BMS

gtaharaedmonds:
### Summary

Refactor the BMS to be more in line with the new coding structure (like we have on the DIM)

### Changelist 

BMS app and IO layer follows the new structure.

### Testing Done

Tested while accumulator was partly dissasembled, things not yet tested include:
- Charging
- Spin Wheels
- TS Voltage Sense
- TS Current Sense
- Pre-Charge

BMS code has been modified to not allow it to enter Precharge state for any reason.

### Checklist
*Please change `[ ]` to `[x]` when you are ready.*
- [x] I have read and followed the code conventions detailed in [README.md](../README.md) (*This will save time for both you and the reviewer!*).
- [x] If this pull request is longer then **500** lines, I have provided *explicit* justification in the summary above explaining why I *cannot* break this up into multiple pull requests (*Small PR's are faster and less painful for everyone involved!*).

PR #1270, opened 2024-05-21
Update Fakegen README based on changes from #1063

will-chaba:
### Changelist 

Update fakegen README with details about return value overrides and callback functions


gtaharaedmonds (approved):
LGTM

PR #1271, opened 2024-05-22
Unpacking CAN frames with JSONCAN in Python

gtaharaedmonds:
### Changelist 

Add the ability to unpack raw CAN frames to signal values in JSONCAN Python. This can be used with telemetry to decode received packets, or used to pull logged frames from the SD card.

### Testing Done

Was able to successfully decode CAN frames logged to the SD card, and got expected values.


forbesashli (approved):
tis fine

PR #1273, opened 2024-05-22
Specify data capture times for SD card and wireless telem in JSONCAN

gtaharaedmonds:
### Changelist 

Add the ability to specify the data capture rate for both SD card logging and wireless telemetry. The schema is as follows:
```json
"<msg name>": {
    <other signals omitted>
    "data_capture": {
        "log_cycle_time": 1000,
        "telem_cycle_time": 1000
     }
}
```

`log_cycle_time` is the cycle time at which the message will be logged to the SD card. If `null`, this message will not be logged. This signal can be omitted and the normal message cycle time will be used instead.

`telem_cycle_time` is the cycle time at which the message will be transmitted via wireless telemetry. If `null`, this message will not be sent. This signal can be omitted and the normal message cycle time will be used instead.

### Testing Done

TODO


Lucien950:
Winning

Lucien950 (approved):
Winning Fr

PR #1274, opened 2024-05-22
Reenable watchdogs on Quadruna

gtaharaedmonds:
### Changelist 

Re-enable all watchdogs on Quadruna.

### Testing Done

Works on VC, but still need to test the rest.


Lucien950 (approved):
Tested and rested

PR #1280, opened 2024-05-23
Tracksight Backend Influx Refactor

Lucien950:
### Changelist 

Does what it says on the tin


software/tracksight/backend/app/process/flask_apps/http_app.py
```diff
@@ -2,75 +2,75 @@
...
-    :returns JSON-ified data response.
-    """
-    response = jsonify(data)
-    response.headers.add("Access-Control-Allow-Origin", "*")
-    return response
-
-
 @app.route("/")
-def hello_world() -> str:
+def hello_world():
```
gtaharaedmonds, line 14: smh why'd you delete my types

Lucien950: merge conflict :( also they don't return responses, they return dictionaries now

PR #1282, opened 2024-05-26
Simplify heartbeats + annotate alerts

gtaharaedmonds:
### Changelist 

A few changes that should make driving less fragile.

1. Changed the heartbeat dependency graph. Now, board A only checks the heartbeat of board B if A uses a signal from board B. There currently are a few instances of boards checking other boards' heartbeats, but then not actually using signals from them. In my mind this is unnecessary and could only lead to unnecessary faults down the line. (check me on this!). This for example means the FSM checks no other boards heartbeats.

2. Moved various faults on the VC related to the LV battery and currents to warnings. If `AccumulatorRailOvercurrentFault` and `BatteryRailOvercurrentFault` is set, we'd probably see total loss of the LV bus, so no need to just exit drive state. If just one is set, there is no immediate reason to stop driving. If there is a `LvChargeFault` or `BoostControllerFault`, there is also no need to exit drive state (we'd probably only notice the consequences of the failure on the next power cycle). By making these faults instead of warnings we reduce the chances of spurious shutdowns.

3. Added descriptions for all alerts.

### Testing Done

Unit testing only.


can_bus/quadruna/RSM/RSM_alerts.json
```diff
@@ -82,6 +82,10 @@
             "id": 452,
             "description": "VC Missing Heart beat"
         },
+        "MissingBMSHeartbeat": {
+            "id": 452,
+            "description": "BMS Missing Heart beat"
```
will-chaba, line 87: nit
```suggestion
            "description": "BMS Missing Heartbeat"
```

firmware/quadruna/FSM/src/app/app_mainState.c
```diff
@@ -11,30 +11,18 @@
 #include "app_suspension.h"
 #include "app_loadCell.h"
 #include "app_shdnLoop.h"
-#include "app_heartbeatMonitor.h"
 
 void mainStateRunOnTick100Hz(void)
 {
+    app_canTx_FSM_Heartbeat_set(true);
```
will-chaba, line 17: I'm a bit confused what's going on here. Do any boards consume the FSM heartbeat? if not why not remove the messages altogether?

firmware/quadruna/VC/test/test_vcBaseStateMachineTest.h
```diff
@@ -110,14 +110,14 @@ class VcBaseStateMachineTest : public BaseStateMachineTest
     // config for heartbeat monitor (can funcs and flags)
     // VC relies on FSM, RSM, BMS, CRIT
```
will-chaba, line 111: Update this

firmware/quadruna/VC/test/test_vcBaseStateMachineTest.h
```diff
@@ -110,14 +110,14 @@ class VcBaseStateMachineTest : public BaseStateMachineTest
...
     bool heartbeatMonitorChecklist[HEARTBEAT_BOARD_COUNT] = {
-        [BMS_HEARTBEAT_BOARD] = true, [VC_HEARTBEAT_BOARD] = false,  [RSM_HEARTBEAT_BOARD] = true,
+        [BMS_HEARTBEAT_BOARD] = true, [VC_HEARTBEAT_BOARD] = false,  [RSM_HEARTBEAT_BOARD] = false,
```
will-chaba, line 113: It was my understanding that the RSM was required as it had the coolant sensors to check for over-heating conditions, and if it drops off we may not shutdown due to overtemp condition

will-chaba (commented):
Looks good, but think the VC should still depend on RSM to be able to shutdown due to over-heating

firmware/quadruna/FSM/src/app/app_mainState.c
gtaharaedmonds: Yes, the VC does for accelerator pedals and brakes.

firmware/quadruna/VC/test/test_vcBaseStateMachineTest.h
gtaharaedmonds: this is still true

firmware/quadruna/VC/test/test_vcBaseStateMachineTest.h
gtaharaedmonds: good call, will fix this.

will-chaba (approved):

PR #1283, opened 2024-05-27
BMS Watchdog Fix

will-chaba:
### Changelist 

Watchdog for 1kHz task timing out due to status being checked before ever getting checked in. Added check-in during initialization to fix. 

Fix applied to all watchdogs.

### Testing Done

BMS now able to initialize without faulting


gtaharaedmonds (approved):
LGTM

PR #1286, opened 2024-05-30
Cell Balancing quick fix

DJ90864:
### Testing Done
tested with the BMS in the battery pack, discharging is working


will-chaba (approved):
Made a quick fix, but looks good!

PR #1287, opened 2024-06-01
Undo BMS Software Fault Latch for Charging

will-chaba:
### Changelist 

Undo software fault latch now that hardware latch is functioning properly


gtaharaedmonds (approved):
LGTM

PR #1289, opened 2024-06-02
Implement BSPD Test Circuit Over CAN

will-chaba:
### Changelist 

Implement BSPD Test Circuit over CAN, including Debug message to enable and added flag to TractiveSystem message to indicate when circuit detects sufficient current flow to trigger BSPD if brake signal received

### Testing Done
\

Tested on BMS


gtaharaedmonds (approved):
LGTM

PR #1290, opened 2024-06-02
blown fuse check ported to quadruna

KelmLelm:
### Changelist 

- Added a blown fuse check algorithm that activates above a certain threshold and checks difference in cell voltages to see if a cell fuse was blown. Sets the alert of a blown cell fuse if this is true. The algorithm that does this is called in a new checkWarnings function in app_accumulator.

### Testing Done

- Unit tests

IMPORTANT:

- Needs to be tested to see if it falsely triggers due to poor balancing of cells. If it does, then either balance better or just set a flag warning that it has been detected. If it doesn't falsely trigger, we can add it to the list of things that can shutdown the car.

### Resolved Tickets

https://ubcformulaelectric.atlassian.net/browse/SW-20


will-chaba (commented):
Looks good! Let's make sure it still runs on the hardware!

will-chaba:
Closing for now

PR #1293, opened 2024-06-03
Fix BSPD warning and handle it in VC

philam2001:
### Changelist 

- Move bspd handling to the VC
- Set torque to 0 instead of pedal map to 0 when bspd sets
- Add apps_init in task.c for fsm

### Testing Done

- [x] Unit testing done
- [ ] Drive state testing needed


will-chaba (approved):
Test on car, looks good 👍

gtaharaedmonds (approved):
gonna merge now and test this afternoon

PR #1294, opened 2024-06-03
Fix logging filesystem failing to initialize with high number of logs

gtaharaedmonds:
### Changelist 

On master, if there are a significant number of CAN logs present on the SD card (somewhere in the range 50-100) the watchdog trips at startup. This means the VC never gets past init. 

Root cause: Currently, mounting the filesystem and starting logging takes about 250ms for every 100 logs on the SD card, and scales linearly. There is no significant correlation with the size of each log file. Since the watchdog trips if not checked in every 200ms, it just simply takes too long. Note that we start a new log after every boot.

Solution: I think anything less than a second of extra delay at bootup is not a big deal, so I don't think this motivates any big changes to our logging. There is no way to stop the STM32 watchdog peripheral once its started, so instead I added a `tasks_preInitWatchdog` function which runs after all peripherals are initialized, but before the watchdog is. So, we initialize logging in here and then can proceed like normal.

Also, I added some CAN signals so there will be visibility on when we've logged a significant amount of data.

### Testing Done

Tested on VC #2. Able to init logging in ~500ms total with 200 logs without the watchdog tripping, and the VC behaves like normal after that. Tested with large files as well (>>100MB) without any slowdown. Performance is approximately O(N) at all values tested. 

Also, interestingly it seems that different SD cards are faster than others. The KOOTION 64GB Micro SDXC card is approximately 2x as fast as the Lexar 300x 32GB SDHC card.

### DIFF
--- a/can_bus/quadruna/VC/VC_alerts.json
+++ b/can_bus/quadruna/VC/VC_alerts.json
@@ -63,6 +63,10 @@
         "BoostControllerFault": {
             "id": 560,
             "description": "Boost controller has faulted."
+        },
+        "HighNumberOfCanDataLogs": {
+            "id": 561,
+            "description": "High number of CAN data logs (>200) on the SD card, which slows down mounting the filesystem."
         }
     },
     "faults": {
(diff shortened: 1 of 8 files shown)


philam2001 (approved):
Screen Shot 2024-06-03 at 8 42 03 PM

PR #1295, opened 2024-06-03
Decode logfs CAN logs for Tracksight

gtaharaedmonds:
### Changelist 

Added a script to decode raw log files written to the SD card. This uses JSONCAN to unpack the signals and writes them to a CSV file. For how to use, run:

```
python3 software/tracksight/backend/scripts/decode_raw_log.py --help
```

These CSV log files are (by default) written to `software/tracksight/backend/data/`. Files in here are mounted to the Docker backend when running in local mode, and so can be uploaded to the local InfluxDB server for viewing on the Tracksight frontend. Pass the paths to the unpacked CSV log files to `run_local.sh` and they will be uploaded automatically.

Also, added a few CAN signals/warnings so we can check the state of logging over CAN.

### Testing Done

Tested this whole pipeline on my laptop with VC #2.


environment/requirements.txt
```diff
@@ -18,12 +18,10 @@ rich
 
 # Tracksight backend dependencies.
 flask
-Definitions
 flask_cors
 pandas
 flask_socketio
 pytest
-asammdf
```
philam2001, line 26: rip

firmware/quadruna/VC/src/tasks.c
```diff
@@ -539,7 +529,11 @@ _Noreturn void tasks_runCanTx(void)
         CanMsg tx_msg;
         io_can_popTxMsgFromQueue(&tx_msg);
         io_telemMessage_pushMsgtoQueue(&tx_msg);
-        io_canLogging_pushTxMsgToQueue(&tx_msg);
```
philam2001, line 542: Didn't you add this in the other pr? #1294

firmware/quadruna/VC/src/tasks.c
```diff
@@ -562,7 +556,6 @@ _Noreturn void tasks_runCanRx(void)
         CanMsg rx_msg;
         io_can_popRxMsgFromQueue(&rx_msg);
         io_telemMessage_pushMsgtoQueue(&rx_msg);
-        io_canLogging_pushTxMsgToQueue(&rx_msg);
```
philam2001, line 565: Same with this

gtaharaedmonds: same with here too. should have fixed these in other PR but was feeling lazy

PR #1296, opened 2024-06-04
Setup infra for wireless Tracksight

gtaharaedmonds:
### Changelist 

Setup infra for wireless telem. No more web sockets.


larakawasme:
very nice

PR #1298, opened 2024-06-05
Disable open wire check because we're getting module communication errors with it enabled

gtaharaedmonds:
### Changelist 

Does what it says on the tin.

### Testing Done

Copy-pasted our testing today from slack:

What we were seeing before this:
- Talked to each module directly in isolation (setting ACCUMULATOR_NUM_SEGMENTS to 1). No issues on any segments from ISOSPI A or ISOSPI B.
- Tested each individual cable in isolation, also no issues.
- However, when we hooked up 3 modules we'd get communication error faults after ~10 seconds. I think once or twice we managed to get to 4 segments, but never 5
 
With this change:
- Disabled open wire check. No idea why this would break anything.
- After that, I could pretty reliably get up to 4 modules talking. 5 still work, but after rearranging the cables I found a configuration that did work.
- Absolutely no idea whats going on here. Is there anything that would cause the communication to be unreliable? Maybe sketchy connector, noise, etc...

I was able to run the battery for a few minutes without any communication faults (with open wire check disabled).

### DIFF
--- a/firmware/quadruna/BMS/src/app/states/app_allStates.c
+++ b/firmware/quadruna/BMS/src/app/states/app_allStates.c
@@ -76,13 +76,15 @@ bool app_allStates_runOnTick100Hz(void)
                 iso_spi_state_counter = 0;
             }
 
-            const uint32_t cycles_to_measure =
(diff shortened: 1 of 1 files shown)


will-chaba (approved):
LGTM

PR #1299, opened 2024-06-06
Send live data to local influx and send to front end

larakawasme:
### Changelist 
Sending live data to local influx
### Testing Done
it works
### Resolved Tickets
YEA


gtaharaedmonds (approved):
go go go

PR #1306, opened 2024-06-10
Crit LED fixes

philam2001:
### Changelist 

- Enable TV light when TV is being used 
- Enable Regen only when Regen is used and available (not faulted)

### DIFF
--- a/firmware/quadruna/VC/src/app/states/app_driveState.c
+++ b/firmware/quadruna/VC/src/app/states/app_driveState.c
@@ -139,8 +139,6 @@ static void driveStateRunOnTick100Hz(void)
     {
         apps_pedal_percentage  = app_regen_pedalRemapping(apps_pedal_percentage);
         sapps_pedal_percentage = app_regen_pedalRemapping(sapps_pedal_percentage);
-        app_canTx_VC_RegenEnabled_set(app_regen_available());
-        app_canTx_VC_Warning_RegenNotAvailable_set(!app_regen_available());
     }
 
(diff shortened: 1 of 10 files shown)


firmware/quadruna/VC/src/app/states/app_driveState.c
```diff
@@ -168,6 +174,9 @@ static void driveStateRunOnExit(void)
     // Disable buzzer on exit drive.
     io_efuse_setChannel(EFUSE_CHANNEL_BUZZER, false);
     app_canTx_VC_BuzzerOn_set(false);
+
+    app_canTx_VC_RegenEnabled_set(false);
+    app_canTx_VC_TorqueVectoringEnabled_set(false);
```
gtaharaedmonds, line 179: I believe `VC_TorqueVectoringEnabled` is also not populated anywhere?

gtaharaedmonds (changes requested):
Sort of unrelated but can you consolidated `RegenEnabled` and `TorqueVectoringEnabled` into the same CAN message? Currently there each 1 bit signals with their own message, which adds unnecessary bus load. Better would be to put them in the same message.

firmware/quadruna/VC/src/app/states/app_driveState.c
philam2001: It's in the TV init function

gtaharaedmonds (approved):
LGTM

Lucien950 (approved):

PR #1309, opened 2024-06-11
Fix Cell Balancing

will-chaba:
### Changelist 

In the extreme case of an over-charged pack, cell-balancing cannot be performed to bring the pack back into spec. This PR adjusts the MAX_CELL_VOLTAGE threshold depending on if balancing is enabled or not, allowing a slightly over-charged pack to be brought back below 4.2v per cell through discharging.

Also fixes bug in the isoSPI state machinecaused by recent change with OWC.

### Testing Done

Tested on pack

### DIFF
--- a/firmware/quadruna/BMS/src/app/app_accumulator.c
+++ b/firmware/quadruna/BMS/src/app/app_accumulator.c
@@ -1,6 +1,5 @@
 #include "app_accumulator.h"
-#include "states/app_balancingState.h"
-#include "states/app_initState.h"
+#include "states/app_chargeState.h"
 #include "app_canTx.h"
 #include "app_canRx.h"
 #include "app_canAlerts.h"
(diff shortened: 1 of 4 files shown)


firmware/quadruna/BMS/src/app/app_accumulator.c
```diff
@@ -479,11 +480,18 @@ bool app_accumulator_checkFaults(void)
         min_allowable_cell_temp = MIN_CELL_CHARGE_TEMP_DEGC;
     }
 
+    const State *current_state = app_stateMachine_getCurrentState();
+
+    const bool cell_balancing_enabled =
+        app_canRx_Debug_CellBalancingRequest_get() && current_state != app_chargeState_get();
+
+    const float max_cell_voltage = cell_balancing_enabled ? MAX_CELL_VOLTAGE_BALANCING : MAX_CELL_VOLTAGE_NOMINAL;
```
gtaharaedmonds, line 488: Maybe leave a comment why we do this cuz its cursed

gtaharaedmonds (approved):
yep sure

PR #1311, opened 2024-06-11
Jiajun/disable unused sensor

Jiajun-Huang:
### Changelist 
Disable load cell and suspension


gtaharaedmonds (changes requested):
Can you also disable the load cell and suspension related alerts for FSM and RSM? Otherwise LGTM

gtaharaedmonds (approved):
LGTM

PR #1312, opened 2024-06-11
Fixes for disabling JSONCAN messages/alerts + improve FSM alerts

gtaharaedmonds:
Fix my shitty code + add more FSM alerts so they give us more information about brake OCSC.

### DIFF
--- a/can_bus/quadruna/FSM/FSM_alerts.json
+++ b/can_bus/quadruna/FSM/FSM_alerts.json
@@ -64,9 +64,17 @@
             "id": 315,
             "description": "Rear brake pressure out of range."
         },
-        "BrakePressureSensorOCSC": {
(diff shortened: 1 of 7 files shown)


can_bus/quadruna/FSM/FSM_alerts.json
```diff
@@ -77,10 +85,12 @@
             "description": "Right suspension sensor voltage out of acceptable range (OCSC)."
         },
         "LoadCell1OCSC": {
+            "disabled": true,
```
gtaharaedmonds, line 88: We only allow 21 faults/warnings per board, and so have to disable these to meet that threshold. @Jiajun-Huang is working on disabling them anyway so this should be fine

philam2001 (approved):

PR #1316, opened 2024-06-13
I made a bug in can unpack maybe its fixed now oops

gtaharaedmonds:


Lucien950 (approved):
NEGATIVE SIGNAL BITS????

PR #1319, opened 2024-06-15
No SD SOC

will-chaba:
### Changelist 

Add back SOC in most basic form, resets based on min cell voltage on initial startup. Not accurate when power cycled when cells are not settled.

SD should be re-enabled and set to run in its own low-priority task in the future.

### Testing Done

Off-target tests pass. Will test at charging before endurance.


gtaharaedmonds (approved):
we're so back

PR #1321, opened 2024-06-15
Inverter Updates Configs

philam2001:
### Changelist 

- Update inverter config from comp, increase max speed here since we were capped at 3500 RPM


Lucien950 (approved):

Lucien950:
@philam2001 close?

Lucien950 (approved):
excellent

PR #1322, opened 2024-06-17
Spicy Charging (Overide for Brusa to charge past 336V)

will-chaba:
### Changelist 

Brusa currently faults if a pack voltage of 336V is seen, and we had no way of adjusting this value over CAN. This PR adds the ability to due so with the addition of an additional CAN flag to override the safety parameters and request a voltage of up to 345V from the Brusa.

This theoretically should be no danger as the voltage requested from the Brusa is a fault threshold (IE stop charging if this voltage is seen. The BMS itself will shut off charging if a cell voltage above 4.2V is seen, and so there is no real danger of increasing this limit, but until that can be verified I wanted to implement the increased limit in a way that it must be explicitly applied.

### Testing Done

None, @DJ90864 @KelmLelm @1mackenziekyle this is on you, godspeed.

### DIFF
--- a/firmware/quadruna/BMS/src/app/states/app_chargeState.c
+++ b/firmware/quadruna/BMS/src/app/states/app_chargeState.c
@@ -6,9 +6,12 @@
 #define C_RATE_FOR_MAX_CHARGE (0.05f)
 #define MAX_CELL_VOLTAGE_THRESHOLD (4.15f)
 #define CURRENT_AT_MAX_CHARGE (C_RATE_FOR_MAX_CHARGE * C_RATE_TO_AMPS)
-#define MAX_CHARGING_VOLTAGE (336.0f)
+#define MAX_CHARGING_VOLTAGE_NOMINAL (336.0f)
+// Allows pack to reach max cell voltage before BRUSA faults due to max pack voltage (likely due inaccurate ADC reading
+// from BRUSA)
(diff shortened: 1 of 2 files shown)


firmware/quadruna/BMS/src/app/states/app_chargeState.c
```diff
@@ -138,7 +142,11 @@ static void chargeStateRunOnTick100Hz(void)
                                            ? app_canRx_Debug_ChargingVoltageTargetValue_get()
                                            : INITIAL_CHARGING_VOLTAGE;
 
-        if (IS_IN_RANGE(0.0f, MAX_CHARGING_VOLTAGE, charging_voltage))
+        const float max_charging_voltage = app_canRx_Debug_ChargingVoltageSafetyOverride_get()
```
1mackenziekyle, line 145: I might be missing something but in the CAN msg is only allocated 1 bit, but its cast to a float here. Should it be allocated 32 bits in the JSONCAN?

will-chaba: The can message is just used as input to the conditional operator, max_charging_voltage is either MAX_CHARGING_VOLTAGE_OVERIDE or MAX_CHARGING_VOLTAGE_NOMINAL.

Lucien950 (approved):
o7 quadruna battery pack

PR #1323, opened 2024-06-17
Temporary Fix for Max Power

will-chaba:
### Changelist 

Derate requested power according to the current fault threshold and pack power. This should still be re-worked to not be a fault, but needs to be done on the VC side. Detailed current limiting will be implemented by @DJ90864 

### Testing Done

None. @DJ90864 you're up, I just can't sit still today


Lucien950 (approved):
I AGREE!

PR #1335, opened 2024-09-07
Efuse Retry Protocol

Jugveer-Sandher:
### Changelist 
Efuse retry protocol:

Implemented retry handler for efuse faults.

The retry handler works as follows: If an efuse faults, we wait for 1 second (100 100Hz cycles) to obtain a current sum. We then check if the efuse is still faulting. If so, we disable this efuse and any dependent efuses (ex. all efuses depend on if LV is working). After a 1 second debounce time, we re-enable the efuse and wait another 1 second while getting the current sum. If the efuse is now working, we re-enable all dependent efuses. If the efuse is still faulting, we throw a fault.

Still pending: Getting current values for the efuses.

### Testing Done
Implemented unit testing


Lucien950:
we're writing a new one for quint

PR #1336, opened 2024-09-10
Dikshit/chimera load bank power supply upgrade

DJ90864:
### Changelist 

Just fixed some sample code to reflect two channels on the Power supply instead of one
### Testing Done

Tested on Power Supply


Lucien950 (approved):
W short PR

PR #1337, opened 2024-09-11
changed the clock source from SysTick to TIM2 and enabled FPU in Free…

DJ90864:
…RTOS

### Changelist 

Quick fix to chage the clock source for HAL Timing from SysTick to TIM2
Also enabled FPU in FreeRTOS for faster floating point operations


Lucien950 (approved):
Double check it flashes properly before merging

firmware/quadruna/BMS/src/cubemx/Inc/FreeRTOSConfig.h
```diff
@@ -51,13 +51,12 @@
...
 #include <stdint.h>
 extern uint32_t SystemCoreClock;
-void            xPortSysTickHandler(void);
 #endif
 #ifndef CMSIS_device_header
 #define CMSIS_device_header "stm32h7xx.h"
 #endif /* CMSIS_device_header */
 
-#define configENABLE_FPU 0
+#define configENABLE_FPU 1
```
KelmLelm, line 59: I don't actually know much about this, but I was wondering why there was a change with the FPU.

DJ90864: I notiched it wasn't enabled in FreeRTOS config, so enabled it since we need it for SoC calculations

KelmLelm: Okay, I see. Makes sense makes sense

KelmLelm (approved):
Don't know too much about the autogenerated code, but it lgtm.

PR #1338, opened 2024-09-17
TCP Bootloading

Jiajun-Huang:
### Changelist 
[ ] add a simple TCP on bootloader ensure reliable transport in boot-loading process 
[ ] Add a TCP to send event trigger, arbitrary length message.


Lucien950:
changed through different PR

PR #1342, opened 2024-09-22
need python 3.8 - 3.11, might as well enforce it??

peterjinweigu:


Lucien950 (approved):
Didn't even know this was a thing

PR #1343, opened 2024-09-27
Remove Vscode Prelaunch Tasks for Tests

liam-ilan:
### Changelist 
- Removed prelaunch tasks from tests. This was causing an error to be thrown since we no longer use Vscode tasks.

### Testing Done
- Ran tests with no warnings in Vscode.

### Resolved Tickets
- https://ubcformulaelectric.atlassian.net/browse/FIRM-94
- [FIRM-99](https://ubcformulaelectric.atlassian.net/browse/FIRM-99)

[FIRM-99]: https://ubcformulaelectric.atlassian.net/browse/FIRM-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
         "build_fw_dev/firmware/quadruna/VC/quadruna_VC_boot.elf",
         "build_fw_dev/firmware/quadruna/VC/quadruna_VC_app.elf",
       ],
-      // "preLaunchTask": "Build Embedded: VC (Quadruna)",
+      "preLaunchTask": "Build Embedded: VC (Quadruna)",
       // cortex configurations
       "request": "launch",
       "type": "cortex-debug",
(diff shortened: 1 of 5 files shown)


Lucien950:
can you investigate if there's a way to launch cmake build tasks from launch.json? cause there's a few ways I see this system working

- control shift b tasks which trigger cmake builds
- control shift b tasks as per usual as last year, but manual preset encoding
- leaving the responsibility to the user to build before running
- cmake launch configurations?? surely has to be a thing
- other tech I do not know about

Lucien950 (changes requested):
can you investigate if there's a way to launch cmake build tasks from launch.json? cause there's a few ways I see this system working

- control shift b tasks which trigger cmake builds
- control shift b tasks as per usual as last year, but manual preset encoding
- leaving the responsibility to the user to build before running
- cmake launch configurations?? surely has to be a thing
- other tech I do not know about

I just don't know too much about this atm

liam-ilan:
Will do! - found this https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/debug-launch.md

Lucien950:
hot I will make changes based on that in [FIRM-99](https://ubcformulaelectric.atlassian.net/browse/FIRM-99)

philam2001 (commented):
Wait does load cmake only load it for the build test fw?

Lucien950 (approved):

PR #1346, opened 2024-10-05
Placed tire model in new folder software/vehicle_modelling_and_controls

1mackenziekyle:
### Changelist 
- This should not change any of the existing software. This is an isolated tool.
- This is a quarter-car tire model that responds reasonably to torque inputs. 
- Data used: 
  - Tire coefficients from Heiko's data anaylsis for tire modelling, which used Tire Test Consortium (TTC)'s data of our same tire model. 
  - Estimate of 100kg vehicle (subject to change) 
  - Estimate of wheel inertia being 0.5 kg*m^2 (subject to change)
- Limitations: Inaccurate parameters right now. 
- Only has a step torque input. This will be replaced by a controller soon

### Testing Done
Quarter-car tire model seems to respond reasonably to a variety of torque inputs, having low slip when the torque is reasonable and high slip when the torque is extremely high.


Lucien950 (approved):
Let's get this moneyimage

philam2001 (approved):
LGTM

shayana18 (approved):

PR #1353, opened 2024-10-16
Display busload on build message

Jiajun-Huang:
### Changelist 
Specify the generator in Cmake preset to Unix Makefile so that is does not generate VS project (and use MSVC) on Windows.
Add a Shared variable in quadruna cmake ( `set(SHARED_CAR "quadruna")` )
Print the estimate the bus load and print them in build message.

### Testing Done
image


CMakePresets.json
```diff
@@ -46,6 +48,7 @@
 		{
 			"name": "FW Test",
 			"description": "Test build for the Firmware, including debug information and no optimizations.",
+			"generator": "Unix Makefiles",
```
Lucien950, line 51: please make your own `CMakeUserPresets.json`

scripts/windisk.py
```diff
@@ -1,34 +0,0 @@
-from littlefs import LittleFS, UserContextWinDisk
```
Lucien950, line 1: :)

scripts/code_generation/jsoncan/calc_bus_load.py
```diff
@@ -19,22 +22,29 @@ def msg_payload_bits(msg):
...
+        msg = msg[1]
+        if msg.is_periodic():
+            freq_hz = 1000 / msg.cycle_time
+            bits_per_s += msg_payload_bits(msg) * freq_hz
+
+    return bits_per_s / float(can_bit_rate) * 100
+
+def report_bus_load(canDatabase : CanDatabase, can_bit_rate):
+    estimated_bus_load = calculate_bus_load(canDatabase, can_bit_rate)
+    if estimated_bus_load > 70:
```
Lucien950, line 37: magic number, please extract

firmware/cmake/embedded.cmake
```diff
@@ -250,5 +250,7 @@ function(embedded_image
             WORKING_DIRECTORY ${REPO_ROOT_DIR}
             DEPENDS ${GENERATE_IMAGE_SCRIPT} ${APP_HEX_PATH} ${BOOT_HEX_PATH}
     )
+
+    log_bus_load(${IMAGE_HEX})
```
Lucien950, line 254: why log it after the image is created vs putting this alongside the other jsoncan generation code?

firmware/quadruna/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(QUADRUNA_DIR "${FIRMWARE_DIR}/quadruna")
 set(SHARED_QUADRUNA_DIR "${QUADRUNA_DIR}/shared")
 set(SHARED_EMBEDDED_QUADRUNA_DIR "${SHARED_QUADRUNA_DIR}/src")
-
+set(SHARED_CAR "quadruna")
```
Lucien950, line 4: we want to be able to build for multiple cars at the same time. maybe pass this information as an argument?

Lucien950 (changes requested):
Mostly good

firmware/quadruna/CMakeLists.txt
Jiajun-Huang: I am not sure is it correct of not, but if we have a different car and have a CMakeLists for it `firmware/{anotherCar}/CMakeList.txt` with a statement `set(SHARED_QUADRUNA_DIR, "anotherCar")` , then the `${SHARED_QUADRUNA_DIR}` in this function will be `"anotherCar"`; so I guess it will not prevent building multiple cars at same time? I observed that building dev boards will give empty value for `${SHARED_QUADRUNA_DIR}`.

firmware/cmake/embedded.cmake
Jiajun-Huang: I didn't put it with jsoncan builds because CMake cache json CAN build and if json CAN build is cache, then it doesn't run the script which make the message does not always print. I know that putting line in here is a little weird and I just came up with a better way to do it.

firmware/quadruna/CMakeLists.txt
Lucien950: oh that is probably something we should be doing more, you're right.

Lucien950 (approved):

Lucien950 (approved):

PR #1359, opened 2024-10-19
`JLinkGDBServerCLExe` -> `JLinkGDBServer` on MacOS to enable Brew Installs

liam-ilan:
### Changelist 
We now invoke `JLinkGDBServer` instead of `JLinkGDBServerCLExe` in `launch.json` for OSX. The installer from the web provides the `JLinkGDBServerCLExe` alias to `JLinkGDBServer`, but the `brew` install only provides `JLinkGDBServer`. 

Also fixes typo in path for `h7dev_boot.elf` in `launch.json`.

**This PR makes it possible to install jlink segger via `brew`**.

### Testing Done
- [x] Flashed H7Dev

### Resolved Tickets
- None

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,7 @@
         "serverpath": "JLinkGDBServerCLExe",
       },
       "osx": {
-        "serverpath": "JLinkGDBServerCLExe",
+        "serverpath": "JLinkGDBServer",
       },
       "windows": {
         "serverpath": "JLinkGDBServerCL",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -254,7 +254,7 @@
       "cwd": "${workspaceRoot}",
       "executable": "build_fw_dev/firmware/dev/h7dev/h7dev.hex",
       "symbolFiles": [
-        "build_fw_dev/firmware/boot/h7boot/h7dev_boot.elf",
```
Lucien950, line 257: What

Lucien950 (changes requested):
I assume this breaks the install from website method?

.vscode/launch.json
liam-ilan: thats where it is 🤷

Lucien950 (approved):
make sure this doesn't break on other people's computers. (tell them to install with brew?)

PR #1361, opened 2024-10-22
BMS Diagnostics Mode WIP

mattiaswong6:
### Changelist 
This is a WIP for review.
Enable tracking of all 16 cell voltages in Segment 0 when Diagnostics Mode enabled.

### Testing Done
None.


Lucien950:
Is this redundant to #1367?

mattiaswong6:
Yes, this is redundant as we moved work to another branch with a more accurate name.

PR #1366, opened 2024-10-27
Update SD Logging Scripts

philam2001:
### Changelist 

- Add args in decode script to take in range of csv files to decode (since they're all numbered and before you had to input one by one in manually)
- Add arg in decode script to convert to mf4 automatically instead of sending another command
- Add a try except in the csv_to_mf4 block to make sure it deletes corrupted or unfinished mf4 files 
- Increase SD byte size to 7.5GB
### Testing Done

- Tested scripts on my laptop
- Need to test SD byte size update


Lucien950 (approved):
looks decent.

Jiajun-Huang (approved):
LGTM

PR #1377, opened 2024-11-05
Optimise logging execution

Jiajun-Huang:
### Changelist 
move enqueue operation to Can_rx task
batch the write to the file system
add support of logfs parsing on windows machine
fix timestamp on csv_to_mf4 convertion

before opt
image
image

after opt
image
image

### Testing Done
Awaiting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@
 *.ini
 
 # Python
+*.pyd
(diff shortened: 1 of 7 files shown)


Lucien950 (approved):
APPROVED!

PR #1379, opened 2024-11-09
Chimera Watchdog Repair

liam-ilan:
### Changelist 
- Last year, when doing board bringup/chimera testing, we disabled the watchdog. Bringing back watchdog broke Chimera, since the chimera version of tasks was not petting watchdog.
- This PR removes `io_chimera_sleepIfTaskEnabled()` in favour of `io_chimera_block()` and `io_chimera_task()`.
  - `io_chimera_task` implements a simple watchdog pinging task.
  - `io_chimera_block` operates like `io_chimera_sleepIfTaskEnabled`, and sleeps the task if chimera is enabled.

### Testing Done
- [ ] Tested in BMS @DJ90864

### DIFF
--- a/firmware/quadruna/BMS/src/tasks.c
+++ b/firmware/quadruna/BMS/src/tasks.c
@@ -346,8 +346,6 @@ void tasks_init(void)
 
 _Noreturn void tasks_run1Hz(void)
 {
-    io_chimera_sleepTaskIfEnabled();
-
     static const TickType_t period_ms = 1000U;
     WatchdogHandle         *watchdog  = hw_watchdog_allocateWatchdog();
     hw_watchdog_initWatchdog(watchdog, RTOS_TASK_1HZ, period_ms);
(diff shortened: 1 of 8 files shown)


Lucien950:
I think we need a far deeper fix for this. The watchdog code from before I just raw ported.

Lucien950:
changes I think should be here:
1. refresh_watchdog removed in favour of bringing a reference to the watchdog into the watchdog monitor
2. refresh HAL watchdog only when there are no timeouts detected
3. if there are no watchdogs registered, then the watchdogs are vacuously fine, and so we should trigger a refresh
4. in chimera task, we trigger watchdog monitor on the vacuous case

Lucien950 (changes requested):
can we close this? i think it's better if we should disable watchdog when chimera is running rather than infinitely petting it

liam-ilan:
Not going to happen - to be fixed in Chimera V2

PR #1380, opened 2024-11-09
SystemView Interrupts

Lucien950:
### Changelist 

This PR enables developers to view interrupts in the SystemView application.


Jiajun-Huang (approved):
Really interesting staff. Should we apply the same change to other boards?

PR #1381, opened 2024-11-10
Increase RTT/Sysview Stack Size

Lucien950:

### DIFF
--- a/firmware/third_party/sysview/SEGGER_SYSVIEW_Conf.h
+++ b/firmware/third_party/sysview/SEGGER_SYSVIEW_Conf.h
@@ -49,7 +49,7 @@
 
 File    : SEGGER_SYSVIEW_Conf.h
 Purpose : SEGGER SystemView configuration file.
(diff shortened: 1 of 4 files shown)


firmware/third_party/sysview/SEGGER_SYSVIEW_Conf.h
```diff
@@ -0,0 +1,83 @@
...
+#define SEGGER_SYSVIEW_CONF_H
+
+/*********************************************************************
+ *
+ *       Defines, configurable
+ *
+ **********************************************************************
+ */
+
+#ifdef DEBUG
```
Jiajun-Huang, line 77: Just a minor thing, it looks like the debug macro is created in Cmake nonetheless. It is better to make the debug macro only for when the CMAKE_BUILD_TYPE = Debug? And we can give more buffer to the config when we build for debug and give less buffer or not at all for release build.

Jiajun-Huang (approved):
Looks solid.

PR #1382, opened 2024-11-15
Tune Traction Control PID based on Simulink

shayana18:
### Changelist 
      - added can message for correction torque for visibility
      - implemented clamp based windup guard and derivative noise filtering in pid compute
      -  added wind up and noise filtering inputs to PID struct
      - mapped error slip to error torque for controller to be outputting a correction torque rather than a correction slip 
      - applied PID constants in config (not sure if we want to keep this const if we want it to dynamically change)

### Testing Done
- needs to be done

### DIFF
--- a/firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
+++ b/firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
@@ -6,8 +6,11 @@ void app_pid_init(PID *pid, const PID_Config *conf)
     pid->Kp         = conf->Kp;
     pid->Ki         = conf->Ki;
     pid->Kd         = conf->Kd;
+    pid->derivative_filtering_coeff = NAN; 
     pid->prev_input = 0.0f;
     pid->integral   = 0.0f;
+    pid->integral_windup_min = NAN; 
+    pid->integral_windup_max = NAN;
     pid->out_min    = conf->out_min;
     pid->out_max    = conf->out_max;
 }
@@ -22,8 +25,23 @@ void app_pid_init(PID *pid, const PID_Config *conf)
 float app_pid_compute(PID *pid, float setpoint, float input)
 {
     pid->error = setpoint - input;
-    pid->integral += pid->error;
-    pid->derivative = (input - pid->prev_input);
+
+    if(app_gaurd_init(pid)){ // if initialized implement the gaurd
+        if(pid->integral_windup_max > pid->integral && pid->integral > pid->integral_windup_min){
+            pid->integral += pid->error;
+        }
+    } 
+    else{
+        pid->integral += pid->error;
+    }
+
+    if(app_filter_init(pid)){
+        pid->derivative = pid->derivative_filtering_coeff * (pid->derivative / (1 + pid->derivative_filtering_coeff));
+    }
+    else{
+        pid->derivative = (input - pid->prev_input);
+    }
+
     pid->prev_input = input;
     float output    = pid->Kp * pid->error + pid->Ki * pid->integral - pid->Kd * pid->derivative;
 
@@ -35,4 +53,12 @@ void app_pid_requestReset(PID *pid)
 {
     pid->prev_input = 0.0f;
     pid->integral   = 0.0f;
+}
+
+bool app_gaurd_init(PID *pid){
+    return (pid->integral_windup_max != NAN && pid->integral_windup_min != NAN); 
+}
+
+bool app_filter_init(PID *pid){
+    return (pid->derivative_filtering_coeff != NAN); 
 }
--- a/firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
+++ b/firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
@@ -6,7 +6,7 @@ void app_pid_init(PID *pid, const PID_Config *conf)
     pid->Kp         = conf->Kp;
     pid->Ki         = conf->Ki;
     pid->Kd         = conf->Kd;
-    pid->derivative_filtering_coeff = NAN; 
+    pid->derivative_filtering_coeff = 0.0f; // will implement a filter later  
     pid->prev_input = 0.0f;
     pid->integral   = 0.0f;
     pid->integral_windup_min = NAN; 
@@ -26,7 +26,7 @@ float app_pid_compute(PID *pid, float setpoint, float input)
 {
     pid->error = setpoint - input;
 
-    if(app_gaurd_init(pid)){ // if initialized implement the gaurd
+    if(app_guard_init(pid)){ // if initialized implement the gaurd
         if(pid->integral_windup_max > pid->integral && pid->integral > pid->integral_windup_min){
             pid->integral += pid->error;
         }
@@ -34,13 +34,10 @@ float app_pid_compute(PID *pid, float setpoint, float input)
     else{
         pid->integral += pid->error;
     }
+    
+    // removing filtering for now 
 
-    if(app_filter_init(pid)){
-        pid->derivative = pid->derivative_filtering_coeff * (pid->derivative / (1 + pid->derivative_filtering_coeff));
-    }
-    else{
-        pid->derivative = (input - pid->prev_input);
-    }
+    pid->derivative = (input - pid->prev_input);
 
     pid->prev_input = input;
     float output    = pid->Kp * pid->error + pid->Ki * pid->integral - pid->Kd * pid->derivative;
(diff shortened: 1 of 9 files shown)


firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
```diff
@@ -22,8 +25,23 @@ void app_pid_init(PID *pid, const PID_Config *conf)
...
+    if(app_gaurd_init(pid)){ // if initialized implement the gaurd
+        if(pid->integral_windup_max > pid->integral && pid->integral > pid->integral_windup_min){
+            pid->integral += pid->error;
+        }
+    } 
+    else{
+        pid->integral += pid->error;
+    }
+
+    if(app_filter_init(pid)){
```
philam2001, line 38: i dont think this is the proper low pass filtering you're trying to achieve since it doesnt use the derivatives history. ur calculating this based on input rn not error. I don't rly know it at the top of my head rn, but just gpt it or something

firmware/quadruna/VC/src/app/vehicle_dynamics/app_tractionControl.c
```diff
@@ -21,18 +21,31 @@ void app_tractionControl_computeTorque(TractionControl_Inputs *inputs, TractionC
...
 
     // NOTE: k strictly in range [-1 0] to prevent exceeding power limit
-    outputs->torque_left_final_Nm  = (1.0f + k) * inputs->torque_left_Nm;
-    outputs->torque_right_final_Nm = (1.0f + k) * inputs->torque_right_Nm;
+
+    outputs->torque_left_final_Nm  =  inputs->torque_left_Nm + correction_torque; 
+    outputs->torque_right_final_Nm =  inputs->torque_right_Nm + correction_torque;
 }
 
 float app_tractionControl_computeSlip(float motor_speed_rpm, float front_wheel_speed_rpm)
```
philam2001, line 51: slip ratio can spike at vehicle speed (front wheel speed) at low speeds btw. u wanna do a check before this. i know this prevents div by 0 but still

firmware/quadruna/VC/src/app/vehicle_dynamics/app_tractionControl.c
```diff
@@ -21,18 +21,31 @@ void app_tractionControl_computeTorque(TractionControl_Inputs *inputs, TractionC
 
     float slip_ratio_max = fmaxf(slip_ratio_left, slip_ratio_right);
     float k              = app_pid_compute(pid, SLIP_RATIO_IDEAL, slip_ratio_max);
```
philam2001, line 23: u can also CLAMP k [-1,0]

firmware/quadruna/VC/src/app/vehicle_dynamics/app_tractionControl.c
```diff
@@ -21,18 +21,31 @@ void app_tractionControl_computeTorque(TractionControl_Inputs *inputs, TractionC
...
+    static float correction_torque; 
+
+    // NOTE: we want to map K to a vlaue that we can multiple with input torque. K is based on the slip error 
+    // we want to map it to torque error, we also want to  make sure we use the torque associated to the wheel from which slip max was taken
+
+    if( slip_ratio_max == slip_ratio_left){
```
philam2001, line 29: are both wheels gonna have uniform conditions? probably not right so why would u apply correction torque to both? also where did u get this correcttion torque equation from that uses wheel speed/ vehicle speed. Could do it like this ?
```
if (slip_ratio_max == slip_ratio_left) {
correction_torque = k * inputs->torque_left_Nm;    
outputs->torque_left_final_Nm  = inputs->torque_left_Nm + correction_torque;
    outputs->torque_right_final_Nm = inputs->torque_right_Nm;  
} else {
    correction_torque = k * inputs->torque_right_Nm; 
outputs->torque_right_final_Nm = inputs->torque_right_Nm + correction_torque;
    outputs->torque_left_final_Nm  = inputs->torque_left_Nm; 
}
```

philam2001 (changes requested):

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
```diff
@@ -6,8 +6,11 @@ void app_pid_init(PID *pid, const PID_Config *conf)
     pid->Kp         = conf->Kp;
     pid->Ki         = conf->Ki;
     pid->Kd         = conf->Kd;
+    pid->derivative_filtering_coeff = NAN;
```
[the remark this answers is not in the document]
shayana18: yeah I see what your saying, I can change that

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
```diff
@@ -36,3 +54,11 @@ void app_pid_requestReset(PID *pid)
     pid->prev_input = 0.0f;
     pid->integral   = 0.0f;
 }
+
+bool app_gaurd_init(PID *pid){
+    return (pid->integral_windup_max != NAN && pid->integral_windup_min != NAN);
```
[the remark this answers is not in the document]
shayana18: I did not know that

firmware/quadruna/VC/src/app/vehicle_dynamics/app_torqueVectoring.c
```diff
@@ -44,8 +46,11 @@ void app_torqueVectoring_init(void)
 {
     app_canTx_VC_TorqueVectoringEnabled_set(true);
     app_pid_init(&pid_power_correction, &PID_POWER_CORRECTION_CONFIG);
+
+    // traction control PID init
     app_pid_init(&pid_traction_control, &PID_TRACTION_CONTROL_CONFIG);
-    traction_control_inputs.pid = &pid_traction_control;
+    traction_control_inputs.pid                             = &pid_traction_control;
+    traction_control_inputs.pid->derivative_filtering_coeff = 50.0;
```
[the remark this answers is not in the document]
shayana18: simulink. When I was tuning I found that when the coefficient was larger than 50 we say a lot more oscillation. The controller becomes less sensitive to that as you lower it so I made it 50

firmware/quadruna/VC/src/app/vehicle_dynamics/app_torqueVectoring.c
```diff
@@ -112,6 +117,12 @@ void app_torqueVectoring_handleAcceleration(void)
     // Traction Control
     if (run_traction_control)
     {
+        // Have to set windup here as wheel speed is not defined when init
+        traction_control_inputs.pid->integral_windup_max =
```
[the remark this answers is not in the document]
shayana18: from the way I was previously doing it yes, but im going to make it a constant cause I think that was wrong

firmware/quadruna/VC/src/app/vehicle_dynamics/app_tractionControl.c
shayana18: yeah that correction torque calculation is wrong. what I was trying to do is map it make to a torque value but I did it wrong... in the future it would be good to map it to a vehicle speed using vehicle speed = (1 +- slip) (wheel_speed) depending on acceleration, then turn that into an angular velocity and then place it in a queue of sleep values from which we can take two entries from to have like a derivative to get acceleration then convert that into a torque with T = M * and_a 


I will change it to what you said

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
```diff
@@ -22,8 +25,20 @@ void app_pid_init(PID *pid, const PID_Config *conf)
 float app_pid_compute(PID *pid, float setpoint, float input)
 {
```
philam2001, line 26: You might b able to autogen c code from simulink u can see how they probably would handle pid.

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
```diff
@@ -22,8 +25,20 @@ void app_pid_init(PID *pid, const PID_Config *conf)
...
+    else{
+        pid->integral += pid->error;
+    }
+    
+    // removing filtering for now 
+
     pid->derivative = (input - pid->prev_input);
+
     pid->prev_input = input;
     float output    = pid->Kp * pid->error + pid->Ki * pid->integral - pid->Kd * pid->derivative;
```
philam2001, line 43: Why is this minus derivative btw

philam2001 (changes requested):

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
shayana18: I thought you and Kyle had a reason for doing that

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
shayana18: yeah I can look at that. I did basically use their equation for the filtering stuff tho

firmware/quadruna/VC/src/app/vehicle_dynamics/app_pid.c
philam2001: This was before my time 😀

philam2001: You may need to fix pid cuz not sure what was written before but I don't think this makes sense

philam2001 (changes requested):
You need to add some unit testing for your changes to see if your code will reflect your intentions

Lucien950:
i'm killing

PR #1386, opened 2024-11-20
Dimos Flutter Migration

Lucien950:


peterjinweigu (approved):
how are is there still cmake for flutter? is it because of the io/can layer?

Lucien950:
cmake because flutter eventually compiles down to C

peterjinweigu:
😀 🤣 😭

PR #1392, opened 2024-11-24
FSM Bringup

kryptoish:
### Changelist 

- Set up CMake and STM32 for FSM on quintuna

### Testing Done

none

### Resolved Tickets

none


firmware/quintuna/FSM/src/tasks.h
```diff
@@ -0,0 +1,9 @@
+
+#ifndef TASKS_H
+#define TASKS_H
+
+class tasks
+{
+};
+
+#endif // TASKS_H
```
Lucien950, line 9:
```suggestion
```

Lucien950 (changes requested):

Lucien950 (approved):

PR #1393, opened 2024-11-24
Quintuna VC Bring Up

Jugveer-Sandher:
### Changelist 
Configured CMake and CubeMx.

### Testing Done
Can load CMake and build binaries.


Lucien950 (approved):
goat status

PR #1398, opened 2024-11-26
SDCard Logging Hotfix

Lucien950:
### Changelist 

- reverts change moving sdcard logging from the can queue
- new method for can filtering in quintuna

### DIFF
--- a/firmware/quadruna/VC/src/tasks.c
+++ b/firmware/quadruna/VC/src/tasks.c
@@ -61,6 +61,13 @@ static bool     sd_card_present            = true;
 static void canRxCallback(CanMsg *rx_msg)
 {
     io_can_pushRxMsgToQueue(rx_msg); // push to queue
+
+    // Log the message if it needs to be logged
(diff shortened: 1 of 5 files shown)


firmware/quintuna/CRIT/src/hw/hw_cans.cpp
```diff
@@ -4,7 +4,12 @@
 
 namespace hw::can
 {
-const CanBus can1{ &hcan1, [](const CanMsg *canMsg) { io::can1queue.pushRxMsgToQueue(canMsg); } };
+const CanBus can1{ &hcan1, [](const CanMsg *canMsg)
+                   {
+                       if (!io_canRx_filterMessageId(canMsg->std_id))
```
Lucien950, line 9: consider this a hardware optimization, because w/best practices this is supposed to happen in the canrx task

firmware/quadruna/VC/src/tasks.c
Jiajun-Huang, line 1: Yes, I forgot there is a filter.

PR #1415, opened 2025-01-04
New Code Style

Lucien950:


gtaharaedmonds (commented):
How much of this has been tested on hardware?

firmware/boot/bootloader.c
```diff
@@ -97,37 +94,28 @@ _Noreturn static void modifyStackPointerAndStartApp(const uint32_t *address)
...
+    for (;;) {
     }
 }
 
-static BootStatus verifyAppCodeChecksum(void)
-{
-    if (*&__app_code_start__ == 0xFFFFFFFF)
-    {
+static BootStatus verifyAppCodeChecksum(void) {
+    if (__app_code_start__ == 0xFFFFFFFF) {
```
gtaharaedmonds, line 102: I vaguely remember you can't do this with linker script variables, and you have to keep the `*&`. I might be wrong but you should try this on hardware.

firmware/boot/bootloader.c
```diff
@@ -104,30 +111,25 @@ _Noreturn static void modifyStackPointerAndStartApp(const uint32_t *address)
 
 static BootStatus verifyAppCodeChecksum(void)
 {
-    if (*&__app_code_start__ == 0xFFFFFFFF)
+    if (__app_code_start__ == 0xFFFFFFFF)
```
Lucien950, line 114:
```suggestion
    if (*&__app_code_start__ == 0xFFFFFFFF)
```

Lucien950:
@gtaharaedmonds all tested on hardware

firmware/shared/src/io/io_can.c
gtaharaedmonds, line 1: Why move this from HW to IO? It's basically a wrapper around STM's CAN driver so I feel like that makes it HW?

firmware/shared/src/io/io_can.c
```diff
@@ -1,111 +1,113 @@
 #include "io_can.h"
-#include <string.h>
+#undef NDEBUG
 #include <assert.h>
```
gtaharaedmonds, line 3: Why do you want non-debug asserts here? Should probably make a dedicated `always_assert`/`debug_assert` or something like that to make this more clear. Otherwise I feel like this'll get confused a lot.

firmware/quadruna/BMS/boot/cubemx/Src/main.c
```diff
@@ -105,7 +108,29 @@ void        runCanTxTask(void *argument);
...
+void HAL_FDCAN_RxFifo1Callback(FDCAN_HandleTypeDef *hfdcan, const uint32_t RxFifo1ITs)
+{
+    assert(hfdcan == &hfdcan1);
+    UNUSED(RxFifo1ITs);
+    CanMsg rx_msg;
+    if (!io_can_receive(&can, FDCAN_RX_FIFO1, &rx_msg))
+        // Early return if RX msg is unavailable.
+        return;
+    io_canQueue_pushRx(&rx_msg);
+}
```
gtaharaedmonds, line 133: Any reason you have the callbacks in `main.c` for FDCAN? They're all in `hw_can.c` for regular CAN

firmware/dev/h7dev/src/cubemx/Src/main.c
```diff
@@ -540,16 +532,6 @@ static void MX_GPIO_Init(void)
...
-static void can_msg_received_callback(CanMsg *rx_msg)
-{
-    // TODO: check gpio present
-    static uint32_t id = 0;
-    rx_msg->std_id     = id;
-    id++;
-    io_can_pushRxMsgToQueue(rx_msg);
-    io_canLogging_loggingQueuePush(rx_msg);
-}
-
```
gtaharaedmonds, line 552: Why was this stuff changed?

firmware/quadruna/BMS/src/tasks.c
```diff
@@ -259,40 +77,16 @@ void tasks_init(void)
...
-    // app_thermistors_init();
-
-    app_soc_init();
-    app_globals_init(&globals_config);
-    app_stateMachine_init(app_initState_get());
 
-    // broadcast commit info
-    app_canTx_BMS_Hash_set(GIT_COMMIT_HASH);
-    app_canTx_BMS_Clean_set(GIT_COMMIT_CLEAN);
+    app_globals_init();
```
gtaharaedmonds, line 89: this probably needs to be done in jobs right

firmware/shared/src/io/io_can.c
Lucien950: my reasoning:
1. liam wants to mock CAN for SIL testing, io is the layer at which we do mocking (basically it's entire purpose)
2. io_can handles a fundementally io object (CanMsg) which is handled at io or greater level. Having this object in hw layer was always an issue for us
3. I think while it interacts with the HAL it is non-trivial enough to be considered a bit more than just a wrapper

firmware/shared/src/io/io_can.c
Lucien950: yes I agree. This is something I carried over from the old code.

firmware/quadruna/BMS/boot/cubemx/Src/main.c
Lucien950: i was doing a lot of thinking about this. I think there are a lot of good reasons to do both, but I think we should revert to old because of a recent change where canqueues are global (despite multiple busses). My understanding is that these two callbacks it regardless of how many CAN busses we are using?

firmware/quadruna/BMS/src/tasks.c
Lucien950: yes, but I think globals needs to be gone just in general

Lucien950: Only VC is really up to new spec, all the other boards I just changed so that they still compile.

firmware/dev/h7dev/src/cubemx/Src/main.c
Lucien950: breaks in new system

firmware/shared/src/io/io_fileSystem_logfs.c
```diff
@@ -140,60 +154,52 @@ FileSystemError io_fileSystem_init(void)
...
-        files_opened[fd] = false;
-        return logfsErrorToFsError(err);
+        files_opened[allocateFd()] = false;
+        return logfsErrorToFsError(
+            logfs_open(&fs, &files[allocateFd()], &files_cfg[allocateFd()], LOGFS_OPEN_RD_WR | LOGFS_OPEN_CREATE));
     }
 
-    return fd;
+    return allocateFd();
 }
```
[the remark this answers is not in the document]
gtaharaedmonds: We should probably test the CAN logging with these changes since that code is relatively fragile. I'm working on some other changes to it so I can test it on Saturday

firmware/shared/src/io/io_fileSystem_logfs.c
```diff
@@ -221,12 +223,15 @@ FileSystemError io_fileSystem_close(int fd)
...
-
+    CHECK_FILE_DESCRIPTOR(fd)
     return logfsErrorToFsError(logfs_sync(&fs, &files[fd]));
 }
+
+bool io_fileSystem_ready()
+{
+    // TODO recheck ready conditions
+    ready &= hw_sd_present();
+    return ready;
```
gtaharaedmonds, line 236: Hmm I don't we should latch this. It's plausible that if we hit some bumps or something the SD card will rattle and the GPIO might blip not present. The IO CAN logging file keeps track of errors and only disables logging if we error 10 times.

firmware/shared/src/io/io_can.c
gtaharaedmonds: Hmm I'd imagine that you'd mock out the CAN queues (io_can.c on master, looks like you renamed to io_canQueue.c)? I think you have to mock at this level since io_can.c uses a bunch of FreeRTOS primtiives.

And hw_can.c is currently only invoked from io_can.c on master, so not sure if it needs to be mocked out?

gtaharaedmonds: But oh yeah fair enough the CAN message struct is defined in hw_can.h. Could always move that to a separate file.

firmware/shared/src/io/io_can.c
gtaharaedmonds: ok maybe leave a TODO that this is something we should do?

gtaharaedmonds: I guess we could use `hw_error` for checks that we want to be run in non-debug builds too

firmware/quadruna/BMS/boot/cubemx/Src/main.c
gtaharaedmonds: Yeah I think these callbacks are always fired. And then you can check which `hfdcan` you get to check which bus its from?

firmware/dev/h7dev/src/cubemx/Src/main.c
gtaharaedmonds: hmm maybe unbreak instead of delete idk

firmware/quadruna/BMS/src/tasks.c
gtaharaedmonds: yeah thats fair. i think there was just a couple cases where I need to share data between multiple states which is the only reason globals ever exited. probably better ways of doing this

firmware/shared/src/io/io_fileSystem_logfs.c
```diff
@@ -140,60 +154,52 @@ FileSystemError io_fileSystem_init(void)
...
-
+    CHECK_FILE_DESCRIPTOR(fd)
     uint32_t num_read;
-    LogFsErr err = logfs_read(&fs, &files[fd], buffer, size, LOGFS_READ_END, &num_read);
-    if (err != LOGFS_ERR_OK || num_read != size)
+    if (logfs_read(&fs, &files[fd], buf, size, LOGFS_READ_END, &num_read) != LOGFS_ERR_OK || num_read != size)
     {
-        return logfsErrorToFsError(err);
+        return logfsErrorToFsError(logfs_read(&fs, &files[fd], buf, size, LOGFS_READ_END, &num_read));
     }
```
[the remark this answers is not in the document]
Lucien950: wait idk why i did this 😭

firmware/shared/src/io/io_can.c
Lucien950: yeah: also i don't think we got always_assert in C so we would have to make that macro ourselves, as in copy pasting the assert macro except not checking NDEBUG.

firmware/quadruna/BMS/boot/cubemx/Src/main.c
Lucien950: I just copy pasted and I remembered that the shared code cannot check which canhandle is active because it differs by board. Otherwise, we would have to handle multibus behaviour in (global) canQueue? seems like bad seperation of concerns.

firmware/quadruna/BMS/src/tasks.c
Lucien950: yeah, I think making a module which is more descriptive but has similar functionality at the end of the day is always better

firmware/shared/src/io/io_fileSystem_logfs.c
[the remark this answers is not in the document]
Lucien950: i have an autoinline key i am realizing i am probably accidentally pressing

firmware/shared/src/io/io_fileSystem_logfs.c
Lucien950: I think I am preserving old functionality, because before if the hw_sd_present is false at any point it stops the logging thread etc.

liam-ilan (commented):
CAN stuff looks LGTM - Will be a hot second until I put it into the SIL. Deferring to @AmirTajaddodi @DJ90864.

AmirTajaddodi (approved):
LGTM

PR #1420, opened 2025-01-11
Chimera Over USB (Chimera V2)

liam-ilan:
> Note to reviewers - This PR isn't small... Highly recommend reading through [`firmware/chimera_v2/README.md`](https://github.com/UBCFormulaElectric/Consolidated-Firmware/blob/liam-ilan/chimera_v2/firmware/chimera_v2/README.md), it'll take you across the codebase to the parts that really matter.

### Changelist 
This PR Implements Chimera V2 (aka. Chimera over USB).

To allow larger messages than 64 bytes (the max CDC USB packet size, definitely overkill 😅), `hw_usb` dumps any received data over usb into an `osMessageQueue`, composed of the bytes received over USB. Since we have no indicator for the start and end of a stream, and all bytes are valid in RPC, we define a length-delimited format. See the diagram below.

chimera_v2_high_res drawio

### Testing Done
- [x] Tested blinky on the F4 Dev Board (GPIO Output).
- [x] Tested on Windows.
- [x] Tested GPIO Read.
- [ ] Tested ADC Read.
- [ ] Tested with IWDG active.
- [ ] Tested I2C Peripherals.

### For Future PRs
When boards arrive, we might want to swap from the single `chimera_v2_main` to a dedicated task.

### Note About Install
Until this PR is merged, Chimera V2 is installed with
```sh
pip install 'git+https://github.com/UBCFormulaElectric/Consolidated-Firmware.git@liam-ilan/chimera_v2#egg=chimera-v2&subdirectory=firmware/chimera_v2'
```


firmware/dev/f4dev/src/cubemx/Src/main.c
```diff
@@ -531,25 +551,7 @@ void StartDefaultTask(void *argument)
...
-        // hw_uart_transmitPoll(&modem_uart, message, sizeof(message), 100); // fun string
-        for (num = 48; num < 57; num++)
-        {
-            predicData[0] = num;
-            hw_uart_transmitPoll(&modem_uart, predicData, sizeof(predicData), 100); // this is for 0->255
-            // sprintf((char *)message, "B%03dB", i); //Generate dynamic message for fun string
-            osDelay(1);
-        }
-    }
+    io_chimera_v2_main(id_to_gpio, NULL, CDC_Transmit_FS, 0, 0);
```
gtaharaedmonds, line 554: Not part of this PR but I'm not sure if the `hw_usb.h` is a great abstraction at the moment. You need to call the `hw_usb_init`, with a handle to STM's transmit func, and also remember to call `MX_USB_DEVICE_Init`. I think the idea of the HW layer is it's a one-stop shop to use a driver, so ideally it handles all initialization, and knows about the transmit function. Not major, definitely doesn't need to be changed in this PR.

firmware/shared/src/hw/hw_usb.c
```diff
@@ -34,23 +34,30 @@ bool hw_usb_checkConnection()
...
     uint8_t    res    = 0;
     osStatus_t status = osMessageQueueGet(rx_queue_id, &res, NULL, 100);
+
     if (status == osOK)
         return res;
+    else if (status == osErrorTimeout)
+        return 0;
+
+    LOG_WARN("usb queue pop returned non-ok status %d", status);
     return 0;
```
gtaharaedmonds, line 54: This is a bit suspicious, there's no way to discern between a failed operation (since it returns zero) vs. successfully dequeueing an actual zero value.

firmware/shared/src/hw/hw_usb.c
```diff
@@ -83,16 +90,16 @@ void hw_usb_transmit_example(uint8_t (*transmit_handle)(uint8_t *Buf, uint16_t L
...
+    // dump the queue
+    for (;;)
     {
         uint8_t result = hw_usb_recieve();
-        LOG_INFO("char %d: %c", i, (char)result);
-        osDelay(1000);
+        _LOG_PRINTF("%c", (char)result);
+        osDelay(100);
     }
 }
```
gtaharaedmonds, line 105: Nitpick: If you want to use `_LOG_PRINTF` maybe just remove the leading underscore. Was supposed to be for marking it as private but it does seem useful, so I'd remove the underscore.

firmware/chimera_v2/generate_proto.sh
Lucien950, line 1: cmake :)

firmware/chimera_v2/proto/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    PROTO_SRCS PROTO_HDRS 
+    shared.proto f4dev.proto
+)
+
+set(SRCS ${PROTO_SRCS})
+set(HDRS ${CMAKE_CURRENT_BINARY_DIR} ${NANOPB_INCLUDE_DIRS})
+
+# Register library.
+message("📚 Registering chimera_v2_proto_cm4/7 libraries")
+embedded_library(
```
Lucien950, line 15: interface library pls

firmware/chimera_v2/proto_autogen/f4dev_pb2.py
Lucien950, line 1: we ignore autogenned files?

firmware/chimera_v2/pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[project]
+name = "chimera_v2"
+version = "0.1"
+dependencies = ["grpcio-tools", "protoletariat", "pyusb"]
```
gtaharaedmonds, line 4: Can you document somewhere how to install these? I always forget lol, would be nice to have a tiny lil README about it

firmware/chimera_v2/chimera_v2.py
gtaharaedmonds, line 1: This lib looks very good

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -0,0 +1,176 @@
...
+#include <pb_decode.h>
+#include <pb_encode.h>
+#include "shared.pb.h"
+
+// Peripheral interfaces.
+#include "hw_adc.h"
+#include "hw_gpio.h"
+
+static Gpio       **id_to_gpio;
+static AdcChannel **id_to_adc;
```
Lucien950, line 17: double pointer?

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -0,0 +1,176 @@
...
+
+static pb_byte_t out_buffer[10000];
```
Lucien950, line 19: god motherfucking damn

Lucien950: boy fat as hell 😭

firmware/shared/src/hw/hw_usb.c
```diff
@@ -34,23 +34,30 @@ bool hw_usb_checkConnection()
...
+    uint8_t handle_status = usb_transmit_handle(msg, len);
+    if (handle_status != 0)
+        LOG_WARN("Chimera: USB handle returned %d status code instead of 0", handle_status);
 }
 
 uint8_t hw_usb_recieve()
 {
     // receive pops off the byte queue and returns
     uint8_t    res    = 0;
     osStatus_t status = osMessageQueueGet(rx_queue_id, &res, NULL, 100);
```
gtaharaedmonds, line 46: Just an FYI, FreeRTOS stream buffers are faster than queues if you just need to ship streams of bytes. But no worries if you're not having any issues right now since I don't think we're sending much data over USB.

https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/04-Stream-and-message-buffers/02-Stream-buffer-example

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -0,0 +1,176 @@
...
+    // Store adc and gpio tables.
+    id_to_gpio = gpio_conf;
+    id_to_adc  = adc_conf;
+
+    // Init usb peripheral.
+    hw_usb_init(transmit_handle);
+
+    // If usb is not connected, continue.
+    if (!hw_usb_checkConnection())
+        return;
```
Lucien950, line 138: idk how usb work, but can you init with it disconnected? if so, can you init in task_init and poll for connction, and run when ready? I would prefer a system where they can plug in after the board is powered up

Lucien950 (changes requested):

firmware/chimera_v2/pyproject.toml
liam-ilan: Will do! Going to try to install this without cloning the repo today 👍

firmware/shared/src/io/io_chimera_v2.c
liam-ilan: Made some changes...
1) The memory only gets allocated if chimera is enabled.
2) The memory is now larger (65 kb)... We support very large rpc messages, it is way overkill... but kind of neat that there virtually is no bound on packet size! I want to keep this - I think we might want to do more with USB down the line

firmware/shared/src/io/io_chimera_v2.c
liam-ilan: Yep - pointer to an array.

firmware/chimera_v2/proto_autogen/f4dev_pb2.py
liam-ilan: I want these in the repo. But yes they are autogenned. For now, don't want to touch the build system, there's a scripts directory in the chimera_v2 directory, that has scripts for generating docs and protobuf scripts.

firmware/chimera_v2/proto/CMakeLists.txt
liam-ilan: Would rather not... had problems with this

firmware/chimera_v2/generate_proto.sh
liam-ilan: Going to leave these scripts out of cmake for the time being... Chimera V2 is very well isolated from the build system, and I want to keep it that way for now. Also asking that we don't write powershell scripts for this for the time being, if a raw Windows user wants to run these scripts I'll add support later.

TLDR; Keeping things simple for the time being.

firmware/chimera_v2/chimera_v2.py
liam-ilan: Thanks! Made some changes to make it even better - we now have autogenned docs!

firmware/chimera_v2/pyproject.toml
liam-ilan: @gtaharaedmonds I added all of these to requirements.txt... the structure runs like this:

- All dependencies are in requirements.txt in the top-level directory
- Dependencies needed to run chimera are in the pyproject.toml

There's a better way to structure this package, (See https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), but our package is so simple that for now we don't need it.

firmware/shared/src/hw/hw_usb.c
liam-ilan: Done!

firmware/shared/src/hw/hw_usb.c
liam-ilan: Yeah agreed - made some significant improvements, should be substantially better now.

Both the transmit and receive functions now return bools. Receive blocks until the message is received. This allows us to drop the start byte from the message format.

firmware/shared/src/hw/hw_usb.c
```diff
@@ -34,23 +34,30 @@ bool hw_usb_checkConnection()
 
 void hw_usb_transmit(uint8_t *msg, uint16_t len)
 {
-    CDC_Transmit_FS(msg, len);
+    uint8_t handle_status = usb_transmit_handle(msg, len);
+    if (handle_status != 0)
+        LOG_WARN("Chimera: USB handle returned %d status code instead of 0", handle_status);
```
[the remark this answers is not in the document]
liam-ilan: On the H7s, it's `CDC_Transmit_HS` - H7Dev has been busted for a while, so I haven't had the chance to test this : (

firmware/dev/f4dev/src/cubemx/Src/main.c
liam-ilan: I think `MX_USB_DEVICE_Init` is autogenerated, would love to put in hw_usb but I don't think I can...

liam-ilan: Alternative to passing the handler is to set some compiler flag for H7 and F4, and then ifdef on that, let me know if you think that would be better.

firmware/shared/src/hw/hw_usb.c
liam-ilan: Took a look - will shoot a shot at this - want to kick it to a later PR.

firmware/dev/f4dev/src/cubemx/Src/main.c
liam-ilan: Removed the dependency-injected handlers. in favour of
```c
// Select between FS or HS methods.
// F4s (CM4s) only support FS (Full Speed), while H7s (CM7s) support HS (High Speed).
// This macro exposes the USB_DEVICE_HANDLER macro (hUsbDeviceFS or hUsbDeviceHS),
// and the TRANSMIT macro (CDC_Transmit_FS or CDC_Transmit_HS).
// Everything is derived from usbd_cdc_if.h.
#if defined(CM4_TARGET)
#define TRANSMIT(buf, len) (CDC_Transmit_FS(buf, len))
extern USBD_HandleTypeDef hUsbDeviceFS;
#define USB_DEVICE_HANDLER (hUsbDeviceFS)
#elif defined(CM7_TARGET)
#define TRANSMIT(buf, len) (CDC_Transmit_HS(buf, len))
extern USBD_HandleTypeDef hUsbDeviceHS;
#define USB_DEVICE_HANDLER (hUsbDeviceHS)
#else
#error Either CM7_TARGET or CM4_TARGET must be defined, \
        so that hw_usb knows which handlers to use.
#endif
```

gtaharaedmonds: lets get this money

gtaharaedmonds (approved):
looks great. docs are awesome

PR #1425, opened 2025-01-18
VC power monitoring

kryptoish:
### Changelist 

- added power monitoring driver code 
- power monitor init, current, voltage (bus) readings, and alert status 

### Testing Done

- not yet

### Resolved Tickets

- [FIRM-151](https://ubcformulaelectric.atlassian.net/browse/FIRM-220)

[FIRM-151]: https://ubcformulaelectric.atlassian.net/browse/FIRM-151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


firmware/quintuna/VC/src/io/io_power_monitor.c
```diff
@@ -0,0 +1,67 @@
...
+#include "hw_i2cs.h"
+
+#define VOLTAGE_BUS_TRANSFER_FACTOR 4.8828e-4f;
+#define VOLTAGE_SENSE_TRANSFER_FACTOR 1.5259e-6f;
+#define SENSE_RESISTOR 2.0e6f;
+#define FSC 5.00e-8f // Full-Scale Current
+#define REFRESH 0x1F
+#define ALERT_STATUS 0x26
+
+const PowerRail boost = {.address = 0x08}; // VBUS2
```
AmirTajaddodi, line 11: minor changes here: Take a look at altium again and make sure we are reading from the right register addresses 
image

firmware/quintuna/VC/src/io/io_power_monitor.c
```diff
@@ -0,0 +1,67 @@
...
+
+    buffer[0] = REFRESH;
+    hw_i2c_transmit(&pwr_mon, buffer, 1); // refresh
+
+    buffer[0] = voltage_address->address;
+    hw_i2c_transmit(&pwr_mon, buffer, 1); // change register pointer address
+
+    hw_i2c_receive(&pwr_mon, buffer, 1); // get raw register data and store it in buffer
+
+    const float voltage_buffer = (float)buffer[0] * VOLTAGE_BUS_TRANSFER_FACTOR;
```
AmirTajaddodi, line 32: for the bus voltages I believe we are getting 2 bytes of information so make sure to account for that

firmware/quintuna/VC/src/io/io_power_monitor.c
```diff
@@ -0,0 +1,67 @@
...
+}
+
+float io_power_monitor_read_current(const PowerRail *current_address)
+{
+    uint8_t buffer[1];
+
+    buffer[0] = REFRESH;
+    hw_i2c_transmit(&pwr_mon, buffer, 1); // refresh
+
+    buffer[0] = (current_address->address) + 0x04;
```
AmirTajaddodi, line 44: I would say lets make things a little bit more explicit and another member to your struct talking about current vs voltage addresses for the chip

firmware/quintuna/VC/src/io/io_power_monitor.c
```diff
@@ -0,0 +1,67 @@
...
+    hw_i2c_transmit(&pwr_mon, buffer, 1); // change register pointer address (Vsense)
+
+    hw_i2c_receive(&pwr_mon, buffer, 1); // get raw register data and store it in buffer
+
+    const float voltage_sense_buffer = (float)buffer[0] * VOLTAGE_SENSE_TRANSFER_FACTOR;
```
AmirTajaddodi, line 49: again make sure to account for the fact it 2 bytes of info

firmware/quintuna/VC/src/io/io_power_monitor.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef TARGET_EMBEDDED
```
AmirTajaddodi, line 5: Remind to tell you why we do this #ifdef TARGET_EMBEDDED action during our next meeting on saturday or when ever you can

firmware/quintuna/VC/src/io/io_power_monitor.h
```diff
@@ -0,0 +1,23 @@
...
+typedef struct
+{
+    const uint8_t address;
```
AmirTajaddodi, line 8: yeah as mentioned before restructure your struct to hold both current and voltage address

AmirTajaddodi (changes requested):
This is a really good. Some final touches and fixes and this should be good to go!!

firmware/quintuna/VC/src/io/io_power_monitor.h
```diff
@@ -0,0 +1,23 @@
...
+#endif
+
+extern const PowerRail bat;
+extern const PowerRail boost;
+extern const PowerRail acc;
+extern const PowerRail ext;
+
+bool io_power_monitor_init();
+
+float io_power_monitor_read_voltage(const PowerRail *voltage_address);
```
Lucien950, line 23: make sure all of your functions are exposed in your header file, otherwise other modules cannot use them.

firmware/quintuna/VC/src/io/io_power_monitor.c
Lucien950, line 1: are you configuring the control register (7-2)? in particular have you investigated burst and fast mode?

Lucien950, line 1: also power calculations?

firmware/quintuna/VC/src/io/io_power_monitor.c
```diff
@@ -0,0 +1,67 @@
...
+{
+    return hw_i2c_isTargetReady(&pwr_mon);
+}
+
+float io_power_monitor_read_voltage(const PowerRail *voltage_address)
+{
+    uint8_t buffer[2];
+
+    buffer[0] = REFRESH;
+    hw_i2c_transmit(&pwr_mon, buffer, 1); // refresh
```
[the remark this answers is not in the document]
Lucien950: cause I would imagine you might want a system which operates off GPIO interrupts letting you know when conversions are done?

Lucien950 (changes requested):

firmware/quintuna/VC/CMakeLists.txt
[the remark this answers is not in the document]
kryptoish: what?

firmware/quintuna/VC/src/io/io_power_monitor.h
```diff
@@ -0,0 +1,25 @@
...
+
+extern const PowerRail boost;
+extern const PowerRail acc;
+extern const PowerRail ext;
+
+bool io_power_monitor_init();
+
+float io_power_monitor_read_voltage(const PowerRail *voltage_address);
+
+uint32_t io_power_monitor_alert_status();
```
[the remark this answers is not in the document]
kryptoish: This function (alert_status) is for checking which alert exactly was tripped from the power monitor. This would be used after a fault I suppose.

firmware/quintuna/VC/src/io/io_power_monitor.c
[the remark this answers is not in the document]
kryptoish: I separated into a different function and will look into any timing concerns.

firmware/quintuna/VC/src/io/io_power_monitor.c
kryptoish: i mean i can make a function for that quickly

firmware/quintuna/VC/src/io/io_power_monitor.c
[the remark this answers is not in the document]
kryptoish: btw isnt that what the timeout duration used for in this case?

firmware/quintuna/VC/src/io/io_power_monitor.c
kryptoish: Will probably make smth for that where we use fast mode as it looks most applicable, but it looks like they are just used to set sampling speeds and to choose which channels are active.

PR #1428, opened 2025-01-19
Move Self Heartbeat Broadcaster

Lucien950:
### Changelist 

Note that the self heartbeat just needs to be set once in the CAN table, and then it will be repeatedly broadcast.


firmware/quadruna/FSM/src/tasks.c
```diff
@@ -101,7 +101,7 @@ void tasks_init(void)
     // Configure and initialize SEGGER SystemView.
     // NOTE: Needs to be done after clock config!
     SEGGER_SYSVIEW_Conf();
-    LOG_INFO("VC reset!");
+    LOG_INFO("FSM reset!");
```
AmirTajaddodi, line 104: Can you make this change for all the boards. Currently all boards are saying VC reset in task init instead of their own name

AmirTajaddodi (changes requested):

firmware/quadruna/FSM/src/tasks.c
Lucien950: damn

AmirTajaddodi (approved):
LGTM!!

Lucien950:
Big if true

PR #1432, opened 2025-01-22
Cortex-M7 compiler flag fix

gtaharaedmonds:
### Changelist 

Looks like I made a mistake when setting up the compiler flags for Cortex-M7s and we've been passing `-mcpu=cortex-m4`. They both support the same instruction set (which I believe this flag is responsible for configuring) which is probably why we haven't seen any issues. However, we should definitely correct this regardless. There are differences in the final binaries so this definitely does something...

### Testing Done

- [x] Builds pass.
- [x] Sanity test on hardware.


Lucien950 (approved):
Excellent

PR #1450, opened 2025-02-08
Switch can queue from message queue to message buffer 

Jiajun-Huang:
### Resolved Tickets

image

After 
image

Before
image


firmware/shared/src/io/io_canQueue.c
```diff
@@ -45,31 +33,40 @@ __weak void canRxQueueOverflowClearCallback() {}
...
 {
     assert(init_complete);
-    static uint32_t  tx_overflow_count = 0;
-    const osStatus_t s                 = osMessageQueuePut(tx_queue_id, tx_msg, 0, 0);
+    static uint32_t tx_overflow_count = 0;
+    size_t          bytes_sent        = xMessageBufferSend(tx_queue_id, tx_msg, sizeof(CanMsg), 0);
 
-    if (s != osOK)
-        // If pushing to the queue failed, the queue is full. Discard the tx_msg and invoke the TX overflow callback.
+    if (bytes_sent != sizeof(CanMsg))
```
Lucien950, line 52: wait is it possible to send only a part of what you ask it to send? or is it all or nothing?

firmware/shared/src/io/io_canQueue.c
```diff
@@ -45,31 +33,40 @@ __weak void canRxQueueOverflowClearCallback() {}
...
-    tx_queue_id   = osMessageQueueNew(TX_QUEUE_SIZE, sizeof(CanMsg), &tx_queue_attr);
-    rx_queue_id   = osMessageQueueNew(RX_QUEUE_SIZE, sizeof(CanMsg), &rx_queue_attr);
+    tx_queue_id   = xMessageBufferCreateStatic(TX_QUEUE_SIZE, tx_queue_buf, &tx_buffer_control_block);
+    rx_queue_id   = xMessageBufferCreateStatic(RX_QUEUE_SIZE, rx_queue_buf, &rx_buffer_control_block);
     init_complete = true;
+
+    if (tx_queue_id == NULL || rx_queue_id == NULL)
+    {
+        assert(0);
+    }
```
Lucien950, line 43:
```suggestion
    assert(tx_queue_id != NULL && rx_queue_id != NULL);
```

Lucien950: demorgan :)

Lucien950 (changes requested):
little changes, test and I will approve!

firmware/shared/src/io/io_canQueue.c
Jiajun-Huang: I think it might return 0 if buffer is full or something wrong happens. so second one.

Lucien950 (approved):
happy with it

PR #1465, opened 2025-02-17
some CAN hotfixes

Lucien950:
### Changelist 

- all quadruna boards back to 500kbits (due to charging for the BMS)
  - bootloaders remain at 1mbit
- globalize CAN interrupt handling
- interesting way to get handles from STM32 handles (better in my opinion dependency inversion)
- move io_can back to hw_can (as @liam-ilan wants to mock CAN at queue level)

### Testing Done

- tested on fsm, crit, vc


firmware/boot/bootloader.c
```diff
@@ -161,6 +162,9 @@ void bootloader_init(void)
     // other MCUs.
     bootloader_boardSpecific_init();
 
+    hw_can_init(&can);
+    io_canQueue_init();
+
```
gtaharaedmonds, line 167: If you start the HW CAN here then I think you should stop and deinit it before jumping to the app.

firmware/quadruna/FSM/src/tasks.c
```diff
@@ -240,7 +247,10 @@ _Noreturn void tasks_runCanTx(void)
     for (;;)
     {
         CanMsg tx_msg = io_canQueue_popTx();
-        io_can_transmit(&can, &tx_msg);
+        if (!hw_can_transmit(&can, &tx_msg))
+        {
+            LOG_ERROR("CAN TX failed!");
+        }
```
gtaharaedmonds, line 253: Maybe do this for all the boards? seems kinda weird to only log the error on some

firmware/quadruna/VC/src/jobs.c
```diff
@@ -59,13 +64,12 @@ void jobs_init()
...
-    if (!io_imu_init())
-    {
-        app_canAlerts_VC_Warning_ImuInitFailed_set(true);
-        LOG_INFO("Imu initialization failed");
-    }
+    // if (!io_imu_init())
+    // {
+    //     app_canAlerts_VC_Warning_ImuInitFailed_set(true);
+    //     LOG_INFO("Imu initialization failed");
+    // }
```
gtaharaedmonds, line 71: Was this since it was hard faulting?

firmware/quintuna/CRIT/src/jobs.c
```diff
@@ -0,0 +1,60 @@
...
+    // Check in for timeouts for all RTOS tasks
+    const uint32_t task_start_ms = io_time_getCurrentMs();
+    io_canTx_enqueueOtherPeriodicMsgs(task_start_ms);
+
+    // Watchdog check-in must be the last function called before putting the
+    // task to sleep. Prevent check in if the elapsed period is greater or
+    // equal to the period ms
+    if (io_time_getCurrentMs() - task_start_ms <= 1)
+    {
+    }
```
gtaharaedmonds, line 53: ?

firmware/quintuna/RSM/boot/hw_sysviewDescConfig.c
```diff
@@ -0,0 +1,7 @@
+#include "SEGGER_SYSVIEW.h"
+
+void hw_sysviewConfig_sendSystemDesc(void)
+{
+    SEGGER_SYSVIEW_SendSysDesc("N=RSM_Boot,D=Cortex-M4,O=FreeRTOS");
+    SEGGER_SYSVIEW_SendSysDesc("I#15=SysTick");
+}
```
gtaharaedmonds, line 7: Would be nice to make a standard file to F4 vs. H7 variants since the interrupt names should be the same. Not relevant here but probably a good idea

firmware/shared/src/hw/hw_can.c
```diff
@@ -87,17 +98,21 @@ bool io_can_transmit(const CanHandle *can_handle, CanMsg *msg)
     tx_header.TransmitGlobalTime = DISABLE;
 
     // Spin until a TX mailbox becomes available.
-    while (HAL_CAN_GetTxMailboxesFreeLevel(can_handle->hcan) == 0U)
+    static const uint16_t TRIES = 1000; // empirically, makes it try for about 1ms
+    for (int i = 0; i < TRIES && HAL_CAN_GetTxMailboxesFreeLevel(can_handle->hcan) == 0U; i++)
         ;
+    if (HAL_CAN_GetTxMailboxesFreeLevel(can_handle->hcan) == 0U)
+        return false;
```
gtaharaedmonds, line 105: Do we need this? If we do we should probably think harder about the timeout to avoid potentially dropping messages during high bus load

firmware/shared/src/hw/hw_fdcan.c
```diff
@@ -1,11 +1,20 @@
...
 #include "io_time.h"
+#include "io_canQueue.h"
 
-void io_can_init(const CanHandle *can_handle)
+/**
+ * @attention THIS MUST BE DEFINED IN YOUR CONFIGURATIONS
+ * @param hfdcan takes a handle to a STM32 HAL CAN object
+ * @returns a pointer to a CanHandle object (the metadata associated with the STM32 HAL CAN object)
+ */
+CanHandle *hw_can_getHandle(const FDCAN_HandleTypeDef *hfdcan);
```
gtaharaedmonds, line 13: Maybe put this in the header file so its more clear?

firmware/shared/src/hw/hw_fdcan.c
```diff
@@ -47,14 +58,18 @@ bool io_can_transmit(const CanHandle *can_handle, CanMsg *msg)
     tx_header.TxEventFifoControl  = FDCAN_NO_TX_EVENTS;
     tx_header.MessageMarker       = 0;
 
-    while (HAL_FDCAN_GetTxFifoFreeLevel(can_handle->hcan) == 0U)
+    static uint16_t TRIES = 1000;
+    for (int i = 0; i < TRIES && HAL_FDCAN_GetTxFifoFreeLevel(can_handle->hcan) == 0U; i++)
         ;
+    if (HAL_FDCAN_GetTxFifoFreeLevel(can_handle->hcan) == 0U)
+        return false;
```
gtaharaedmonds, line 65: Same comment as hw_can.c

gtaharaedmonds (changes requested):
A few minor things but this all looks good

firmware/quadruna/FSM/src/tasks.c
Lucien950: oh yes i forgor

firmware/quadruna/VC/src/jobs.c
Lucien950: yes :)

firmware/quintuna/CRIT/src/jobs.c
Lucien950: uh watchdog, we'll figure that out later

firmware/quintuna/RSM/boot/hw_sysviewDescConfig.c
Lucien950: oh that is true. I was going to do that but then you would have to configure all the interrupts in the table which is hard

firmware/shared/src/hw/hw_can.c
Lucien950: if a single message is taking more than 1ms then you are in a situation where you can only send 10 messages per 100hz task. i think you're probably cooked anyways in that case. i am happy to bump this up to like 5ms but otherwise, I think we should start clearing stale messages from the queue.

firmware/shared/src/hw/hw_fdcan.c
Lucien950: but then if you include "hw_can.h" then you will get this function? Is that desired?

firmware/shared/src/hw/hw_can.c
gtaharaedmonds: Hmm in my opinion this doesn't really solve that problem? If If our messages are regularly taking this long then yeah we're probably cooked. Either we pop them off the front due to a timeout, or the CAN queues fill up and we get a warning. The advantage of letting them fill up is we know when this happens due to the warning.

Also I think this does add a risk that we configure the TRIES var wrong then we end up with dropped messages during moments of high bus load. I honestly don't know what the maximum latency on adding a messages is, have you checked on the car or anything like that? 

Seems like this doesn't really help solve any problems (and potentially hides it if this means the warning doesn't get set), and adds a potential risk which would be tricky to debug.

firmware/shared/src/hw/hw_fdcan.c
gtaharaedmonds: uhh i suppose not, maybe just comment then?

firmware/quintuna/RSM/boot/hw_sysviewDescConfig.c
Lucien950: actually you cannot do this because then all interrupts (whether or not it is caused) will be listed in systemview which is annoying

gtaharaedmonds (approved):
yeeeep

PR #1466, opened 2025-02-18
BMS Tractive System Fault State

DJ90864:


gtaharaedmonds (changes requested):
We still want to protect the cells - this fault means we're above the max discharge current for 0.1s. We should probably change this so:

1. The timeout is 3s (recommended by datasheet)
2. The BMS sets a non-latching fault or warning so the current draw is reduced, and the cells are protected

Lucien950:
@KelmLelm close on your discretion

PR #1467, opened 2025-02-18
Change CAN queueing behaviour on VC

gtaharaedmonds:

### DIFF
--- a/firmware/quadruna/VC/src/jobs.h
+++ b/firmware/quadruna/VC/src/jobs.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "io_canMsg.h"
+
 /**
  * Anything that needs to be initialized for unit testing must be put here
  */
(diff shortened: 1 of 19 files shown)


firmware/quadruna/VC/src/jobs.h
```diff
@@ -16,3 +18,5 @@ void jobs_run1Hz_tick(void);
 void jobs_run100Hz_tick(void);
 void jobs_run1kHz_tick(void);
 void jobs_runCanRx_tick(void);
+
+void jobs_canRxCallback(const CanMsg* rx_msg);
```
gtaharaedmonds, line 22: @liam-ilan Thought you might want this for tests? If not let me know and I'll probably just move it to be static in `hw_cans.c`.

firmware/shared/src/hw/hw_can.c
Lucien950, line 1: forgor assert existance of callback function

Lucien950 (changes requested):
also the other boards

Lucien950 (approved):
ok

PR #1469, opened 2025-02-24
BMS Quintuna hw layer

DJ90864:
### Changelist 
Implemented Quintuna HW Layer (except Chimera)

### Testing Done
Not tested yet


firmware/quintuna/BMS/src/hw/hw_cans.c
```diff
@@ -0,0 +1,19 @@
...
+
+const CanHandle *hw_can_getHandle(const FDCAN_HandleTypeDef *hfdcan)
+{
+    assert(hfdcan == can1.hcan || hfdcan == can2.hcan);
+
+    if (hfdcan == can1.hcan) {
+        return &can1;
+    }
+    else {
+        return &can2;
```
Lucien950, line 17: switch case and default case assert(false)

firmware/quintuna/BMS/src/hw/hw_cans.c
```diff
@@ -0,0 +1,19 @@
+#include "hw_cans.h"
+#include "main.h"
+
+#include <assert.h>
+
+CanHandle can1 = { .hcan = &hfdcan1 };
+CanHandle can2 = { .hcan = &hfdcan2 };
```
Lucien950, line 7: how does this work again?  do we have a gpio line saying which to read from? or do we just read from both all the time? do we deinit/reinit how?

firmware/quintuna/BMS/src/hw/hw_adcs.c
```diff
@@ -0,0 +1,47 @@
...
+
+void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef *hadc)
+{
+    if (hadc == adc1.hadc) {
+        hw_adcchip_updateCallback(&adc1);
+    }
+    else if (hadc == adc3.hadc) {
+        hw_adcchip_updateCallback(&adc3);
+    }
+}
```
Lucien950, line 47: we might change this later but this is good for now.

Lucien950 (approved):
a few very very small things.

otherwise very excellent work

PR #1481, opened 2025-03-07
configured red and green tsim

Ashkanaaa:


firmware/quintuna/DAM/src/io/io_tsim.c
```diff
@@ -0,0 +1,29 @@
+#include "io_tsim.h"
+#include "hw_gpio.h"
+
+static const TsimLightConfig *config = NULL;
+
+void io_tsim_init(const TsimLightConfig *tsim_config)
+{
+    config = tsim_config;
+}
```
Lucien950, line 9: do not use the config method anymore. Please directly access the GPIO pin through includes.

Lucien950 (changes requested):
Joe made a hardware mistake with this functionality in particular. Please speak with him about the issue before we merge this.

firmware/quintuna/DAM/src/io/io_tsim.c
Jiajun-Huang: so guess no config struct as well?

Jiajun-Huang (approved):
LGTM!

PR #1483, opened 2025-03-08
Velocity Estimator EKF

Aditya-Dhiman4:
### Changelist 
Added ekf.py containing base EKF class that can be adapted for any kind of state estimation.

Added velocity_estimator.py which uses the EKF class to perform velocity estimator, it will not work 
if it is ran as it does not have a way to get car data.

### Testing Done
Added test_ekf.py containing simple test cases, will probably make it more rigorous after.


philam2001 (commented):
can u link the papers u use for this/ formulas

software/velocity_estimation_ekf/velocity_estimation.py
```diff
@@ -0,0 +1,175 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import math
+from ekf import EKF
+
+# Velocity estimator using EKF:
+# http://zotero.org/groups/5809911/vehicle_controls_2024/items/ZKJWDKCJ/reader
+
+def predict_velocity(x, u, dt):
```
philam2001, line 9: make your variables more readable

philam2001, line 9: more specific for our use case

philam2001 (changes requested):

Lucien950:
KILL!

PR #1485, opened 2025-03-09
DAM Shutdown Loop Configs

Ashkanaaa:


Jiajun-Huang (approved):

Lucien950:
Sorry, I did not communicate this to @Jiajun-Huang. We need to configure all the shutdown nodes together, because there is a specific CAN message assignments to broadcast shutdown loop nodes and determine which is the first shutdown node which is open.

PR #1486, opened 2025-03-09
Clion Configurations

Lucien950:


kryptoish (approved):
Looks fine to me. But you might wanna recheck if the Quintuna_CRIT_Debug.xml is supposed to be there because i don't see it as one of the configs on my side although you said you renamed it.

PR #1494, opened 2025-03-12
Fix SSM Blinky Chimera V2 Example

liam-ilan:
### Changelist 
- Fixes the Chimera V2 SSM blinky example
- Minor README fix.


Lucien950 (approved):
nbet

PR #1495, opened 2025-03-13
Shared IMU

shayana18:
… on car

### Changelist 
- VC imu driver made into a wrapper driver 
- shared imu driver implemented
### Testing Done
- Will flash car and check IMU response over CAN


firmware/quadruna/VC/CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ list(APPEND IO_SRCS
         "${SHARED_IO_INCLUDE_DIR}/io_led.c"
         "${SHARED_IO_INCLUDE_DIR}/io_time.c"
         "${SHARED_IO_INCLUDE_DIR}/io_taskMonitor.c"
+        "${SHARED_IO_INCLUDE_DIR}/io_imu.c"
+
```
Lucien950, line 36: extra space

firmware/quadruna/VC/src/io/io_imu_config.h
```diff
@@ -0,0 +1,5 @@
+#include "hw_i2cs.h"
+#include "hw_i2c.h"
```
Lucien950, line 2: redundant import

firmware/quadruna/VC/src/io/io_imu_config.h
```diff
@@ -0,0 +1,5 @@
...
+#include "io_imu.h"
+
+extern const I2cDevice *imu_i2c_handle;
```
Lucien950, line 5: can you add a wrapper, even if it only contain the i2c handle? this makes it possible to unit test this. (by ifdefing the i2cdevice away and having an empty struct)

Lucien950: also I think it's probably best to configure the sensitivity within the struct rather than globally?

Lucien950: idk tho cause low key might be a bad idea

Lucien950: also this makes it so that only the .c file needs to import hw files (which is what we need, only io.c can import hw files, and io.h CANNOT import HW files)

firmware/shared/src/io/io_imu.c
```diff
@@ -10,9 +8,29 @@
...
+// IMU Register Addresses
+#define XL_OUTPUT_RATE_REG 0x10  // [3:0] select output rate
+#define G_OUTPUT_RATE_REG 0x11   // [3:0] select output rate
+// High Byte Reg = Low Byte Reg + 1
+#define G_ROLL_LOW_BYTE_REG 0x22
+#define G_PITCH_LOW_BYTE_REG 0x24
+#define G_YAW_LOW_BYTE_REG 0x26
+#define XL_X_LOW_BYTE_REG 0x28
+#define XL_Y_LOW_BYTE_REG 0x2A
+#define XL_Z_LOW_BYTE_REG 0x2C
```
Lucien950, line 26: can you put these a bit closer to the usages? otherwise you need to ping pong read between the usages and the definitions

Lucien950 (changes requested):
few things to look at

firmware/quadruna/VC/src/io/io_imu_config.h
shayana18: this is a bit outdated. Take a look at the TV rework PR. I added a struct

firmware/shared/src/io/io_imu.c
shayana18: I don't really wanna move it inside functions. You can just however over the macro to get the address no?

PR #1503, opened 2025-03-15
Fix CheckConnection USB behaviour - trigger Chimera on Every Loop of 100Hz task. 

liam-ilan:
### Changelist 
- Normal USB devices have the D+ pullup tied to a GPIO for firmware control. Firmware marks that the device is ready by tying this GPIO high.
- We don't have this - so USB does not work on MCU boot.
- Instead, we check for usb on every loop, and jump into chimera_v2 if connected (hence the checkConnection fix).


Lucien950 (approved):
lets go

PR #1508, opened 2025-03-19
Fix CAN logging bugs

gtaharaedmonds:
### Changelist 

Fix a few bugs in CAN logging that were causing logging to fail silently. Also reworked some of the error handling so it should be more robust, and be more clear what's going wrong (due to asserts failing or RTT logs).

### Testing Done

- [x] Tested the bug fixes on the car but made some changes since then that I'll need to re-verify


firmware/quadruna/VC/src/jobs.c
```diff
@@ -23,19 +23,23 @@
...
     const CanMsg msg = io_jsoncan_copyToCanMsg(tx_msg);
     io_canQueue_pushTx(&msg);
-    // ReSharper disable once CppRedundantCastExpression
-    if (io_fileSystem_ready() && app_dataCapture_needsLog((uint16_t)msg.std_id, msg.timestamp))
+
+    if (app_dataCapture_needsLog((uint16_t)msg.std_id, msg.timestamp))
+    {
         io_canLogging_loggingQueuePush(&msg);
-    // ReSharper disable once CppRedundantCastExpression
+    }
```
gtaharaedmonds, line 37: note i should log errors if these fail

Jiajun-Huang (approved):
Look good. Crazy amount error checking.

PR #1510, opened 2025-03-20
Swap PWM to GPIO on CRIT

liam-ilan:
### Changelist 
- Swap PWM to GPIO on CRIT.


firmware/chimera_v2/pyproject.toml
```diff
@@ -7,7 +7,7 @@ dependencies = [
     "protobuf", "grpcio-tools", "protoletariat", 
 
     # We use libusb-package to get a version of pyusb's usb.core with libusb packaged in.
-    "libusb-package", 
+    "libusb-package", "pyusb",
```
Lucien950, line 10: mashallah

liam-ilan: @Lucien950 might not be temporary - lets merge 🙏

Lucien950:
also is there an argument to configure it so that that it's possible to use as a GPIO? since the pin/port is exposed

liam-ilan:
> also is there an argument to configure it so that that it's possible to use as a GPIO? since the pin/port is exposed

honestly not sure why it needs to be a pwm in the first place...

Lucien950:
but tbf also gpio init is not called on it? hmm...

Lucien950:
LED DIMMING is PWM because PWMing LEDs reduces their brightness

Lucien950:
https://youtu.be/qGzvMNjYE24

liam-ilan:
huh?

liam-ilan:
ohhh

liam-ilan:
still lets merge for now... @Lucien950

Lucien950 (approved):
excellent

PR #1514, opened 2025-03-20
CRIT Fix SPI Timer

liam-ilan:
### Changelist 
- CRIT Shift registers need a max of 10MHz clock - we were running them at 24, reconfigured prescaler to drop down to 3.

Tested in Hardware


SoroushRohanizadeh (approved):
good

PR #1516, opened 2025-03-22
Use Feature Flags to Remove Unused Chimera V2 Peripherals

liam-ilan:
### Changelist 
- Chimera V2 now has a centralized configuration struct.
- Fields in this struct get ifdefd away if a peripheral is not supported.
- Reconfigured all boards to this style.
- Updated docs.

### Testing Done
- [ ] Tested on a board.


firmware/shared/src/io/io_chimera_v2.h
```diff
@@ -13,19 +64,6 @@ extern bool io_chimera_v2_enabled;
...
-    pb_size_t         gpio_tag,
-    const Gpio       *id_to_gpio[],
-    pb_size_t         adc_tag,
-    const AdcChannel *id_to_adc[],
-    pb_size_t         i2c_tag,
-    const I2cDevice  *id_to_i2c[],
-    pb_size_t         spi_tag,
-    const SpiDevice  *id_to_spi[]);
\ No newline at end of file
+void io_chimera_v2_mainOrContinue(io_chimera_v2_Config *config);
```
Lucien950, line 69: just pass by value like we are not saving anything by pass by value

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -1,5 +1,6 @@
 #include <stdlib.h>
-#include <assert.h>
+#include "main.h"
```
Lucien950, line 2: already includede?

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -1,5 +1,6 @@
 #include <stdlib.h>
```
Lucien950, line 1: why?

firmware/shared/src/io/io_chimera_v2.c
```diff
@@ -8,130 +9,122 @@
...
+    if (chimera_v2_config->gpio_net_name_tag != net_name->which_name)
     {
-        LOG_ERROR("Chimera: Expected GPIO net name with tag %d, got %d", gpio_net_name_tag, net_name->which_name);
+        LOG_ERROR(
+            "Chimera: Expected GPIO net name with tag %d, got %d", chimera_v2_config->gpio_net_name_tag,
+            net_name->which_name);
         return NULL;
     }
 
     if (net_name->which_name == GpioNetName_f4dev_net_name_tag)
```
Lucien950, line 36: can you switch case this? There are compiler optimizations + it reads better i think

Lucien950 (changes requested):
looks pretty good, makes a lot of changes that I neglected to make.

a few things to look at. also btw how much of this did you copy from me? I was thorough in removing extra imports, cmake pruning, etc. idk how much of that did you copy over? I can check again if required.

firmware/shared/src/io/io_chimera_v2.h
liam-ilan: That change would not be consistent with other libraries - Would like to keep as pass-by-reference.

firmware/shared/src/io/io_chimera_v2.c
liam-ilan: I like to explicitly include everything used in a library so that implicit includes don't dissapear - I'm cool with using a different practice, let me know.

firmware/shared/src/io/io_chimera_v2.c
liam-ilan: Good catch - dropped 👍

firmware/shared/src/io/io_chimera_v2.c
liam-ilan: This is a style thing, old chimera used to switch, but switch statements in C are leaky, and this is a place prone dropping a break accidentally, not testing all the commands, and breaking chimera for a user months down the line.

TLDR; Would rather avoid switching here, and in general unless the leaky behavior is desired...

firmware/shared/src/io/io_chimera_v2.h
Lucien950: wdym consistent with other libraries? I don't think there is a comperable library here, in particular since the way we do configs is so significantly diffeent.

i think the pattern where you pass in a config pointer then store it statically does not make sense in general, and especially does not make sense here, because you do not need to persist state.

liam-ilan: Wait a second... you're right - no state to persist. Dropped the static pointer.

I'm referring to is `io_apps`, I think in general the practice should be to pass a pointer to a struct, instead of the struct itself, even if the performance benefits aren't relevant.

Lucien950 (approved):
image

PR #1517, opened 2025-03-22
Bms quintuna io layer jack into main io

yawollopkcaj:


firmware/quintuna/BMS/src/io/io_bmsShdn.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <stdbool.h>
+
+#define BMS_SHDN_NODE_COUNT 2
+
+bool io_bmsShdn_TS_ILCK_OK_get(void);
```
KelmLelm, line 7: I think you carried over the functions from quadruna

Lucien950 (changes requested):
few things to look at

firmware/quintuna/BMS/src/io/io_bspdTest.c
```diff
@@ -0,0 +1,17 @@
+#include "io_bspdTest.h"
+#include "hw_gpios.h"
+
+void io_bspdTest_enable(const bool enable)
```
KelmLelm, line 4: Not the clearest naming convention, might want to change to io_bspdTest_enable(void) and io_bspdTest_disable(void) or io_bspdTest_set(bool) (or something to that effect). I'm more in favour of enable and disable

firmware/quintuna/BMS/src/io/io_bspdTest.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <stdbool.h>
+
+void io_bspdTest_enable(bool enable);
+bool io_bspdTest_isCurrentThresholdExceeded(void);
```
KelmLelm, line 6: Last thing you could do is add a description for the functions that negate the "n_signal" to make it clear which return status means what.

mattiaswong6 (approved):

PR #1521, opened 2025-03-22
Quintuna F4 Bootloader

Jugveer-Sandher:
### Changelist 
Bootloader compatible with quintuna F4 boards and yeeted quadruna bootloader configs from `canup`.


Lucien950 (approved):
WAOW

PR #1529, opened 2025-03-23
Shared Load Switch driver

AmirTajaddodi:
### Changelist 

Made a quick change to load switch drivers


firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,44 @@
...
+        return 0;
+    }
+    return hw_adc_getVoltage(current_sense) * ADC_VOLTAGE_TO_CURRENT_A;
+}
+
+void io_STloadswitch_Reset(const ST_LoadSwitch *loadswitch)
+{
+    hw_gpio_writePin(loadswitch->stby_reset_gpio, true);
+    hw_gpio_writePin(loadswitch->stby_reset_gpio, true);
+    hw_gpio_writePin(loadswitch->stby_reset_gpio, false);
```
Lucien950, line 33: word

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void io_TILoadswitch_Reset(const TI_LoadSwitch *loadSwitch)
+{
+    if (!hw_gpio_readPin(loadSwitch->pgood))
```
Lucien950, line 38: check here?

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,44 @@
...
+
+bool io_loadswitch_isChannelEnabled(const Efuse *channel)
+{
+    return hw_gpio_readPin(channel->enable_gpio);
+}
+
+float io_loadswitch_getChannelCurrent(const Efuse *channel)
+{
+    const AdcChannel *current_sense = channel->sns_adc_channel;
+    if (current_sense == NULL)
```
Lucien950, line 22: surely you assert that it exists? otherwise silent fail

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,44 @@
+#include "io_loadswitch.h"
+#include "hw_gpio.h"
+
+#define ADC_VOLTAGE_TO_CURRENT_A 1.720f
+
+void io_loadswitch_setChannel(const Efuse *channel, const bool enabled)
+{
+    if (channel->enable_gpio != NULL)
```
Lucien950, line 8: no silent fails

Lucien950 (changes requested):
problem!

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: what do you mean

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: Yeah this should be a asset coz if current_sense =  null thats kind of a skill issue

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: whoopsies its supposed to go false true false to create a rising edge

firmware/shared/src/io/io_loadswitch.h
```diff
@@ -0,0 +1,61 @@
...
+} ST_LoadSwitch;
+
+typedef struct
+{
+    const Efuse *efuse;
+    const Gpio  *pgood;
+} TI_LoadSwitch;
+#else
+#include "app_utils.h"
+EMPTY_STRUCT(LoadSwitch)
```
Lucien950, line 29: i would imagine you need to mock both of them...?

firmware/shared/src/io/io_loadswitch.h
```diff
@@ -0,0 +1,61 @@
...
+    const Efuse *efuse1;
+    const Efuse *efuse2;
+    const Gpio  *stby_reset_gpio;
+} ST_LoadSwitch;
+
+typedef struct
+{
+    const Efuse *efuse;
+    const Gpio  *pgood;
+} TI_LoadSwitch;
```
Lucien950, line 26: i imagine the relationship is backwards? like the common use case would be if I had a efuse, that I wanted to reset the load switch associated with it?

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,41 @@
+#include "io_loadswitch.h"
+#include "hw_gpio.h"
+#include <stdbool.h>
+#include <assert.h>
+
+#define ADC_VOLTAGE_TO_CURRENT_A 1.720f
+
+void io_loadswitch_setChannel(const Efuse *channel, const bool enabled)
+{
+    assert(channel->enable_gpio == NULL);
```
Lucien950, line 10: i beg your pardon

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,41 @@
+#include "io_loadswitch.h"
+#include "hw_gpio.h"
```
Lucien950, line 2: you don't need to double include it

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,41 @@
...
+
+void io_loadswitch_setChannel(const Efuse *channel, const bool enabled)
+{
+    assert(channel->enable_gpio == NULL);
+    hw_gpio_writePin(channel->enable_gpio, enabled);
+}
+
+bool io_loadswitch_isChannelEnabled(const Efuse *channel)
+{
+    return hw_gpio_readPin(channel->enable_gpio);
```
Lucien950, line 16: why no check if channel enable gpio exists here?

Lucien950: also please check if channel is non null as well

firmware/shared/src/io/io_loadswitch.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+float io_loadswitch_getChannelCurrent(const Efuse *channel)
+{
+    const AdcChannel *current_sense = channel->sns_adc_channel;
+    assert(current_sense == NULL);
```
Lucien950, line 22: assert fails on false... not on true....

firmware/shared/src/io/io_loadswitch.c
Lucien950: i am asking why you check if pgood is good here, it is not consistent with the ST load switch and is insulting to the intelligence of the user of the reset function

Lucien950 (changes requested):
ts pmo 🥀

firmware/shared/src/io/io_loadswitch.c
Lucien950: are you sure this generates enough of an edge? please TODO a hardware test here

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: fair point ill change this

firmware/shared/src/io/io_loadswitch.h
AmirTajaddodi: Fixed

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: fixed whoopsies

firmware/shared/src/io/io_loadswitch.h
AmirTajaddodi: I thought of the loadswitch as being in entirety the chip and each efuse is the channel associated with that loadswitch

firmware/shared/src/io/io_loadswitch.c
AmirTajaddodi: yessir

firmware/shared/src/io/io_loadswitch.c
Lucien950: can you assert non null here? as you do in the write pin spot

Lucien950 (approved):
just the one last thing and we're good

PR #1531, opened 2025-03-25
Elcon communication CAN messages

yawollopkcaj:
### Changelist 
Added Elcon charging over CAN

Note that the voltage and current values follow big-endian format

[Elcon CAN Specification.pdf](https://github.com/user-attachments/files/19440806/Elcon.CAN.Specification.pdf)


Lucien950:
looks generally good, can you put the CAN specification in confluence as well?

Lucien950 (changes requested):

Lucien950 (approved):

PR #1532, opened 2025-03-25
Improved stack monitor

gtaharaedmonds:
### Changelist 

Reworked our stack monitor driver. What happens now is when FreeRTOS detects a stack overflow in one of the thread stacks, it calls `vApplicationStackOverflowHook`. This writes to a "context" variable in the boot request RAM which indicates a stack overflow, and then resets the MCU. Then, on the next boot cycle, it reads this context variable and populates the CAN table (as well as the task number which caused the overflow).

### Testing Done

Tested on the CRIT.


firmware/quadruna/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(SHARED_IO_INCLUDE_QUADRUNA_DIR "${SHARED_EMBEDDED_QUADRUNA_DIR}/io")
 set(SHARED_HW_INCLUDE_QUADRUNA_DIR "${SHARED_EMBEDDED_QUADRUNA_DIR}/hw")
 
 add_subdirectory(VC)
-add_subdirectory(BMS)
-add_subdirectory(FSM)
-add_subdirectory(RSM)
-add_subdirectory(CRIT)
```
Lucien950, line 13: ?

firmware/shared/src/hw/hw_stackMonitor.c
```diff
@@ -0,0 +1,32 @@
...
+    {
+        const float high_watermark = stackHighWatermarkPercent(task);
+
+        if (high_watermark > STACK_HIGH_CALLBACK_THRESHOLD)
+        {
+            assert(stack_monitor_config.watermark_exceeded_callback);
+            stack_monitor_config.watermark_exceeded_callback(stack_monitor_config.tasks[task].id);
+        }
+    }
+}
```
Lucien950, line 32: are these to be used anymore?

Lucien950 (changes requested):
clean up and we are gtg

firmware/shared/src/hw/hw_stackMonitor.c
gtaharaedmonds: these are old my b

firmware/quadruna/CMakeLists.txt
gtaharaedmonds: these are old my b

AmirTajaddodi (approved):

PR #1538, opened 2025-03-27
Reset reason driver

gtaharaedmonds:
### Changelist 

Little driver which reads a few registers that determine the reason the MCU reset i.e. watchdog, reset pin, brown out, etc. 
We should broadcast this over CAN with appropriate warnings which I'll do in a follow-up when CAN is set up for Quintuna.

### Testing Done

- [x] Testing on F4 dev


Lucien950 (approved):
my 14th reason 🥀🥀🥀

PR #1549, opened 2025-04-01
DAC Driver

Lucien950:


liam-ilan (commented):
Going to look through the STM32CubeMX Changes here, probably copy them over to my branch, and try them out - will update on status. DAC driver looks pretty close to what I have on the HIL - ideally want to keep all SSM logic on Python side

Huge thanks 🙏

Lucien950:
i think there is a vision here where you can make chimera super general

Lucien950:
basically make chimera a framework to handle the dispatch arbitray commands (and not necessraily tie it to just gpio/adc/i2c/spi/etc.etc.)

Lucien950:
I just don't really know what the API would look like, cause on web it's all strings (which is not very efficient?), but on the other hand protobufs makes it so that the python is super binded to the firmware

but on the other hand I suppose the python should be super binded to the firmware

firmware/dev/SSM/src/cubemx/Src/main.c
```diff
@@ -350,11 +366,14 @@ void StartDefaultTask(void *argument)
...
     MX_USB_DEVICE_Init();
     /* USER CODE BEGIN 5 */
+    hw_gpio_writePin(&dac_chipSelect, true);
+    assert(io_dac_writeReg(DACChannel_A, 1.69));
 
     // Invoke Chimera V2.
     for (;;)
     {
         hw_chimera_v2_mainOrContinue(&chimera_v2_config);
+        hw_gpio_togglePin(&debug_led);
```
Lucien950, line 376:
```suggestion
```

firmware/dev/SSM/src/io/io_dac.c
```diff
@@ -0,0 +1,63 @@
...
+        // byte 1
+        uint8_t address : 4;
+        uint8_t cmd : 4;
+
+        // byte 2
+        uint8_t data_1;
+
+        // byte 3
+        uint8_t data_2;
+    } tx_msg = { .cmd = 0x3, .address = reg_addr, .data_1 = 0xff & data >> 4, .data_2 = 0xf & data };
```
Lucien950, line 28: note that this commands both turns on the DAC and writes to it. there might be value in separating it? but I don't see it

liam-ilan:
Closing this PR, thanks for the help!

PR #1550, opened 2025-04-02
chimera fixes for vc

AmirTajaddodi:
### Changelist 

\
Chimera Fixes for VC for testing @liam-ilan


liam-ilan (approved):
LGTM - heads up, we now increment the Chimera V2 package version with every python change. I incremented it already from 0.2 to 0.3 so you're good.

PR #1559, opened 2025-04-06
BMS Hotfix from Chimera

Lucien950:
### Changelist 

- SPI to interrupt
- SPI direction change
- GPIOs to open drain


KelmLelm (approved):
send it

PR #1560, opened 2025-04-06
Chunk Outgoing USB Packets to 64 Bytes

liam-ilan:
### Changelist 
USB has two BULK endpoints (IN/OUT), each one can pass a maximum of 64 bytes at a time. We already handled reads from Chimera with the chunking logic, this PR also chunks outgoing requests. Fixes a bug in the HIL.


Lucien950 (approved):
image

PR #1565, opened 2025-04-08
LTC6813 Quintuna Port

mattiaswong6:
### Testing Done
None yet.


firmware/quintuna/BMS/src/cubemx/Src/main.c
```diff
@@ -1101,6 +1117,7 @@ void RunTask100Hz(void *argument)
     /* init code for USB_DEVICE */
     MX_USB_DEVICE_Init();
     /* USER CODE BEGIN 5 */
+    tasks_runLtc();
```
Lucien950, line 1120:
```suggestion
```

firmware/quintuna/BMS/src/jobs.c
```diff
@@ -1,6 +1,28 @@
...
+
+void jobs_runLtc(void)
+{
+    float        cell_voltages[NUM_SEGMENTS][CELLS_PER_SEGMENT];
+    static float cell_temps[NUM_SEGMENTS][THERMISTORS_PER_SEGMENT];
+
+    static bool voltage_read_success[NUM_SEGMENTS][VOLTAGE_REGISTER_GROUPS] = { false };
+    static bool temp_read_success[NUM_SEGMENTS][AUX_REGISTER_GROUPS]        = { false };
+
+    bool balance_config[NUM_SEGMENTS][CELLS_PER_SEGMENT] = { false };
```
Lucien950, line 19: make all static for consistency

mattiaswong6:
@Lucien950 do you know why the Test Firmware BMS check is not passing despite building without errors locally? It's saying that the io_ltc functions called in jobs.c are undefined, is this a linker problem?

Lucien950:
Locally it's building fine because the IO fakes are cached on your computer. If you cmake clean it will fail again

firmware/CMakeLists.txt
```diff
@@ -39,6 +39,6 @@ IF ("${TARGET}" STREQUAL "binary")
     add_subdirectory(chimera_v2)
 ENDIF ()
 add_subdirectory(logfs)
-add_subdirectory(quadruna)
+#add_subdirectory(quadruna)
```
Lucien950, line 42:
```suggestion
add_subdirectory(quadruna)
```

KelmLelm (commented):
Just had one question about the conditions of running owc and also noted one file that was probably changed by mistake. Mostly just glanced over the files I know nothing about like the self-tests, internals and statuses. Everything else looked good to me. One thing I will say is that there will be massive blocks of setter functions, etc. when we get more segments.

firmware/quintuna/BMS/src/jobs.c
```diff
@@ -1,13 +1,29 @@
...
+
+void jobs_runLtc_tick(void)
+{
+    app_segments_configSync();
+    app_segments_broadcastCellVoltages();
+    app_segments_broadcastTempsVRef();
+    app_segments_broadcastStatus();
+    if (app_canRx_Debug_EnableDebugMode_get())
+    {
+        app_segments_openWireCheck();   // cell test
```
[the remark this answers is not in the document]
Lucien950: should we? when i wrote the code I was thinking about individual cell owc but i think you right? Idk if it should fault tho?

firmware/quintuna/FSM/src/jobs.h
```diff
@@ -10,4 +10,4 @@ void jobs_init(void);
  */
 void jobs_run1Hz_tick(void);
 void jobs_run100Hz_tick(void);
-void jobs_run1kHz_tick(void);
+void jobs_run1kHz_tick(void);
```
Lucien950, line 13:
```suggestion
void jobs_run1kHz_tick(void);

```

firmware/quintuna/BMS/src/jobs.c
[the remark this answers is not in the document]
KelmLelm: I believe since in the rules it says 

EV.7.3.4 The BMS must monitor for:

  d. Missing or interrupted voltage or temperature measurements

EV.7.3.5 If the BMS detects one or more of the conditions of EV.7.3.4 above, the BMS must:
a. Open the Shutdown Circuit EV.7.2.2
b. Turn on the BMS Indicator Light and the Tractive System Status Indicator EV.5.11.5
The two lights must stay on until the BMS is manually reset EV.7.2.3

and owc telling us we have an open wire would mean we would have to fault/open the shutdown loop if I'm not mistaken. I don't know if there's any flexibility with it though to make it a warning

firmware/quintuna/BMS/src/app/app_segments.c
```diff
@@ -0,0 +1,439 @@
...
+#include <hw_utils.h>
+#include <string.h>
+
+// TODO extended balancing
+
+static ADCSpeed s = ADCSpeed_7kHz;
+
+static uint16_t segment_vref[NUM_SEGMENTS];
+
+#define CONVERT_100UV_TO_VOLTAGE(v_100uv) ((float)v_100uv * 1E-4f)
```
KelmLelm, line 17: Is there a better way to access float values for cell voltages (for example to get the min/max cell voltage in app_accumulator?) Should we just be access the BMS tx table directly for that?

gtaharaedmonds (approved):
tun tun tun sahur

PR #1566, opened 2025-04-08
Exit codes

gtaharaedmonds:
### Changelist 

Nice exit code support for common reasons a function might fail. Also some utilities for early returning or logging the error variants. 

### Testing Done

TODO


Lucien950 (approved):
I AGREE!

gtaharaedmonds:
> This is excellent - needs to go into software onboarding.
> 
> One note - I think `hw_usb` could use the exit-code-ification that `spi`, `uart`, and `i2c` got - won't block the PR on that though.

yee i agree. think i leave for follow-up though

PR #1567, opened 2025-04-22
Undefined Behaviour Sanitizer

Lucien950:


firmware/quadruna/BMS/CMakeLists.txt
```diff
@@ -51,12 +51,15 @@ list(APPEND HW_SRCS
         "${SHARED_HW_INCLUDE_DIR}/hw_sd.c"
         "${SHARED_HW_INCLUDE_DIR}/hw_crc.c"
         "${SHARED_HW_INCLUDE_DIR}/hw_fdcan.c"
+        "${SHARED_HW_INCLUDE_DIR}/hw_ubsan.c"
 )
 set(HW_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/src/hw" "${CMAKE_CURRENT_SOURCE_DIR}/src" "${SHARED_HW_INCLUDE_DIR}" "${SHARED_HW_INCLUDE_QUADRUNA_DIR}")
 
 IF ("${TARGET}" STREQUAL "binary")
     set(SRCS ${APP_SRCS} ${IO_SRCS} ${HW_SRCS} ${SYSTEM_SRCS} ${CUBEMX_SRCS})
```
Lucien950, line 59: possibly the strategy might be to make SRCS = USER_SRCS + CUBEMX_SRC and USER_SRCS = APP + IO + HW + SYSTEM

firmware/cmake/bootlib.cmake
```diff
@@ -1,24 +1,28 @@
...
 function(stm32f4_boot_binary
         BOOT_NAME
         SRCS
+        CUBEMX_SRCS
         INCLUDE_DIRS
         CONFIG_DEFINE
         SYSCALLS
         IOC_PATH
 )
+    list(APPEND ${SRCS} ${CUBEMX_SRCS})
```
Lucien950, line 20: maybe better to split it, should look more carefully

firmware/cmake/stmlib.cmake
```diff
@@ -25,9 +25,16 @@ message("  🔃 Registered generate_stm32cube_code() function")
 # Generate STM32CubeMX driver code for TARGET_NAME using the given IOC_PATH in
 # the directory where this function is called from.
 function(generate_stm32cube_code
+        LIB_NAME
         TARGET_NAME
```
Lucien950, line 29: maybe we just go with one name

firmware/cmake/embedded.cmake
```diff
@@ -184,6 +184,7 @@ function(embedded_binary
 )
     message("  ➕ [embedded.cmake, embedded_binary()] Creating Embedded Target for ${BIN_NAME}")
     set(ELF_NAME "${BIN_NAME}.elf")
+    set_source_files_properties(${BIN_SRCS} PROPERTIES COMPILE_FLAGS "-fsanitize=undefined")
```
Lucien950, line 187: I don't like this but target properties will leak into all the interface libraries. I want to just target user code, which I is exclusively touched here

firmware/shared/src/hw/hw_ubsan.c
```diff
@@ -0,0 +1,357 @@
...
+static void val_to_string(char *str, const size_t size, const struct type_descriptor *type, void *value)
+{
+    if (!type_is_int(type))
+    {
+        return;
+    }
+
+    if (type_bit_width(type) == 128)
+    {
+        LOG_WARN("what the fuck 128 bit type???");
```
AmirTajaddodi, line 94: I would say we should change the long warn here lmaoo

firmware/shared/src/hw/hw_ubsan.c
```diff
@@ -0,0 +1,357 @@
...
+    {
+        // Trap in a infinite loop.
+    }
+}
+
+/**
+ *  FUCK ASS NUMBER TESTERS
+ */
+
+int test_and_set_bit(const int bit, unsigned long *addr)
```
AmirTajaddodi, line 25: for code saftey reasons should this not also be bool

firmware/shared/src/hw/hw_ubsan.c
```diff
@@ -0,0 +1,357 @@
...
+/**
+ * HANDLING POINTER BUGS
+ */
+static const char *const type_check_kinds[] = { "load of",
+                                                "store to",
+                                                "reference binding to",
+                                                "member access within",
+                                                "member call on",
+                                                "constructor call on",
+                                                "downcast of",
```
AmirTajaddodi, line 154: Is there a reason here for double downcast of call in this array

AmirTajaddodi (approved):
I mean I like the code and I dont see any issues other than small call nitpick

firmware/shared/src/hw/hw_ubsan.c
AmirTajaddodi: although I have never seen a 128 bit type 😭

firmware/shared/src/hw/hw_ubsan.c
```diff
@@ -0,0 +1,357 @@
...
+{
+    BREAK_IF_DEBUGGER_CONNECTED();
+
+    for (;;)
+    {
+        // Trap in a infinite loop.
+    }
+}
+
+/**
```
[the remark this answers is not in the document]
Lucien950: :)

firmware/shared/src/hw/hw_ubsan.c
Lucien950: which is bool? i think the bit is the bit number

firmware/shared/src/hw/hw_ubsan.h
```diff
@@ -0,0 +1,189 @@
...
+    /// representation is the integer itself if it fits into a ValueHandle, and
+    /// a pointer to the integer otherwise. TypeName contains the true width
+    /// of the type for the signed _BitInt(N) type stored after zero bit after
+    /// TypeName as 32-bit unsigned integer.
+    TK_BitInt = 0x0002,
+    /// Any other type. The value representation is unspecified.
+    type_unknown = 0xffff
+};
+
+struct type_descriptor
```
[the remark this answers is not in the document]
Lucien950: linux plagarism moment

firmware/shared/src/hw/hw_ubsan.c
Lucien950: okdie dokes

firmware/shared/src/hw/hw_ubsan.c
AmirTajaddodi: yeah but then they use that in suppress and report and the output of the function as bool but like if this is the linux implementation I aint one to judge

firmware/shared/src/hw/hw_ubsan.c
AmirTajaddodi: W

firmware/shared/src/hw/hw_ubsan.h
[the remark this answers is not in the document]
AmirTajaddodi: LMAOO

PR #1581, opened 2025-05-05
Remove CAN from Chimera V2

liam-ilan:
### Changelist 
- Removes `CanDevice` from Chimera V2.
- Removes associated files and dependencies.
- Increment Chimera V2 to `0.5`

### Testing Done
- None.


Lucien950 (approved):
word

PR #1584, opened 2025-05-07
Amir/bms can and jobs config

AmirTajaddodi:


firmware/quintuna/BMS/src/jobs.c
```diff
@@ -1,6 +1,54 @@
...
+    const CanMsg rx_msg       = io_canQueue_popRx();
+    JsonCanMsg   json_can_msg = io_jsoncan_copyFromCanMsg(&rx_msg);
+    io_canRx_updateRxTableWithMessage(&json_can_msg);
+}
+
+void jobs_canRxCallback(const CanMsg *rx_msg)
+{
+    if (io_canRx_filterMessageId(rx_msg->std_id))
+    {
+        io_canQueue_pushRx(rx_msg);
```
Lucien950, line 50: perhaps return from here

firmware/quintuna/BMS/src/tasks.c
```diff
@@ -34,8 +35,12 @@ void tasks_init(void)
     hw_usb_init();
     hw_adcs_chipsInit();
     hw_pwms_init();
+    hw_can_init(&can1);
+    hw_can_init(&can2);
```
Lucien950, line 39: isn't this the one where we want to init based on state machine? charging vs not charging?

Lucien950 (approved):

PR #1585, opened 2025-05-08
Bootloading updates

shayana18:
ignore pipfile change, ill fix that .... also still allot to do. Im still working on figuring out re-transmission and error handling

### DIFF
--- a/firmware/boot/bootloader.c
+++ b/firmware/boot/bootloader.c
@@ -23,6 +23,8 @@
 extern CRC_HandleTypeDef hcrc;
 extern TIM_HandleTypeDef htim6;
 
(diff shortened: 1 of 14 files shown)


firmware/boot/bootloader.c
```diff
@@ -239,7 +280,7 @@ _Noreturn void bootloader_runTickTask(void)
 
     for (;;)
     {
-        // Broadcast a message at 1Hz so we can check status over CAN.
+        LOG_INFO("IM GETTING HIT");
```
Lucien950, line 283: i'm gonna touch you

Lucien950 (changes requested):

PR #1591, opened 2025-05-13
ASYNC CANTX

Lucien950:

### DIFF
--- a/firmware/shared/src/hw/hw_can_f4.c
+++ b/firmware/shared/src/hw/hw_can_f4.c
@@ -4,6 +4,8 @@
 #include "io_time.h"
 #include "io_canQueue.h"
 
+#include <io_log.h>
+
 // The following filter IDs/masks must be used with 16-bit Filter Scale
 // (FSCx = 0) and Identifier Mask Mode (FBMx = 0). In this mode, the identifier
 // registers are associated with mask registers specifying which bits of the
@@ -52,8 +54,8 @@ void hw_can_init(CanHandle *can_handle)
     // Configure interrupt mode for CAN peripheral.
     assert(
         HAL_CAN_ActivateNotification(
-            can_handle->hcan, CAN_IT_TX_MAILBOX_EMPTY | CAN_IT_RX_FIFO0_MSG_PENDING | CAN_IT_RX_FIFO1_MSG_PENDING) ==
-        HAL_OK);
+            can_handle->hcan, CAN_IT_TX_MAILBOX_EMPTY | CAN_IT_RX_FIFO0_MSG_PENDING | CAN_IT_RX_FIFO1_MSG_PENDING |
+                                  CAN_IT_WAKEUP | CAN_IT_BUSOFF) == HAL_OK);
 
     // Start the CAN peripheral.
     assert(HAL_CAN_Start(can_handle->hcan) == HAL_OK);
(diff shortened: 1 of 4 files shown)


firmware/shared/src/hw/hw_can_h7.c
```diff
@@ -24,8 +28,9 @@ void hw_can_init(CanHandle *can_handle)
     // Configure interrupt mode for CAN peripheral.
     assert(
         HAL_FDCAN_ActivateNotification(
-            can_handle->hcan, FDCAN_IT_RX_FIFO0_NEW_MESSAGE | FDCAN_IT_RX_FIFO1_NEW_MESSAGE, 0) == HAL_OK);
-    assert(HAL_FDCAN_ActivateNotification(can_handle->hcan, FDCAN_IT_BUS_OFF, 0) == HAL_OK);
+            can_handle->hcan,
+            FDCAN_IT_RX_FIFO0_NEW_MESSAGE | FDCAN_IT_RX_FIFO1_NEW_MESSAGE | FDCAN_IT_BUS_OFF | FDCAN_IT_TX_COMPLETE,
+            1) == HAL_OK);
```
Lucien950, line 33: note that this only activates the notifcatoin on bufferindex = 1. I think there's only one tx queue so it shoudln't be a problem but...

firmware/shared/src/hw/hw_can_f4.c
```diff
@@ -138,3 +159,27 @@ void HAL_CAN_RxFifo1MsgPendingCallback(CAN_HandleTypeDef *hcan)
 {
     handle_callback(hcan);
 }
+
+static void mailbox_complete_handler(CAN_HandleTypeDef *hcan)
```
AmirTajaddodi, line 163: static inline enjoyer??

firmware/shared/src/hw/hw_can_h7.c
```diff
@@ -38,6 +43,28 @@ void hw_can_deinit(const CanHandle *can_handle)
...
+    {
+        if (poll <= 1000)
+        {
+            poll++;
+            continue;
+        }
+        assert(transmit_task == NULL);
+        assert(osKernelGetState() == taskSCHEDULER_RUNNING && !xPortIsInsideInterrupt());
+        transmit_task           = xTaskGetCurrentTaskHandle();
+        const BaseType_t status = xTaskNotifyWait(0, 0, NULL, portMAX_DELAY);
```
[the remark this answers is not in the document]
Lucien950: that is a good question that I have also asked gus

firmware/shared/src/hw/hw_can_f4.c
Lucien950: fr fr

firmware/shared/src/hw/hw_can_h7.c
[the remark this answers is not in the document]
gtaharaedmonds: I think the 1000 try timeout is unnecessary as well

gtaharaedmonds (changes requested):
Nice. I'd stick with the old API of ulTaskNotifyTake / vTaskNotifyGiveFromISR since that what we're using elsewhere, unless there's a specific reason you think xTaskNotifyWait / xTaskNotifyFromISR is better. Also don't think the 1000 poll count serves much purpose

gtaharaedmonds (approved):

PR #1600, opened 2025-05-18
Clear RX table API JSONCAN

Jiajun-Huang:


scripts/code_generation/jsoncan/src/codegen/c_generation/template/app_canRx.c.j2
```diff
@@ -78,3 +78,21 @@ void app_canRx_{{ signal.name }}_update({{ signal.datatype() }} value)
 }
 {%- endfor %}
 {%- endfor %}
+
+void app_canUtils_clear_board_rx_table(CanBoard board)
```
Lucien950, line 82: i think we should just make this an app_canRx thing cuz the prefix would be confusing if you had to figure out which module imported it?

Jiajun-Huang: That is a mistake. Let me fix it. I the function in until before then moved it here.

Lucien950 (changes requested):
part 1

scripts/code_generation/jsoncan/src/codegen/c_generation/app_can_rx_module.py
```diff
@@ -1,23 +1,45 @@
...
+
+        for board in alerts_boards:
+            rx_msgs = []
+            if isinstance(db.nodes[node].rx_msgs_names, All): # if all then talk all the tx message for that board
+                rx_msgs = db.get_board_tx_messages(board)
+            else:
+                for msg in db.get_board_tx_messages(board): # else need to filter the message that is rx by the board
+                    if msg.name in db.nodes[node].rx_msgs_names:
+                        rx_msgs.append(msg)
+            self._alert_boards_messages[board] = rx_msgs
```
Lucien950, line 27: why do we need to recalculate? this seems less trivial calculations and more structural reorganization

Lucien950 (changes requested):

Lucien950 (approved):
OK!

PR #1602, opened 2025-05-18
initial app globals implementation

DJ90864:
### Changelist 
Implemented App gloabls


Lucien950 (changes requested):
No

Please put these in the correct modules

PR #1604, opened 2025-05-18
BMS app shutdown done

DJ90864:
### Changelist 
Implemented App shdn files


firmware/quintuna/BMS/src/app/app_shdnNode.c
Lucien950, line 1: can you also just call this function in allstates before merge?

Lucien950 (changes requested):

PR #1605, opened 2025-05-19
Charging implementation

yawollopkcaj:
TODO:

        reference Formula SAE Electric rules

	use timers for precharge state

	what should max allowable charging current and voltage?

	calculate max charging current with AC to DC calculations

	find out case when plugged into wall outlet
	
	ask elec if we will always plug into 240V 40A outlet or need to make cases

	take charger efficiency into account

	figure out if the car needs time between disconnecting the charger / precharge and moving states


gtaharaedmonds (approved):
A few remaining things to do but I'm happy to merge now so we can write tests + BMS drive state + fix a few charging things off of master

PR #1608, opened 2025-05-19
JSONCAN with FDCAN

gtaharaedmonds:
## Overview

JSONCAN now supports FDCAN-sized messages (up to 64 bytes). Each node is configured as FD or not FD, and there are checks that a non-FD node can't send or receive an FD message. Also updated the H7 CAN driver so it supports sending longer messages.

## Testing

Added simple JSONCAN integration test. Haven't been able to test on hardware since Quintuna boards aren't set up yet. 

[FIRM-399](https://ubcformulaelectric.atlassian.net/browse/FIRM-399)

[FIRM-399]: https://ubcformulaelectric.atlassian.net/browse/FIRM-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/code_generation/jsoncan/src/codegen/c_generation/io_can_tx_module.py
```diff
@@ -37,4 +38,5 @@ def source_template(self):
             node_bus_names=self._node_bus_names,
             busses_for_msg=self._tx_config.get_busses_for_msg,
             messages=self._messages,
+            fd=self.fd,
```
Jiajun-Huang, line 41: I think this line should be in header_template function

scripts/code_generation/jsoncan/src/json_parsing/parse_bus.py
```diff
@@ -51,7 +61,7 @@ class _BusJson(TypedDict):
                         "modes": [str],
                         "default_mode": str,
                         "nodes": [str],
-                        Optional("FD"): bool,
+                        Optional("fd"): bool,
```
Jiajun-Huang, line 64: Do we still need this field?

Jiajun-Huang (changes requested):
Goods really good. Just some small things

scripts/code_generation/jsoncan/src/json_parsing/parse_bus.py
Lucien950: I think it is good to indicate which busses are fd such that you can
1. route with it
2. decide which nodes need fd structs

Jiajun-Huang: Do we need to worry about that when two fd nodes sending fd messages (>8 bytes) on a non-fd bus or a bus with a non-fd node?

can_bus/quintuna/bus.json
```diff
@@ -1,4 +1,30 @@
...
+    "FSM": {
+      "fd": false
+    },
+    "RSM": {
+      "fd": false
+    },
+    "VC": {
+      "fd": true
+    }
+  },
```
Lucien950, line 27: Can you automatically detect if a node is on a fd bus? busses already have the fd property

scripts/code_generation/jsoncan/src/can_database.py
```diff
@@ -288,7 +291,7 @@ class All:
     pass
 
 
-@dataclass(frozen=True)
```
Lucien950, line 291: i would prefer to keep this around, is it possible to determine if a node is required to be fd before construction?

Lucien950: it's nice to keep it around to prevent people for searching for mutations across the codebase

scripts/code_generation/jsoncan/src/codegen/dbc_generation/dbc_generation.py
```diff
@@ -34,11 +34,9 @@
...
 DBC_VALUE_TABLE_TEMPLATE = "VAL_ {id} {signal_name} {entries};\n"
 
-
-# BA_DEF_ BO_  "GenMsgCycleTime" INT {cycle_time_min} {cycle_time_max};
-# BA_DEF_ SG_  "GenSigStartValue" INT {start_value_min} {start_value_max};
-# BA_DEF_DEF_  "GenMsgCycleTime" {cycle_time_default};
-# BA_DEF_DEF_  "GenSigStartValue" {start_value_default};
+# If a signal isn't recevied by any node then DBC parsing might fail.
+# Hacky solution: Every signal receives "DEBUG"
+DBC_DEFAULT_RECEIVER = "DEBUG"
```
Lucien950, line 39: can you not call this debug as to differentiate with our debug node?

scripts/code_generation/jsoncan/test/integration/test_pack_unpack.cpp
```diff
@@ -1,3 +1,4 @@
+#include <cassert>
```
Lucien950, line 1: word

Lucien950: in a gtest environment?

scripts/code_generation/jsoncan/src/json_parsing/parse_tx.py
```diff
@@ -247,11 +248,9 @@ def _get_parsed_can_message(
             "telem_cycle_time", msg_cycle_time
         )
 
-    # Check if message ID is unique
-
     signals = []
     next_available_bit = 0
-    occupied_bits: list[Optional_t[str]] = [None] * 64
+    occupied_bits: list[Optional[str]] = [None] * max_len_bits
```
Lucien950, line 253: optional here refers to the Schema type optional not the optional python type

scripts/code_generation/jsoncan/src/json_parsing/parse_bus.py
Lucien950: if a bus is fd, we assume that all nodes on the bus are fd.

I think the responsibility of checking that a fd message is not routed on a non fd bus should be as part of the routing logic

scripts/code_generation/jsoncan/src/json_parsing/json_can_parsing.py
```diff
@@ -162,18 +159,21 @@ def _add_tx_msg(self, msg: CanMessage, tx_node_name: str) -> None:
                 f"Message '{msg.name}' transmitted by node '{tx_node_name}' is also transmitted by '{self._msgs[msg.name].tx_node_name}'"
             )
 
-        # Check if this message ID is a duplicate
+        # Check if this message ID is a duplcate
```
Lucien950, line 162: vro ??? 🥀🥀🥀

can_bus/quintuna/bus.json
Lucien950: i think a lot of poor decisions downstream stem from this

Lucien950 (changes requested):
notes

can_bus/quintuna/bus.json
Jiajun-Huang: What I see from the code is that the firmware generation and routing decision is solely based on the node.fd; whereas bus.fd field is not used at all. I agreed that it make more sense to assume that the node on fd bus will have the fd capability, and we can auto detect it.

gtaharaedmonds: changed it so that a node is determined to be FD if it is connected to an FD bus

scripts/code_generation/jsoncan/src/codegen/dbc_generation/dbc_generation.py
gtaharaedmonds: I think my idea is that these are the same - "DEBUG" is essentially someone with PCAN so it makes sense that the person doing the monitoring is receiving them all

scripts/code_generation/jsoncan/test/integration/test_pack_unpack.cpp
gtaharaedmonds: yep my bad

scripts/code_generation/jsoncan/src/json_parsing/parse_tx.py
gtaharaedmonds: real shit

scripts/code_generation/jsoncan/src/json_parsing/json_can_parsing.py
gtaharaedmonds: did i stutter

gtaharaedmonds: jk my bad lol

scripts/code_generation/jsoncan/src/json_parsing/parse_bus.py
gtaharaedmonds: changed it so that nodes being FD is determined by if it is connected to an FD bus

scripts/code_generation/jsoncan/src/json_parsing/parse_tx.py
```diff
@@ -1,5 +1,6 @@
 from math import ceil
-from typing import Dict, Tuple, Optional as Optional_t, List
+import typing
```
Lucien950, line 2:
```suggestion
```

scripts/code_generation/jsoncan/src/json_parsing/json_can_parsing.py
```diff
@@ -42,20 +42,13 @@ class JsonCanParser:
         str, bool
     ]  # _collects_data[node_name] = True if the node collects data
 
-    # internal state
-    # _node_tx_msgs: Dict[str, Set[str]]  # _tx_msgs[node_name] gives a list of all the messages it txs
-    _node_rx_msgs: Dict[
```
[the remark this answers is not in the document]
Lucien950: also possibly I am being gaslit by weird python const behaviour because i suppose the owner should be the one ensuring that the data is not mutable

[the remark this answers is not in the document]
gtaharaedmonds: i think its ok its now more convenient to make the cannodes so you can reference them for FD and etc at the start and then update them as we go

Lucien950 (approved):

PR #1609, opened 2025-05-19
Linting JSONCAN

gtaharaedmonds:
Only linting JSONCAN - so my follow-up PRs don't have confusing diffs since my editor auto-lints everything. Using [black](https://github.com/psf/black).


Lucien950 (approved):
YES!

PR #1613, opened 2025-05-20
Reset before booting

gtaharaedmonds:
### Changelist 

Currently the bootloader, when jumping to the app, de-inits everything it inits and then jumps to the app. This is a bit scuffed since then we need to manually deinit everything - this is error prone and we were seeing some problems with the clocks. 

Now, when the bootloader wants to start the app, it writes to the boot flag, resets itself, and upon the next bootup (before anything is initialized) jumps to the app. This is the same mechanism as the app jumping to the bootloader.

Also switched the bootloader to use software CRC (`app_crc`) since it needs to check the app's validity before initializing the CRC peripheral now.

### Testing Done

Tested booting to app and bootloading new binaries on the FSM.


firmware/shared/src/hw/hw_bootup.c
```diff
@@ -28,4 +44,23 @@ void hw_bootup_enableInterruptsForApp()
...
+{
+    boot_request.magic   = BOOT_MAGIC;
+    boot_request.request = request;
+}
+
+BootRequest hw_bootup_getBootRequest(void)
+{
+    if (boot_request.magic == BOOT_MAGIC)
+    {
+        return boot_request.magic;
```
Lucien950, line 59: what

Lucien950 (changes requested):
agree!

firmware/shared/src/hw/hw_bootup.c
gtaharaedmonds: hehe haha

firmware/shared/src/hw/hw_bootup.c
```diff
@@ -1,11 +1,27 @@
 #include "hw_bootup.h"
-#include <stdlib.h>
-#include "hw_hal.h"
-#include "FreeRTOS.h"
```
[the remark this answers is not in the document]
gtaharaedmonds: yep readded

Lucien950 (approved):
WAOW

firmware/cmake/bootlib.cmake
```diff
@@ -53,18 +52,19 @@ function(stm32f4_boot_binary
 
     # Add shared files.
     list(APPEND SRCS
+            "${SHARED_APP_INCLUDE_DIR}/app_crc.c"
             "${SHARED_IO_INCLUDE_DIR}/io_canQueue.c"
+            "${SHARED_HW_INCLUDE_DIR}/hw_flash.c"
             "${SHARED_IO_INCLUDE_DIR}/io_time.c"
             "${SHARED_HW_INCLUDE_DIR}/hw_flash.c"
```
Jugveer-Sandher, line 59: There's two `hw_flash.c`

Jugveer-Sandher (approved):
Beauty

firmware/cmake/bootlib.cmake
gtaharaedmonds: good catch

PR #1628, opened 2025-06-01
pcm on state done and hv state done

shayana18:

### DIFF
--- a/firmware/quintuna/VC/src/app/states/app_hvState.c
+++ b/firmware/quintuna/VC/src/app/states/app_hvState.c
@@ -3,6 +3,7 @@
 #include "app_powerManager.h"
 #include "io_loadswitches.h"
 #include <app_canTx.h>
+#include <app_canRx.h>
 #include <app_canUtils.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -27,7 +28,12 @@ static void hvStateRunOnEntry(void)
     app_powerManager_updateConfig(power_manager_state);
 }
 static void hvStateRunOnTick1Hz(void) {}
-static void hvStateRunOnTick100Hz(void) {}
+static void hvStateRunOnTick100Hz(void) {
+    if(app_canRx_FSM_BrakeActuated_get() && app_canRx_CRIT_StartSwitch_get())
+    {
+        app_stateMachine_setNextState(&drive_state);
+    }
+}
 static void hvStateRunOnExit(void) {}
 
 State hv_state = { .name              = "HV",
--- a/firmware/quintuna/VC/src/app/states/app_pcmOnState.c
+++ b/firmware/quintuna/VC/src/app/states/app_pcmOnState.c
@@ -1,6 +1,14 @@
 #include "app_states.h"
 #include "app_canTx.h"
 #include "app_powerManager.h"
+#include "app_timer.h"
+#include "app_canAlerts.h"
+#include "io_pcm.h"
+
+#define HV_READY_VOLTAGE (18.0f)
+#define PCM_MAX_VOLTAGE (30.0f)
+#define PCM_UNDER_VOLTAGE_TIMEOUT (100) //if voltage doesnt rise above 18V in this amout of time after entering pcmOnState then go into fault state
+#define PCM_TOGGLE_TIMEOUT (10) // to retry PCM wait toggle PCM where it stays off for 10ms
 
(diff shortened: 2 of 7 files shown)


firmware/quintuna/VC/src/app/states/app_pcmOnState.c
```diff
@@ -9,24 +17,80 @@ static PowerManagerConfig power_manager_state = {
...
+{
+
+    static float pcm_prev_voltage = 0.0f; 
+    static float pcm_curr_voltage = 0.0f;
+
+    switch (app_timer_updateAndGetState(pcm_under_voltage_timer))
+    {
+
+        case TIMER_STATE_RUNNING:
+            pcm_curr_voltage = (float)app_canTx_VC_ChannelOneVoltage_get();
```
AmirTajaddodi, line 51: We should explicitly state what voltage we are getting instead of saying channel one voltage and channel two voltage

firmware/quintuna/VC/src/app/states/app_pcmOnState.c
```diff
@@ -9,24 +17,80 @@ static PowerManagerConfig power_manager_state = {
...
+                io_pcm_toggle();
+            }
+            break;
+        
+        case TIMER_STATE_IDLE:
+            app_timer_restart(pcm_under_voltage_timer);
+            break;
+    }
+    
+    if (pcm_curr_voltage > PCM_MAX_VOLTAGE && pcm_prev_voltage)
```
AmirTajaddodi, line 74: One thing to note that is the power monitoring chip has an alert. Lets see if we can use that instead of us manually checking the OVL UVL conditions

firmware/quintuna/VC/src/app/states/app_hvState.c
```diff
@@ -27,7 +28,12 @@ static void hvStateRunOnEntry(void)
     app_powerManager_updateConfig(power_manager_state);
 }
 static void hvStateRunOnTick1Hz(void) {}
-static void hvStateRunOnTick100Hz(void) {}
+static void hvStateRunOnTick100Hz(void) {
+    if(app_canRx_FSM_BrakeActuated_get() && app_canRx_CRIT_StartSwitch_get())
```
AmirTajaddodi, line 32: take a look at the how quadruna implemented the logic crit start switch

firmware/quintuna/VC/src/app/states/app_pcmOnState.c
```diff
@@ -9,24 +17,80 @@ static PowerManagerConfig power_manager_state = {
                        [EFUSE_CHANNEL_R_INV]   = { .efuse_enable = true, .timeout = 0, .max_retry = 5 },
                        [EFUSE_CHANNEL_DAM]     = { .efuse_enable = true, .timeout = 0, .max_retry = 5 },
                        [EFUSE_CHANNEL_FRONT]   = { .efuse_enable = true, .timeout = 0, .max_retry = 5 },
-                       [EFUSE_CHANNEL_RL_PUMP] = { .efuse_enable = false, .timeout = 200, .max_retry = 5 },
-                       [EFUSE_CHANNEL_RR_PUMP] = { .efuse_enable = false, .timeout = 200, .max_retry = 5 },
-                       [EFUSE_CHANNEL_F_PUMP]  = { .efuse_enable = false, .timeout = 200, .max_retry = 5 },
-                       [EFUSE_CHANNEL_L_RAD]   = { .efuse_enable = false, .timeout = 200, .max_retry = 5 },
-                       [EFUSE_CHANNEL_R_RAD]   = { .efuse_enable = false, .timeout = 200, .max_retry = 5 } }
+                       [EFUSE_CHANNEL_RL_PUMP] = { .efuse_enable = true, .timeout = 200, .max_retry = 5 },
+                       [EFUSE_CHANNEL_RR_PUMP] = { .efuse_enable = true, .timeout = 200, .max_retry = 5 },
```
AmirTajaddodi, line 21: ok so one thing I am thinking is that the efuse for the pumps no longer need to be staggered because we are just doing a gradual increase in flow rate

AmirTajaddodi (changes requested):

firmware/quintuna/VC/src/app/states/app_pcmOnState.c
shayana18: so you just want to enable the efuse in init?

PR #1631, opened 2025-06-02
Auto install jsoncan

Jiajun-Huang:


Lucien950:
it is now written in rust

PR #1634, opened 2025-06-03
Quintuna VC State Machine

Lucien950:

### DIFF
--- a/can_bus/quintuna/VC/VC_rx.json
+++ b/can_bus/quintuna/VC/VC_rx.json
@@ -3,6 +3,8 @@
     "BMS_TractiveSystem",
     "BMS_Contactors",
     "BMS_Vitals",
+    "FSM_Vitals",
+    "RSM_Vitals",
     "FSM_ShdnNodeStatus",
     "DAM_ShdnNodeStatus",
     "RSM_ShdnNodeStatus",
--- a/firmware/quintuna/VC/src/app/states/app_bmsOnState.c
+++ b/firmware/quintuna/VC/src/app/states/app_bmsOnState.c
@@ -37,7 +37,9 @@ static void bmsOnStateRunOnTick100Hz(void)
     // Update: Currently we are having a state in which we tell the PCM MCU to send the turn on command and wait for hv
     // bus to go online
     const bool bms_ready_for_drive = app_canRx_BMS_State_get() == BMS_DRIVE_STATE;
-    const bool bms_in_faultstate   = app_canRx_BMS_State_get() == BMS_FAULT_STATE;
+
+    // TODO if any board is in fault state it will transition in all states. You DO NOT need to check it here
+    const bool bms_in_faultstate = app_canRx_BMS_State_get() == BMS_FAULT_STATE;
 
     if (bms_ready_for_drive)
     {
--- a/firmware/quintuna/VC/src/app/states/app_states.h
+++ b/firmware/quintuna/VC/src/app/states/app_states.h
@@ -2,12 +2,14 @@
 
 #include "app_stateMachine.h"
 
-extern State bmsOn_state;
-extern State drive_state;
-extern State hvInit_state;
-extern State init_state;
-extern State inverterOn_state;
-extern State driveWarning_state;
-extern State pcmOn_state;
-extern State hv_state;
 extern State fault_state;
+extern State init_state;
+
+extern State inverterOn_state;
+extern State bmsOn_state;
+extern State pcmOn_state;
+extern State hvInit_state;
+extern State hv_state;
+
+extern State drive_state;
+extern State driveWarning_state;
--- a/firmware/quintuna/VC/test/test_VCBase.hpp
+++ b/firmware/quintuna/VC/test/test_VCBase.hpp
@@ -6,12 +6,14 @@ extern "C"
 {
 #include "jobs.h"
 #include "io_loadswitches.h"
+#include "app_canRx.h"
 }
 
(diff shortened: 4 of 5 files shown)


shayana18:
NOTE: the idea is to reduce the number of faults due to which we exit fault state under the assumption that if something is sever enough for us to exit drive, BMS handles that. The new approach is to set torque to 0 when an unlatched fault occurs.  We also wanted to reduce the number of faults that can cause us to not be able to drive, This was reduced down to BMS latched faults and BMS heart beat, brakes were not included as brakes are fully mechanically controlled, Papps and apps were not included as when OCSC happens or disagreement a 0 torque command is sent. CRIT was not included because it doesn't really have anything important enough for us to not be able to command torque cause of (unless we want to include heart beat to cover the case that the driver turns off the switch but CRIT is cooked so driver is not able to exit drive). DAM is also not included, all it is responsible for is TSIM led

firmware/quintuna/VC/src/app/states/app_driveState.c
```diff
@@ -1,37 +1,246 @@
...
+ 
+
+    if(SWITCH_OFF == app_canRx_CRIT_StartSwitch_get())
+    {
+        app_canTx_VC_INVFRTorqueSetpoint_set(INV_OFF);
+        app_canTx_VC_INVRRTorqueSetpoint_set(INV_OFF);
+        app_canTx_VC_INVFLTorqueSetpoint_set(INV_OFF);
+        app_canTx_VC_INVRLTorqueSetpoint_set(INV_OFF);
+        app_stateMachine_setNextState(&hv_state);
+        return false;
```
gtaharaedmonds, line 163: you probably don't need to set the torques to zero here. just set them to zero when you leave drive state and that'll handle all possible cases.

firmware/quintuna/VC/src/app/states/app_driveState.c
```diff
@@ -1,37 +1,246 @@
...
+        app_canTx_VC_INVRLTorqueSetpoint_set(INV_OFF);
+    }
+
+    if(INVERTER_FAULT == warning_check)
+    {
 
-static void driveStateRunOnExit(void) {}
+        app_canAlerts_VC_Warning_InverterRetry_set(true);
+        app_stateMachine_setNextState(&hvInit_state);
+        // MAKE FUNCTION IN TORQUE DISTRIBUTION WHEN 4WD merged
```
gtaharaedmonds, line 144: you should think carefully about how multiple simultaneous conditions will be handled. for example, if there's an inverter fault and the start switch gets flipped in the same cycle, then `next_state` will be set to `hvInit_state` but then set again to `hv_state` 10 lines below. So inverters won't get retried in this case.

gtaharaedmonds: there may be other examples of bugs like this - just one I noticed rn

firmware/quintuna/VC/src/app/states/app_pcmOnState.c
```diff
@@ -9,24 +25,112 @@ static PowerManagerConfig power_manager_state = {
...
+
+            break;
+
+        case TIMER_STATE_IDLE:
+            app_timer_restart(pcm_voltage_in_range_timer);
+            break;
+    }
+
+
+    app_stateMachine_setNextState(&hvInit_state);
```
gtaharaedmonds, line 100: am i tripping or do you not actually check PCM voltage anywhere?

can_bus/quintuna/VC/VC_alerts.json
```diff
@@ -1,9 +1,80 @@
...
+      "description": "Missing heartbeat signal from the CRIT."
+    },
+    "MissingFSMHeartbeat": {
+      "id": 552,
+      "description": "Missing heartbeat signal from the FSM."
+    },
+    "MissingRSMHeartbeat": {
+      "id": 553,
+      "description": "Missing heartbeat signal from the RSM."
+    },
```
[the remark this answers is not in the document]
shayana18: Yes will update

firmware/quintuna/VC/src/app/states/app_driveState.c
```diff
@@ -1,37 +1,246 @@
...
+}
+
+static void driveStateRunOnTick100Hz(void)
+{
+    float apps_pedal_percentage  = app_canRx_FSM_PappsMappedPedalPercentage_get() * 0.01f;
+    float sapps_pedal_percentage = app_canRx_FSM_SappsMappedPedalPercentage_get() * 0.01f;
+
+    if (!driveStatePassPreCheck())
+    {
+        return;
```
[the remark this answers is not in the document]
shayana18: I think the idea is that warning are only warnings if they should make torque to 0 atleast that's the convention they are supposed to follow so I'll just add it here and also add it to runOnExit for a safety

firmware/quintuna/VC/src/app/states/app_driveState.c
```diff
@@ -1,37 +1,246 @@
...
+    {
+        app_canTx_VC_TorqueVectoringEnabled_set(true);
+    }
+    
+    return true;
+}
+
+static void runDrivingAlgorithm(float apps_pedal_percentage, float sapps_pedal_percentage)
+{
+    if (app_faultCheck_checkSoftwareBspd(apps_pedal_percentage, sapps_pedal_percentage))
```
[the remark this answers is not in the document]
shayana18: Is being implemented, can comment out for now

gtaharaedmonds (approved):
lets get this money

PR #1642, opened 2025-06-06
RSM/FSM sensor validation 

shayana18:


shayana18:
RSM pump chip validated, is now shorted to 5V rail on RSM

PR #1646, opened 2025-06-08
DAM SD working

Jiajun-Huang:
### Changelist 
Clock tree change 
sdmmc configure change 
logfs write and read function change to make it more stable.
some clean up


Lucien950 (approved):
happy

PR #1651, opened 2025-06-09
Split app_segments into multiple source files

gtaharaedmonds:
Split app_segments into multiple .c files since it was getting pretty massive (and I still need to add a bunch of stuff). Also added voltage and temp stats.


Lucien950 (commented):
this got nothing to do with bms faults 🥀🥀🥀

Lucien950:
hating a 24 hour job fr

PR #1654, opened 2025-06-10
Quintuna BMS Power and Current Limiting

yawollopkcaj:
TODO:
- verify constants in app_powerCurrentLimit.c related to new battery specs
- implement high and low SOC based limiting (once app soc gets merged with master. see Quadrina Current Limiting pr for ref)
- implement low and high volt clamp based limiting (once app soc gets merged with master. see Quadrina Current Limiting pr for ref)
- !!! make current and power limits dependant i.e. if c lim is x and p lim is y then if x * voltage < y, x * voltage = y
- current derating for reaching near max current draw?


gtaharaedmonds (changes requested):
Looks great! Can you add some tests too? Definitely want this guy to be stable so we don't actually reduce/increase our limit accidentally.

Also: I thought the plan was to only broadcast a power limit, and use the current limit as another clamping factor in the power limit? Did we change that?

firmware/quintuna/BMS/src/app/app_bspdPrefault.c
Lucien950, line 1: wait this is a VC responsability please delete

Lucien950 (changes requested):

Lucien950 (changes requested):

KelmLelm (approved):
firreee pr

PR #1665, opened 2025-06-13
Enable Watchdogs on Quintuna

gtaharaedmonds:
### Changelist 

- Enabled watchdogs on all boards
- Set up solid alerts for watchdog timeout
- Send async CAN bootup on a reset

### Testing Done

- [ ] Test on all boards, ideally with spinning wheels


Lucien950 (approved):
stay winning goat

PR #1669, opened 2025-06-14
VC Control Rear Pump

Aditya-Dhiman4:
### Changelist 
added app layer pumps in rsm
added can msg in app layer pumps in vc that controls rsm pump


firmware/quintuna/RSM/src/app/app_pumpControl.c
Lucien950, line 1: surely we need to keep the logic centralized and coordinated, just RSM dumb pump it

Lucien950 (approved):
pog

PR #1673, opened 2025-06-14
Fix precharge voltage check

gtaharaedmonds:
### Changelist 

Fix precharge voltage check to not use hardcoded 300V. Also fix charger CAN.

### Testing Done

Tested briefly on battery (just CAN)


KelmLelm (approved):

Lucien950 (commented):
damn 😭

PR #1674, opened 2025-06-14
CONST 🥀🥀🥀

Lucien950:


KelmLelm (approved):
Idk about the VC changes, but all the const seems fine to me

PR #1677, opened 2025-06-16
Boot Config

Lucien950:
### Changelist 

**NOTE WE DO NOT NEED TO CHANGE CANUP CAUSE IT ALREADY REFLECTS THE CORRECTED IDS**
image
image


Jugveer-Sandher (approved):
noice mate

PR #1694, opened 2025-06-19
BMS boot tweaks

gtaharaedmonds:
Tweak task priorities on BMS bootloader to be in line with Quadruna and other bootloaders.


KelmLelm (approved):
Not entirely sure I know what the data params do, but I trust

PR #1697, opened 2025-06-20
Reroute Collisions Handling

Lucien950:


gtaharaedmonds (approved):
yep looks good, i also looked at generated code and the diffs are exactly as expected

PR #1704, opened 2025-07-20
Vehicle Model

Lucien950:


Aditya-Dhiman4 (approved):
looks good and works, we can add additional physics later

Aditya-Dhiman4 (dismissed):
Why does the normal force on each wheel look strange? The behaviour makes sense but the values are extremely high for the front right and rear right

PR #1727, opened 2025-09-26
Master

arraystartat0:
Updated files to give a feedback blink on the torque vectoring LED when the torque vectoring switch is in closed state and removed vanilla_sig code conflicting with PB2 since it was a comp hotfix.

### DIFF
--- a/firmware/quintuna/CRIT/src/io/io_switches.h
+++ b/firmware/quintuna/CRIT/src/io/io_switches.h
@@ -7,3 +7,5 @@ extern Switch regen_switch;
 extern Switch vanilla_override_switch;
 extern Switch launch_control_switch;
 extern Switch telem_switch;
+//new
+extern Switch torque_vectoring_switch;
--- a/firmware/quintuna/CRIT/src/tasks.c
+++ b/firmware/quintuna/CRIT/src/tasks.c
@@ -15,6 +15,8 @@
 #include "io_canQueue.h"
 #include "io_canRx.h"
 #include "io_bootHandler.h"
+//new
+#include "hw_gpios.h"   // for 'led' and torque_vectoring_sig
 
 // hw
 #include "hw_hardFaultHandler.h"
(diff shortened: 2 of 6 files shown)


firmware/quintuna/CRIT/src/io/io_switches.h
```diff
@@ -7,3 +7,5 @@ extern Switch regen_switch;
 extern Switch vanilla_override_switch;
 extern Switch launch_control_switch;
 extern Switch telem_switch;
+//new
```
shayana18, line 10: fine for tutorial projects but comments like this aren't necessary. Comments should only be added when they are there to either organize or provide context to something that can be easily confused

firmware/quintuna/CRIT/src/tasks.c
```diff
@@ -15,6 +15,8 @@
 #include "io_canQueue.h"
 #include "io_canRx.h"
 #include "io_bootHandler.h"
+//new
```
shayana18, line 18: same as above

firmware/quintuna/CRIT/src/tasks.c
```diff
@@ -136,13 +138,33 @@ void tasks_run100Hz()
...
         if (!hw_chimera_v2_enabled)
         {
             jobs_run100Hz_tick();
         }
 
+        //new: torque vectoring switch -> blink the MCU debug LED 
+        // Read the input directly.
+        const GPIO_PinState sw = HAL_GPIO_ReadPin(torque_vectoring_sig.port, torque_vectoring_sig.pin);
+        if (sw == GPIO_PIN_SET) {
+            if (++blink_div >= 25) {         // 25 * 10 ms = 250 ms
```
shayana18, line 156: Try to use macros for constants like this, they are great for providing variable like usability but take no memory as its a preprocessor directive
https://www.simplilearn.com/tutorials/c-tutorial/what-is-macros-in-c

PR #1737, opened 2025-10-04
telem backend docker update

j-ackyao:
abstracts some docker environments
updates requirements


j-ackyao:
docker serial port probably works on linux machine 👍

eriklsy (approved):
👍 hopefully no one screams about logfs-python

PR #1824, opened 2026-01-21
Digital Filters cpp refactor

arraystartat0:
### Changelist 
Refactored the digital filters library to cpp

### Testing Done
Unit testing -> all passed and built.


firmware/shared/src/app/app_sensor_filter.cpp
Lucien950, line 1: can you refactor this to be propertly object oriented, because rn it's using C style functions and objects

Lucien950 (changes requested):

firmware/shared/src/app/app_sensor_filter.cpp
arraystartat0: On it

Lucien950 (approved):

PR #1890, opened 2026-03-07
RSM App Validation

sakiyamadev:
### Changelist 
Validating the CAN project onto VectorCANoe


Lucien950 (dismissed):
you need to move the include of hw_gpios from io_brakelight.hpp to io_brakelight.cpp, but otherwise this looks good

AmirTajaddodi (approved):

PR #1903, opened 2026-03-11
HW Cordic Driver

Aditya-Dhiman4:
### Changelist 
image
Sensational

### Resolved Tickets

[FIRM-440](https://ubcformulaelectric.atlassian.net/browse/FIRM-440)

[FIRM-440]: https://ubcformulaelectric.atlassian.net/browse/FIRM-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


firmware/shared/srcpp/io/io_math.hpp
```diff
@@ -0,0 +1,56 @@
...
+ * to fixed point and scale it between -1 and 1 and then rescale the output which would take unecessary cycles.
+ * Thus the we only pass in angle and the user can simply multiply the result by the desired modulus outside. of
+ * the cordic driver.
+ */
+
+using namespace hw::cordic;
+
+namespace io::math
+{
+// TODO: is it annoying for usability to have std::expected here? default to standard library implementation or arm math
```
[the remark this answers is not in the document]
Aditya-Dhiman4: im gonna tickle u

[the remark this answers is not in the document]
Lucien950: he's right

firmware/shared/srcpp/hw/hw_cordic.hpp
```diff
@@ -0,0 +1,82 @@
...
+        mod += M_2PI_F;
+    }
+
+    return static_cast<int32_t>(mod * M_PI_INV * CONVERSION);
+}
+
+inline float convertToFloat(int32_t fixed_point)
+{
+    return static_cast<float>(fixed_point) * CONVERSION_INV;
+}
```
Lucien950: these seem like .CPP things and not .HPP things

firmware/shared/srcpp/hw/hw_cordic.hpp
```diff
@@ -0,0 +1,82 @@
...
+#endif
+
+#include "main.h"
+#ifndef HAL_CORDIC_MODULE_ENABLED
+#error "HAL_CORDIC_MODULE_ENABLED must be defined and set to 1"
+#endif
+
+#include "hw_utils.hpp"
+#include "util_errorCodes.hpp"
+#include "util_units.hpp"
```
Lucien950: consider what needs to be cpp and hpp here as well

firmware/dev/h5dev/src/jobs.cpp
```diff
@@ -8,7 +8,15 @@
 #include "hw_adc.hpp"
 #include "io_canMsg.hpp"
 #include <util_errorCodes.hpp>
+#include "io_math.hpp"
+#include <cmath>
+
+extern "C"
+{
 #include "main.h"
+}
```
Lucien950: no need to extern C it

Lucien950 (changes requested):
few things to look at

firmware/shared/srcpp/hw/hw_cordic.hpp
Aditya-Dhiman4: i mean its a pretty small function plus I have these exposed so they can easily be used in io math, other wise I would have to use this in the calculate function which would make it a bit messy

firmware/shared/srcpp/hw/hw_cordic.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * @brief Configure the CORDIC peripheral to:
+ * - Perform the desired math operation
+ * - Scaling factor
+ * - Width of input and output data
+ * - Number of 32 bit R/W expected before and after calculation
+ * - Precision: 1 - 15 cycles (Higher -> More Precision)
+ */
+std::expected<void, ErrorCode> configure(uint32_t func, uint32_t scale, uint32_t nbwrite);
+
+std::expected<void, ErrorCode> calculate(uint32_t nbwrite, std::span<const int32_t> args, std::span<int32_t> result);
```
Lucien950: why not just one function? also why are the sizes passed in independently of the 

i would also use templates to ensure that the size of the spans passed in are the same

i would also consider returning an array type instead of passing in a result span

Lucien950 (changes requested):
last thing

firmware/shared/srcpp/hw/hw_cordic.hpp
Aditya-Dhiman4: valid ill make it a single function. the sizes are passed independently because the peripheral has its own sizing macros with nbread and nbwrite but I can make that internal and have the function figure out which one to use on its own. depending on the cordic function used it may have a different amount of input and output arguments

Lucien950 (approved):
approve so you can merge once the last two comments are resolved

PR #1910, opened 2026-03-15
Shared Watchdog + Implementation on all boards

sakiyamadev:
### Changelist 
Implement hardware and software watchdog on RSM


Lucien950 (approved):
otherwise W pr

AmirTajaddodi (changes requested):
One quick note can u take a look at what we do for quintuna shared/hw_wachdog.c and implement that for the cpp version. Also please take a look what we do in tasks init. Also since I want you to learn and interact with other boards plz implement watchdog for all the board such as VC, RSM, CRIT, DAM and FSM

AmirTajaddodi (approved):

PR #1920, opened 2026-03-21
Hexray NTP

emilysxng:
- UART TX/RX contension fixed


firmware/shared/srcpp/hw/hw_uart.cpp
```diff
@@ -12,27 +12,77 @@ void hw::Uart::deinit() const
...
+    assert(rxTaskInProgress != nullptr);
+
+    BaseType_t higherPriorityTaskWoken = pdFALSE;
+    vTaskNotifyGiveFromISR(rxTaskInProgress, &higherPriorityTaskWoken);
+    portYIELD_FROM_ISR(higherPriorityTaskWoken);
+}
+
+void hw::Uart::onErrorFromISR() const
+{
+    // just wake both up lmao
```
Lucien950, line 35: this is indeed very suspicious but idk how to make this better

emilysxng: I already made these changes locally 😭

Lucien950: Rip I thought Erik said you pushed all your changes already

emilysxng: Erik is a liar

firmware/third_party/pybind11
Lucien950, line 1: DELETE!

Lucien950 (dismissed):

j-ackyao:
LGTM 👍

firmware/third_party/pybind11
j-ackyao: i gotchu 👍

firmware/quintuna/DAM/src/jobs.c
```diff
@@ -76,6 +77,8 @@ void jobs_init(void)
 
 void jobs_run1Hz_tick(void)
 {
+    transmitNTPStartMsg(); // Ping backend for t1, t2
```
Lucien950: please bind this to the interrupt on the button

firmware/hexray/DAM/src/app/app_ntp.hpp
```diff
@@ -0,0 +1,27 @@
...
+    uint64_t t3;
+};
+
+constexpr uint32_t PREDIV_S   = 999;
+constexpr uint64_t MS_PER_DAY = 86400000ULL;
+
+uint64_t      rtcTimeToMs(io::rtc::Time t);
+io::rtc::Time msToRtcTime(uint64_t ms);
+int64_t       computeOffset(const Timestamps &ts);
+bool          parseNTPPacketBody(std::span<uint8_t> body, Timestamps &ts);
```
Lucien950, line 21: how much of this API do you actually have to expose?

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (!io::rtc::get_time(t0))
+    {
+        LOG_ERROR("Could not get RTC time");
+        return;
+    }
+    ntpTimestamps.t0 = app::ntp::rtcTimeToMs(t0);
+
+    const io::telemMessage::NTPMsg ntp_msg = io::telemMessage::NTPMsg();
+    if (!_900k_uart.transmit(
+            std::span<const uint8_t>{ reinterpret_cast<const uint8_t *>(&ntp_msg), ntp_msg.wireSize() }))
```
Lucien950: i would suggest generating the span as part of the NTPMsg struct

Lucien950: or better yet in the TelemMessage parent struct

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,183 @@
...
+        return;
+    }
+
+    // --------------------------- Debug message
+    LOG_INFO(
+        "Sent NTP Msg! at time: %02u:%02u:%02u.%03lu", t0.hours, t0.minutes, t0.seconds,
+        (unsigned long)(999 - t0.subseconds));
+}
+
+void pollForRadioMessages(void)
```
Lucien950: i still think you should do the UART buffering method we discussed earlier

Lucien950: nice side effect of that is that all of this what seems like app code is in app

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,183 @@
...
+    uint8_t            headerData[7];
+    std::span<uint8_t> rxBufferHeader(headerData, 7);
+    auto               result = _900k_uart.receive(rxBufferHeader);
+    if (!result)
+    {
+        LOG_ERROR("Could not get rxBufferHeader, error %d", result.error());
+        return;
+    }
+
+    // Keep shifting until magic bytes found
```
Lucien950, line 79: ring queue o clock buddy

firmware/hexray/DAM/src/io/io_telemRx.cpp
Lucien950: also I would be careful in terms of building a structure where you can have multiple broadcasters over UART (not just broadcasting and rxing NTP), because you would have to protect the mutex for uart

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <stdint.h>
+#include <util_errorCodes.h>
+#include <cstring>
+
+static app::ntp::Timestamps ntpTimestamps;
+
+void io_telemRx()
+{
+    pollForRadioMessages();
+}
```
Lucien950: does this have to be a seperate function?

firmware/hexray/DAM/src/jobs.cpp
Lucien950, line 1: no more running things?

firmware/hexray/DAM/src/tasks.cpp
```diff
@@ -56,7 +59,7 @@ extern "C"
     uint32_t start_ticks = osKernelGetTickCount();
     forever
     {
-        HAL_IWDG_Refresh(&hiwdg);
```
Lucien950, line 59: 🤨

firmware/hexray/DAM/src/io/io_telemRx.hpp
Lucien950, line 1: no namespace or anything is crazy

firmware/quintuna/DAM/src/io/io_telemMessage.c
```diff
@@ -83,4 +83,14 @@ TelemCanMsg io_telemMessage_buildCanMsg(const CanMsg *rx_msg, const float time_o
...
+
+TelemNTPMsg io_buildNTPMessage(void)
+{
+    TelemNTPMsg ntp_msg;
+
+    ntp_msg.identifier = 2;
+    ntp_msg.header     = io_telemMessage_buildHeader(&ntp_msg.identifier, 1);
+
+    return ntp_msg;
+}
```
Lucien950: constructor

firmware/quintuna/DAM/src/io/io_telemMessage.h
```diff
@@ -51,4 +51,11 @@ typedef struct __attribute__((packed))
...
-TelemCanMsg io_telemMessage_buildCanMsg(const CanMsg *rx_msg, float time_offset, uint8_t *size);
\ No newline at end of file
+typedef struct __attribute__((packed))
+{
+    TelemMessageHeader header;
+    uint8_t            identifier;
+} TelemNTPMsg;
+
+TelemCanMsg io_telemMessage_buildCanMsg(const CanMsg *rx_msg, float time_offset, uint8_t *size);
+TelemNTPMsg io_buildNTPMessage(void);
```
Lucien950: constructor

firmware/quintuna/DAM/src/io/io_telemRx.c
Lucien950, line 1: i'll ignore

firmware/quintuna/DAM/src/io/io_telemRx.h
Lucien950, line 1: are we using any of this? i really hope no

Lucien950 (changes requested):

firmware/quintuna/DAM/src/io/io_telemRx.h
eriklsy: hallucinations

firmware/hexray/DAM/src/jobs.cpp
eriklsy: originally had hw uart call inside jobs. Decided to inline the functionality into tasks

firmware/hexray/DAM/src/tasks.cpp
eriklsy: oop

firmware/hexray/DAM/src/jobs.cpp
Lucien950: i would be careful about this in order to make the code testable

eriklsy: then the hw needs to be in some wrapper? Thats what I had on master but I thought we thought it was funny. Or maybe this running logic is app code

firmware/hexray/DAM/src/io/io_telemRx.cpp
eriklsy: having a single queue could potentially make the ntp messages offset weird, I'll add a mutex for uart I think its best for now

.vscode/settings.json
```diff
@@ -13,5 +13,5 @@
     "--query-driver=/usr/local/bin/arm-none-eabi-gcc"
   ],
   "C_Cpp.default.cppStandard": "c++23",
-  "cortex-debug.variableUseNaturalFormat": true
+  "cortex-debug.variableUseNaturalFormat": false
```
Aditya-Dhiman4, line 16: What the fucky

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,61 @@
+#include "app_ntp.hpp"
+
+#include <cstring>
+#include <cstdint>
+
+constexpr uint32_t PREDIV_S   = 999;
+constexpr uint64_t MS_PER_DAY = 86400000ULL;
```
Aditya-Dhiman4, line 7: Static this

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace app::ntp
+{
+
+uint64_t rtcTimeToMs(io::rtc::Time t)
```
Aditya-Dhiman4, line 12: Should this be pass by reference?

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+    uint32_t ms_part = PREDIV_S - t.subseconds;
+    return static_cast<uint64_t>(t.hours) * 3600000ULL + static_cast<uint64_t>(t.minutes) * 60000ULL +
+           static_cast<uint64_t>(t.seconds) * 1000ULL + static_cast<uint64_t>(ms_part);
+}
+
+io::rtc::Time msToRtcTime(uint64_t ms)
```
Aditya-Dhiman4, line 19: Const argument

firmware/hexray/DAM/src/app/app_ntp.cpp
Aditya-Dhiman4, line 1: Consider making some of these functions static or putting it into an anonymous namespace. Just make it clearer what logic u are exposing to the user

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+int64_t computeOffset(const Timestamps &ts)
+{
+    return ((int64_t)(ts.t1 - ts.t0) + (int64_t)(ts.t2 - ts.t3)) / 2;
+}
+
+bool parseNTPPacketBody(std::span<uint8_t> body, Timestamps &ts)
+{
+    if (body.size() < 17)
```
Aditya-Dhiman4, line 43: Magic number

firmware/hexray/DAM/src/io/io_ntpButton.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+namespace io::ntpButton
+{
+bool wasJustPressed()
+{
+    static bool prev_pressed = false;
+    const bool  now_pressed  = ntp_pin.readPin();
+    const bool  rising_edge  = now_pressed && !prev_pressed;
+    prev_pressed             = now_pressed;
+    return rising_edge;
```
Aditya-Dhiman4, line 12: Add denounce?

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,61 @@
...
+uint64_t rtcTimeToMs(io::rtc::Time t)
+{
+    uint32_t ms_part = PREDIV_S - t.subseconds;
+    return static_cast<uint64_t>(t.hours) * 3600000ULL + static_cast<uint64_t>(t.minutes) * 60000ULL +
+           static_cast<uint64_t>(t.seconds) * 1000ULL + static_cast<uint64_t>(ms_part);
+}
+
+io::rtc::Time msToRtcTime(uint64_t ms)
+{
+    io::rtc::Time t{};
```
Aditya-Dhiman4, line 21: Lowkey the logic of this struct could be changed, you could give a constructor where u pass ms and anything else as an arg and in the constructor do all of those calculations in the constructor, idk if i am cooking here or not

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+// Send message to backend through radio to get t1,t2. Called periodically
+std::expected<void, ErrorCode> io::telemRx::transmitNTPStartMsg()
+{
+    // Take the UART lock first so any wait for an in-flight transmit happens *before* we capture t0.
+    hw::MutexGuard g{ _900k_uart_tx_mutex };
+
+    // Take note of the sending time (t0).
+    io::rtc::Time t0;
+    const auto    t0_result = io::rtc::get_time(t0);
```
Aditya-Dhiman4, line 43: Maybe make this a method in the time struct?

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,192 @@
...
+            return std::unexpected(ErrorCode::ERROR);
+        }
+    }
+
+    LOG_INFO(
+        "Header: %02X %02X %02X %02X %02X %02X %02X", rxBufferHeader[0], rxBufferHeader[1], rxBufferHeader[2],
+        rxBufferHeader[3], rxBufferHeader[4], rxBufferHeader[5], rxBufferHeader[6]);
+
+    // Read rest of packet (contains t1, t2) using size given to you
+    uint8_t size = rxBufferHeader[2];
```
Aditya-Dhiman4, line 101: Const?

Aditya-Dhiman4 (commented):
Bomboclattt

firmware/hexray/DAM/src/app/app_ntp.cpp
eriklsy: huh like modify the time directly?

.vscode/settings.json
eriklsy: oops dunno how that happened

firmware/hexray/DAM/src/app/app_ntp.cpp
Aditya-Dhiman4: Ur not really modifying the time here it will just prevent a copy

Aditya-Dhiman4: Because you need a time struct in order to run this function, you could make this function a member of the struct

Aditya-Dhiman4 (commented):
More bomboclatt comments

firmware/hexray/DAM/src/app/app_ntp.cpp
eriklsy: ehhh maybe in future

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (body.size() < MIN_NTP_PACKET_SIZE)
+        return false;
+
+    if (body[0] != 1) // message id
+        return false;
+
+    uint64_t t1;
+    uint64_t t2;
+    std::memcpy(&t1, &body[1], sizeof(uint64_t));
+    std::memcpy(&t2, &body[9], sizeof(uint64_t));
```
Aditya-Dhiman4: Why mus we explicitly mem copy here can we not just set the variables as normal?

firmware/shared/srcpp/util/util_ringBuffer.hpp
Aditya-Dhiman4, line 1: Have you checked if FreeRTOS have thread safe queue/ring buffer implementations? Im sure it does so you would not have to implement your own, plus if you really wanted you could make a C++ wrapper around that too

https://freertos.org/Documentation/02-Kernel/02-Kernel-features/02-Queues-mutexes-and-semaphores/01-Queues

Aditya-Dhiman4: I would ask zeddy about this to confirm

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,62 @@
...
+    {
+        LOG_ERROR("Failed to transmit NTP message");
+        return std::unexpected(ErrorCode::ERROR);
+    }
+
+    LOG_INFO(
+        "Sent NTP Msg! at time: %02u:%02u:%02u.%03lu", t0.hours, t0.minutes, t0.seconds,
+        static_cast<unsigned long>(999 - t0.subseconds));
+    return t0;
+}
```
Aditya-Dhiman4: This critical section is quite large, plus I worry in the event the uart peripheral somehow fails and becomes blocking (although highly unlikely). Could we add some timeout here where we free the mutex so we dont potentially block other tasks? 

Additionally could we shorten this critical section so that the mutex is freed before `if (!tx_result)`? Plus if there is anything else we could take out of the critical section like getting time, although that may not be thread safe.

i would say the most important change would be to ensure this critical section does not block for too long at all costs, and shortening the critical section may be benefital.

firmware/hexray/DAM/src/io/io_ntpButton.cpp
Aditya-Dhiman4, line 1: this is a later thing but i have a small refactor on `io_swiches.cpp` that includes debouncing so you can use that instead of making ur own debounce. Dont worry about this now though just letting u know

firmware/hexray/DAM/src/io/io_ntpButton.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+namespace io::ntpButton
+{
+// Returns true exactly once per rising edge
+// Must be called at a fixed cadence from a SINGLE caller
+bool wasJustPressed();
```
Aditya-Dhiman4: nit: [[nodiscard]]?

firmware/hexray/DAM/src/app/app_ntp.cpp
eriklsy: backend sending byte array 👍

firmware/shared/srcpp/hw/hw_mutexGuard.hpp
Lucien950, line 1: wow

Lucien950: cinema

firmware/hexray/DAM/src/hw/hw_uarts.cpp
```diff
@@ -5,6 +5,15 @@
 
 constexpr hw::Uart _900k_uart{ huart2 };
 
+// Because of NTP needing the h
+static StaticSemaphore_t _900k_uart_tx_mutex_storage;
+SemaphoreHandle_t        _900k_uart_tx_mutex = nullptr;
+
+void hw_uarts_init()
+{
+    _900k_uart_tx_mutex = xSemaphoreCreateMutexStatic(&_900k_uart_tx_mutex_storage);
```
Lucien950: can you just implement this as part of the uart, I'm uninterested in trying to check if every single call to transmit/recieve is guarded with mutex

firmware/hexray/DAM/src/io/io_ntpButton.cpp
Lucien950: +1

firmware/shared/srcpp/util/util_ringBuffer.hpp
```diff
@@ -0,0 +1,71 @@
...
+        }
+        size_ += src.size();
+        return {};
+    }
+
+    // Read T at logical offset from head without consuming.
+    [[nodiscard]] T peek(std::size_t offset) const { return buf_[(head_ + offset) % N]; }
+
+    // Copy out a contiguous logical run starting at offset, without consuming.
+    [[nodiscard]] std::expected<void, ErrorCode> copyOut(std::size_t offset, std::span<T> dst) const noexcept
```
Lucien950: at offset rather than at front? wouldn't that make your data discontiguous?

firmware/hexray/DAM/src/io/io_ntpButton.cpp
Lucien950: lowkey if anything I would just use io to wrap directly the button, then debounce in app layer

firmware/hexray/DAM/src/io/io_crc.cpp
Lucien950, line 1: lowkey if the hardware is blocking the CRC just calculate it in software lmao

firmware/hexray/DAM/src/io/io_telemRx.cpp
Lucien950, line 1: this is all app level behaviours

firmware/hexray/DAM/src/jobs.cpp
```diff
@@ -26,8 +30,12 @@ void jobs_init()
...
         [](const JsonCanMsg &tx_msg)
         {
-            const io::CanMsg msg = app::jsoncan::copyToCanMsg(tx_msg);
-            LOG_IF_ERR(can_tx_queue.push(msg));
+            const io::CanMsg msg    = app::jsoncan::copyToCanMsg(tx_msg);
+            auto             result = can_tx_queue.push(msg);
+            if (not result)
+                LOG_ERROR("Failed to push TX CAN message: %d", static_cast<int>(result.error()));
+            (void)telem_tx_queue.push(
+                io::telemMessage::TelemCanMsg(msg, static_cast<uint64_t>(io::time::getCurrentMs())));
```
Lucien950: use LOG_IF_ERR

Lucien950: also ur not supposed to push every single message, only the ones which pass the filter

firmware/hexray/DAM/src/jobs.cpp
```diff
@@ -37,6 +45,18 @@ void jobs_init()
...
+        const auto ntp_result = io::telemRx::transmitNTPStartMsg();
+        if (!ntp_result)
+        {
+            LOG_ERROR("transmitNTPStartMsg() failed with error: %d", static_cast<int>(ntp_result.error()));
+        }
+        else
+        {
+            app::ntp::recordT0(app::ntp::rtcTimeToMs(*ntp_result));
+        }
+    }
```
Lucien950: can you create a seperate task for ntp?

firmware/hexray/DAM/test/test_app_telemRx.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "io_rtc.hpp"
+
+// Test-only spies on the fake RTC, defined in fake_io_rtc.cpp.
+namespace fakes::rtc
+{
+void          setNow(const io::rtc::Time &t);
+io::rtc::Time lastSetTime();
+bool          wasSetCalled();
+void          reset();
+} // namespace fakes::rtc
```
Lucien950, line 25: don't do this create a header file
this is the same reason we don't randomly grab externs from anywhere

firmware/hexray/DAM/test/test_app_telemRx.cpp
```diff
@@ -0,0 +1,171 @@
...
+{
+    std::vector<uint8_t> body(17, 0);
+    body[0] = 1; // MessageId::Ntp
+    std::memcpy(&body[1], &t1, sizeof(uint64_t));
+    std::memcpy(&body[9], &t2, sizeof(uint64_t));
+
+    std::vector<uint8_t> frame{ 0xCC, 0x33, static_cast<uint8_t>(body.size()), 0, 0, 0, 0 };
+    frame.insert(frame.end(), body.begin(), body.end());
+    return frame;
+}
```
Lucien950: just build the array statically

firmware/hexray/DAM/test/test_app_telemRx.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+TEST_F(TelemRxTest, SlewsRtcToNextSecondBoundary)
+{
+    // Seed the RTC to 0:0:0.000 (subseconds=999 means "0 ms into this second").
+    io::rtc::Time start{};
+    start.hours      = 0;
+    start.minutes    = 0;
+    start.seconds    = 0;
+    start.subseconds = 999;
+    fakes::rtc::setNow(start);
```
Lucien950: construct temporary and inline

firmware/hexray/DAM/src/app/app_telemRx.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+inline constexpr uint8_t     kMagic0     = 0xCC;
+inline constexpr uint8_t     kMagic1     = 0x33;
+inline constexpr std::size_t kHeaderSize = 7; // magic(2) + size(1) + crc(4)
+
+enum class MessageId : uint8_t
+{
+    NTP        = 1,
+    Remote_NTP = 2,
+};
```
Lucien950: this does not need to be public?

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,61 @@
...
+        return std::unexpected(ErrorCode::ERROR);
+    }
+
+    io::rtc::Time t3{};
+    const auto    t3_result = io::rtc::get_time(t3);
+    if (!t3_result)
+    {
+        LOG_ERROR("telemRx: could not capture t3");
+        return std::unexpected(t3_result.error());
+    }
```
Lucien950: just call it t

firmware/hexray/DAM/src/io/io_telemRx.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    LOG_INFO(
+        "Sent NTP Msg! at time: %02u:%02u:%02u.%03lu", t0.hours, t0.minutes, t0.seconds,
+        static_cast<unsigned long>(999 - t0.subseconds));
+    return t0;
+}
+
+// TODO: replace blocking chunk read with UART-IDLE / DMA driven ingestion so
+// rx_time is captured at the byte-arrival instant rather than at chunk end.
+std::expected<io::telemRx::RxChunk, ErrorCode> io::telemRx::pumpOnce(std::span<uint8_t> scratch)
```
Lucien950: this function is very confusing because you return the span that you gave it with a timestamp? like either commit to
1. return an array with a timestamp (which I prefer)
2. modify in place the span and return a time

firmware/hexray/DAM/src/io/io_telemMessage.hpp
```diff
@@ -1,17 +1,19 @@
...
 enum class TelemMessageIds : uint8_t
 {
-    CAN         = 1,
-    NTP         = 2,
-    NTPDate     = 3,
-    BaseTimeReg = 4,
+    CAN     = 1,
+    NTP     = 2,
+    NTPDate = 3,
+    // BaseTimeReg = 4, deprecated see quintuna for previous definitions. delete later
```
Lucien950: why depr? are we not sending relative time anymore?

firmware/hexray/DAM/src/io/io_telemMessage.hpp
```diff
@@ -23,40 +25,41 @@ struct [[gnu::packed]] Header
...
     CanMsgBody msg;
     TelemCanMsg() = default;
-    explicit TelemCanMsg(const io::CanMsg &rx_msg, float time_offset);
-    [[nodiscard]] size_t wireSize() const;
+    explicit TelemCanMsg(const io::CanMsg &rx_msg, uint64_t time_offset);
+};
+
+struct [[gnu::packed]] NTPMsg : TelemMessage
+{
+    uint8_t identifier;
```
Lucien950: what

Lucien950: make this the correct type: TelemMessageIds 
also make it so that the parent class holds this (because all types of telem message needs it) and initialize from the constructor

firmware/hexray/DAM/src/io/io_telemMessage.hpp
```diff
@@ -23,40 +25,41 @@ struct [[gnu::packed]] Header
...
 
-struct [[gnu::packed]] TelemCanMsg
+struct [[gnu::packed]] TelemMessage
+{
+    Header                                 header;
+    [[nodiscard]] size_t                   wireSize() const { return sizeof(Header) + header.payload_size; }
+    [[nodiscard]] std::span<const uint8_t> asBytes() const
+    {
+        return { reinterpret_cast<const uint8_t *>(this), wireSize() };
+    }
```
Lucien950, line 51: i would use TelemMessage as a utility class, and have the memory layout managed by a seperate child struct

firmware/hexray/DAM/src/io/io_telemRx.hpp
```diff
@@ -0,0 +1,25 @@
...
+};
+
+// TX side: transmit the NTP start message and return the RTC time captured
+// just before transmission (this is the t0 the APP layer needs).
+std::expected<io::rtc::Time, ErrorCode> transmitNTPStartMsg();
+
+// RX side: block on the radio UART, fill `scratch`, and return the chunk +
+// receive timestamp. Caller owns `scratch` and forwards the result to whatever
+// consumer wants it (no app coupling here).
+std::expected<RxChunk, ErrorCode> pumpOnce(std::span<uint8_t> scratch);
```
Lucien950: pump is crazy work just call it like read or something

firmware/hexray/DAM/src/tasks.cpp
```diff
@@ -57,29 +66,46 @@
...
     {
-        jobs_runTelemRx();
+        const auto rx_result = io::telemRx::pumpOnce(scratch);
+        if (!rx_result)
+        {
+            LOG_ERROR("pumpOnce() failed with error: %d", static_cast<int>(rx_result.error()));
+            continue;
+        }
+        app::telemRx::ingest(rx_result->bytes, app::ntp::rtcTimeToMs(rx_result->rx_time));
+        app::telemRx::drain();
```
Lucien950: i argued, and the central conceit of the design I proposed is that ingest and drain are in seperate tasks, so that drain can be blocking

Lucien950: the key feature this would require is a way to block until the queue is at least a certain size

firmware/hexray/DAM/src/tasks.cpp
```diff
@@ -112,30 +138,31 @@
         const auto msg = can_rx_queue.pop();
         if (not msg)
             continue;
-        io::can_rx::updateRxTableWithMessage(app::jsoncan::copyFromCanMsg(msg.value()));
+        const auto &can_msg = msg.value();
+        (void)telem_tx_queue.push(
```
Lucien950: LOG_IF_ERR if you dont' want to handle ti

Lucien950: but this you should def handle

firmware/hexray/DAM/src/app/app_telemRx.hpp
```diff
@@ -0,0 +1,39 @@
...
+// rx_time_ms is the RTC time captured immediately after this chunk landed;
+// it is associated with any frame whose last byte falls inside this chunk.
+void ingest(std::span<const uint8_t> bytes, uint64_t rx_time_ms);
+
+// Drain every complete, validated frame currently in the ring buffer.
+void drain();
+
+// Diagnostic / test access.
+std::size_t ringSize();
+void        reset();
```
Lucien950, line 31: ifdef test

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    uint32_t readU32Le(std::size_t offset)
+    {
+        uint32_t v = 0;
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 0)) << 0;
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 1)) << 8;
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 2)) << 16;
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 3)) << 24;
+        return v;
+    }
```
Lucien950: lowkey keep this in the queue object

firmware/hexray/DAM/src/app/app_telemRx.hpp
Lucien950: also what is up with k

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+        std::span<uint8_t>                body{ body_buf.data(), body_size };
+        if (!g_ring.copyOut(kHeaderSize, body))
+            return false;
+
+        if (io::crc::calculatePayloadCrc(body) != expected_crc)
+        {
+            LOG_ERROR("telemRx: CRC mismatch, resyncing");
+            g_ring.discard(1); // resync one byte at a time
+            return true;
+        }
```
Lucien950: do crc check basically as soon as you get the header so you can discard faster and prevent above copy

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+        g_ring.discard(kHeaderSize + body_size);
+        dispatchFrame(body, g_last_rx_time_ms);
```
Lucien950: uh i think you need a better way of keeping track of the time this seems very very suspecious in fact it's definately wrong

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 2)) << 16;
+        v |= static_cast<uint32_t>(g_ring.peek(offset + 3)) << 24;
+        return v;
+    }
+
+    // Dispatch the frame to the appropriate handler based on the message id
+    void dispatchFrame(std::span<const uint8_t> body, uint64_t rx_time_ms)
+    {
+        if (body.empty())
+            return;
```
Lucien950, line 68: i think in general you need asserts to make sure that the span passed it in is the right size

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+                const auto    set_res = io::rtc::set_time(tuned);
+                if (!set_res)
+                {
+                    LOG_ERROR("telemRx: RTC set_time failed");
+                    return;
+                }
+                LOG_INFO(
+                    "Tuned RTC! New Time: %02u:%02u:%02u.%03lu", tuned.hours, tuned.minutes, tuned.seconds,
+                    static_cast<unsigned long>(PREDIV_S - tuned.subseconds));
+                break;
```
Lucien950: just pass it to app_ntp ✌️😭 🥀

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+        {
+            case MessageId::NTP:
+            {
+                io::rtc::Time now{};
+                const auto    t = io::rtc::get_time(now);
+                if (!t)
+                {
+                    LOG_ERROR("telemRx: RTC get_time failed");
+                    return;
+                }
```
Lucien950: all this work to keep track of when a packet arrives in the queue just to throw it all away here

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+                    io::time::delay(wait_ms);
+                }
+
+                io::rtc::Time tuned   = app::ntp::msToRtcTime(*new_ms + wait_ms);
+                const auto    set_res = io::rtc::set_time(tuned);
+                if (!set_res)
+                {
+                    LOG_ERROR("telemRx: RTC set_time failed");
+                    return;
+                }
```
Lucien950: try a few times come on

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,189 @@
...
+                else
+                {
+                    app::ntp::recordT0(app::ntp::rtcTimeToMs(*ntp_result));
+                    LOG_INFO("telemRx: Remote NTP transmit successful");
+                    LOG_INFO(
+                        "Remote NTP Msg! at time: %02u:%02u:%02u.%03lu", ntp_result->hours, ntp_result->minutes,
+                        ntp_result->seconds, static_cast<unsigned long>(999 - ntp_result->subseconds));
+                }
+                break;
```
Lucien950: again just pass it to app_ntp

firmware/hexray/DAM/src/app/app_telemRx.cpp
Lucien950, line 1: handle only what you need to handle in this module please don't have such tight coupling to other modules

Lucien950 (changes requested):

firmware/shared/srcpp/util/util_ringBuffer.hpp
eriklsy: its basically a larger peek

eriklsy: offset if u care

eriklsy: wait a bit differnet than peek but it doesn't clear the buffer so its okay

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
+#include "app_ntp.hpp"
+
+#include <atomic>
```
Lucien950, line 3: please test this on hardware because I am not entirely confident this will work

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // Wire format is little endian
+    uint64_t t1;
+    uint64_t t2;
+    std::memcpy(&t1, &body[1], sizeof(uint64_t));
+    std::memcpy(&t2, &body[9], sizeof(uint64_t));
+
+    ts.t1 = t1 % MS_PER_DAY;
+    ts.t2 = t2 % MS_PER_DAY;
+    return true;
+}
```
Lucien950: this is a parsing responsibility, not an NTP responsibility. Please parse in a seperate module and pass to this module an agreed data structure (probably a struct)

firmware/hexray/DAM/src/app/app_ntp.cpp
Lucien950: i agree with what he's saying, that this function should just be a constructor of io::rtc::Time

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+    remaining_ms %= MS_PER_MINUTE;
+
+    t.seconds = static_cast<uint8_t>(remaining_ms / MS_PER_SECOND);
+
+    const uint32_t ms_remainder = static_cast<uint32_t>(remaining_ms % MS_PER_SECOND);
+    t.subseconds                = PREDIV_S - ms_remainder;
+    return t;
+}
+
+int64_t computeOffset(const Timestamps &ts)
```
Lucien950: static

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+void setNtpInProgress()
+{
+    g_ntp_in_progress.store(true);
+}
+
+void clearNtpInProgress()
+{
+    g_ntp_in_progress.store(false);
+}
```
Lucien950: this should probably be handled by this module?

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+bool parseNTPPacketBody(std::span<const uint8_t> body, Timestamps &ts)
+{
+    if (body.size() < MIN_NTP_PACKET_SIZE)
+        return false;
+
+    // Wire format is little endian
+    uint64_t t1;
+    uint64_t t2;
+    std::memcpy(&t1, &body[1], sizeof(uint64_t));
+    std::memcpy(&t2, &body[9], sizeof(uint64_t));
```
Lucien950: cast the point to the body into a pointer of a struct and interpret from ther

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    ts.t1 = t1 % MS_PER_DAY;
+    ts.t2 = t2 % MS_PER_DAY;
```
Lucien950: wait NTP should be date independent because it uses unix time?

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+bool handleFrameAndTuneRtc(std::span<const uint8_t> body, uint64_t t3_ms)
+{
+    io::rtc::Time now{};
+    const auto    get_res = io::rtc::get_time(now);
+    if (!get_res)
+    {
+        LOG_ERROR("ntp: RTC get_time failed");
+        clearNtpInProgress();
+        return false;
+    }
```
Lucien950: retry a few times

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,168 @@
...
+    int64_t       new_ms = static_cast<int64_t>(current_rtc_ms) + theta;
+    if (new_ms < 0)
+        new_ms = 0;
+    return static_cast<uint64_t>(new_ms);
+}
+
+bool handleFrameAndTuneRtc(std::span<const uint8_t> body, uint64_t t3_ms)
+{
+    io::rtc::Time now{};
+    const auto    get_res = io::rtc::get_time(now);
```
Lucien950: i guess this slipped through the cracks in the first review, but make sure that io::rtc::get_time returns an expected<io::rtc::Time, ErrorCode> rather than taking in a reference and populating it

Lucien950: in fact the default constructor with no argument should be deleted

firmware/shared/srcpp/util/util_ringBuffer.hpp
Lucien950: wait no but it would still make ur data discontiguous

firmware/hexray/DAM/src/cubemx/Src/stm32h5xx_hal_msp.c
Lucien950, line 1: reset this to 5

firmware/hexray/DAM/src/cubemx/DAM.ioc
Lucien950, line 1: why revert CAN settings/

firmware/hexray/DAM/src/hw/hw_uarts.cpp
```diff
@@ -5,6 +5,8 @@
 
 constexpr hw::Uart _900k_uart{ huart2 };
 
+void hw_uarts_init() {}
```
Lucien950: wut

just inline this function in the header file

firmware/hexray/DAM/src/io/io_ntpButton.cpp
Lucien950, line 1: lowkey I would tank the ifdefs to inline this in the header file, since it's so short

firmware/hexray/DAM/src/jobs.cpp
Lucien950: the whole point of io layer is an interface for testability

Lucien950 (dismissed):
more things? lemme just write this for now I still have thoughts

firmware/hexray/DAM/src/app/app_ntp.cpp
eriklsy: atomic? or this whole thing lmao

firmware/hexray/DAM/src/app/app_ntp.cpp
eriklsy: our rtc exposes date structs now so Im not sure, that was done earlier i think by emily

firmware/hexray/DAM/src/app/app_epochClock.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include "io_rtc.hpp"
+
+namespace app::epochClock
+{
+
+namespace
+{
+    // STM32 RTC stores year as [0, 99] representing 2000..2099.
+    constexpr int RTC_YEAR_BASE = 2000;
```
Aditya-Dhiman4, line 12: shouldnt this be unsigned?

firmware/hexray/DAM/src/app/app_epochClock.cpp
```diff
@@ -0,0 +1,130 @@
...
+        const int64_t  era = (z >= 0 ? z : z - 146096) / 146097;
+        const unsigned doe = static_cast<unsigned>(z - era * 146097);
+        const unsigned yoe = (doe - doe / 1460 + doe / 36524 - doe / 146096) / 365;
+        const int      y   = static_cast<int>(yoe) + static_cast<int>(era) * 400;
+        const unsigned doy = doe - (365 * yoe + yoe / 4 - yoe / 100);
+        const unsigned mp  = (5 * doy + 2) / 153;
+        const unsigned d   = doy - (153 * mp + 2) / 5 + 1;
+        const unsigned m   = mp < 10 ? mp + 3 : mp - 9;
+        return { y + (m <= 2 ? 1 : 0), m, d };
+    }
```
Aditya-Dhiman4, line 54: this could be a constructor in the struct

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace
+{
+    Timestamps        g_ts{};
+    std::atomic<bool> g_ntp_in_progress{ false };
+
+    void clearInProgress()
+    {
+        g_ntp_in_progress.store(false);
+    }
```
Aditya-Dhiman4: you can just inline this as this is a trivial method and only used by this file

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return ((int64_t)(ts.t1 - ts.t0) + (int64_t)(ts.t2 - ts.t3)) / 2;
+    }
+
+    void recordT0(uint64_t t0_ms)
+    {
+        g_ts.t0 = t0_ms;
+    }
+
+    uint64_t handleFrame(uint64_t t1, uint64_t t2, uint64_t t3_ms, uint64_t current_rtc_ms)
```
Aditya-Dhiman4: could make args const

firmware/hexray/DAM/src/app/app_ntp.cpp
Aditya-Dhiman4, line 1: you can probably make a lot of arguments const throught this code

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+bool handleFrameAndTuneRtc(uint64_t t1, uint64_t t2, uint64_t t3_ms)
+{
+    std::expected<uint64_t, ErrorCode> current_epoch_ms = std::unexpected(ErrorCode{});
+    for (uint8_t attempt = 0; attempt < RTC_GET_MAX_ATTEMPTS; ++attempt)
+    {
+        current_epoch_ms = app::epochClock::getEpochMs();
+        if (current_epoch_ms)
+            break;
+    }
```
Aditya-Dhiman4, line 77: zeddy made a retry abstraction in util_retry.hpp u can check that out and possibly replace it

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (!current_epoch_ms)
+    {
+        LOG_ERROR("ntp: RTC getEpochMs failed");
+        clearInProgress();
+        return false;
+    }
+
+    const uint64_t new_ms = handleFrame(t1, t2, t3_ms, *current_epoch_ms);
```
Aditya-Dhiman4: nit: * operator is fine but kinda looks like dereferencing a pointer, could use .value()

firmware/hexray/DAM/src/app/app_ntp.cpp
```diff
@@ -0,0 +1,127 @@
...
+#ifdef TARGET_TEST
+const Timestamps &timestamps()
+{
+    return g_ts;
+}
+#endif
+
+bool handleFrameAndTuneRtc(uint64_t t1, uint64_t t2, uint64_t t3_ms)
+{
+    std::expected<uint64_t, ErrorCode> current_epoch_ms = std::unexpected(ErrorCode{});
```
Aditya-Dhiman4: What error code are u initializing it to? it should be std::unexpected(ErrorCode::ERROR) or something along the lines of that

firmware/hexray/DAM/src/app/app_telemRx.cpp
```diff
@@ -0,0 +1,211 @@
...
+        uint64_t t2;
+    } __attribute__((packed));
+
+    static_assert(sizeof(NtpReplyWire) == ntpBodySize);
+
+    NtpReply parseNtpReply(std::span<const uint8_t> body)
+    {
+        const auto *wire = reinterpret_cast<const NtpReplyWire *>(body.data());
+        return NtpReply{ wire->t1, wire->t2 };
+    }
```
Aditya-Dhiman4, line 57: very constructorable

software/tracksight/rust_backend/src/tasks/serial_handler.rs
```diff
@@ -151,6 +151,13 @@ pub async fn run_serial_task(
 
                     match telem_message {
                         TelemetryIncomingMessage::Can { body } => {
+                            println!(
```
j-ackyao: Probably use the `vprintln!()` macro to prevent spamming in the terminal, can be enabled by `cargo run --features verbose`

Aditya-Dhiman4 (approved):
Image
approb

Aditya-Dhiman4:
image

Aditya-Dhiman4:
image

eriklsy (approved):

PR #1954, opened 2026-04-08
enum and alert

j-ackyao:
endpoints changed @Eroxl 

also updated docker configuration to run with backend and db


Eroxl:
My king ❤️

eriklsy (approved):
w

PR #1970, opened 2026-04-23
CRIT Switches and Buttons

Aditya-Dhiman4:


firmware/shared/srcpp/io/io_switch.cpp
Lucien950, line 1: please nuke this in favour of using app_signals to debounce
also this is app layer logic not io layer. I consider io layer to be a unfiltered representation of the hardware

Lucien950: like io_switch is def doing way too much

Lucien950 (changes requested):

firmware/shared/srcpp/io/io_switch.cpp
Aditya-Dhiman4: we used to have debouncing in io switch, shouldnt io layer give us information in the most reliable way possible

Aditya-Dhiman4: so do u want me to make an app switch object or something because then your gonna have crazy code duplication

Lucien950 (approved):

PR #1973, opened 2026-04-26
VC Drive State

Aditya-Dhiman4:
### Resolved Tickets

[FIRM-653](https://ubcformulaelectric.atlassian.net/browse/FIRM-653)

[FIRM-653]: https://ubcformulaelectric.atlassian.net/browse/FIRM-653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lucien950:
fuck it just merge it in lemme just review vc_state_machine

firmware/hexray/VC/src/app/app_sbgEllipse.cpp
Lucien950, line 1: surely all of this is just one function which returns immediately the error code???

firmware/hexray/VC/src/app/torque_vectoring/driving_algorithm.hpp
Lucien950, line 1: just implement this with the controls implementation cause this is very disconnected with what the API looks like there

firmware/hexray/VC/src/app/states/drive_state.cpp
Lucien950, line 1: rassclart

firmware/hexray/VC/src/app/states/app_driveState.cpp
```diff
@@ -0,0 +1,93 @@
...
+        return false;
+    // check possibly other faults
+
+    // TODO: Should this be done at FSM?
+    bool papps_ocsc = app::can_rx::FSM_Warning_PappsOCSC_get();
+    bool sapps_ocsc = app::can_rx::FSM_Warning_SappsOCSC_get();
+
+    // throttle pedal position is unreliable, do not drive
+    if (app::can_rx::FSM_Warning_AppsDisagreement_get() || (papps_ocsc && sapps_ocsc))
+        return false;
```
Lucien950: this already a warning we don't need to do all this

firmware/hexray/VC/src/app/states/app_driveState.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    // if primary apps is unreliable, use secondary, else use primary
+    if (papps_ocsc)
+        apps = app::can_rx::FSM_SappsMappedPedalPercentage_get();
+    else
+        apps = app::can_rx::FSM_PappsMappedPedalPercentage_get();
+
```
Lucien950: bomboclatt

firmware/hexray/VC/src/app/states/app_driveState.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (app::bspdWarning::checkSoftwareBspd(apps))
```
Lucien950, line 49: the paradigm should be
1. check for warnings
2. if there are warnings, don't drive
this is a unnecessary edge case

firmware/hexray/VC/src/app/states/app_driveState.cpp
```diff
@@ -0,0 +1,93 @@
...
+    app::tv::algo::run(apps);
+}
+
+static void driveStateRunOnExit(void)
+{
+    // TODO:
+    // disable inverters
+    inverter_enable_toggle(false, false, false, false);
+    set_torque_limit_negative(NO_TORQUE_Nm, NO_TORQUE_Nm, NO_TORQUE_Nm, NO_TORQUE_Nm);
+    set_torque_limit_positive(NO_TORQUE_Nm, NO_TORQUE_Nm, NO_TORQUE_Nm, NO_TORQUE_Nm);
```
Lucien950, line 92: surely we don't need a macro for 0

Lucien950 (changes requested):

firmware/hexray/VC/src/app/states/app_driveState.cpp
Aditya-Dhiman4: is that not what happens? precheck checks for any warning and if it fails (thus there is a warning) we do not drive

firmware/hexray/VC/src/app/states/app_driveState.cpp
Aditya-Dhiman4: yeah i realized FSM does handle all of this it would just give us 0 pedal percent ill fix

firmware/hexray/VC/src/app/states/app_driveState.cpp
Aditya-Dhiman4: how is bspd an unecessary edge case?

firmware/hexray/VC/src/app/states/app_driveState.cpp
Aditya-Dhiman4: we had a macro for 0 torque in like each individual state, i centralized it so its less messy

firmware/hexray/VC/src/app/states/drive_state.cpp
Aditya-Dhiman4: mi bomboo

firmware/hexray/VC/src/app/app_sbgEllipse.cpp
Aditya-Dhiman4: So u would want broadcast to return an error code? plus the reason why I have these getters is so we dont need to use the CAN api

firmware/hexray/VC/src/app/torque_vectoring/driving_algorithm.hpp
Aditya-Dhiman4: I mean i think this is easily adaptable to the controls algo no? we would basically pass in vehicle state into the algo along with inputs, or have vehicle state include the same members as inputs. So in `optimal_tv()` you would just define your state by collecting all of your sensor data and driver inputs, then call `estimate` (this would be the vehicle state estimator call), then `update`, then `kappa_update`? Plus the `send_torque` and `send_speed` functions would just use wheelsets instead of 4 different inputs args

Aditya-Dhiman4: the main changes would end up consisting in the functions that run each algo anyways

PR #1976, opened 2026-04-26
FIRM-167 Hexray SoC

KelmLelm:
### Changelist 

- Added SoC app layer and SoC storage IO layer.
- Added SD card HW layer
- Changed CMakeLists to include logfs in the build target
NOTE: LUT for cells and function calls for segments need to be replaced when each are done. Lastly I still gotta fix test so they build 🥀 
Since the "C" version of the file system references the "C" functions for hw_sd_present(), had to call hw directly from add layer. Having the new C++ version should fix.

### Resolved Tickets

[FIRM-167](https://ubcformulaelectric.atlassian.net/browse/FIRM-167) 
[FIRM-367](https://ubcformulaelectric.atlassian.net/browse/FIRM-367)

[FIRM-167]: https://ubcformulaelectric.atlassian.net/browse/FIRM-167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FIRM-367]: https://ubcformulaelectric.atlassian.net/browse/FIRM-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


firmware/hexray/BMS/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ file(GLOB_RECURSE IO_SRCS CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/io/
 list(APPEND IO_SRCS
         "${SHARED_IO_INCLUDE_DIR_CPP}/io_time.cpp"
         "${SHARED_IO_INCLUDE_DIR_CPP}/io_shdnLoopNode.cpp"
+        "${SHARED_IO_INCLUDE_DIR}/io_fileSystem_logfs.c"
```
Lucien950: cpp please :)

Lucien950: actually i think emily is cooking this rn

firmware/hexray/BMS/src/app/app_soc.hpp
Lucien950, line 1: how many of these functions actually need to be exposed to outside callers? Make sure to expose the smallest possible surface area to other modules to make it easier to reason about

firmware/hexray/BMS/src/app/app_soc.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <cstdint>
+
+namespace app::soc
+{
+
+inline constexpr uint32_t AH_TO_COULOMBS  = 3600.0f;
+inline constexpr float    STATE_OF_HEALTH = 1.0f;
+// TODO: Replace with cell capacity and cells in parallel constants in app_segments once that's finished
+inline constexpr float SERIES_ELEMENT_FULL_CHARGE_C = 3.0f * AH_TO_COULOMBS * 4.0f * STATE_OF_HEALTH;
```
Lucien950: like this does this need to be public?

firmware/hexray/BMS/src/app/app_soc.hpp
Lucien950: cuz lowkey, outside of init, get soc, write soc to disk, and broadcast, the rest of the functions do not need to be public

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+        ++lut_index;
+    }
+
+    if (lut_index == V_TO_SOC_LUT_SIZE)
+    {
+        lut_index = V_TO_SOC_LUT_SIZE - 1U;
+    }
+
+    return ocv_soc_lut[lut_index];
+}
```
Lucien950, line 113: binary search please

Lucien950: the voltages are monotonic

Lucien950: just chat gpt it bruh

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void init()
+{
+    soc_prev_current_A = 0.0f;
+    soc_is_corrupt     = true;
+    soc_charge_c       = -1.0;
+
```
Lucien950: put this in the else block to be more explicit

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    soc_charge_c = static_cast<double>(soc_percent / 100.0f * SERIES_ELEMENT_FULL_CHARGE_C);
+
+    soc_is_corrupt = true;
+}
+
+bool readSocFromSd(float &saved_soc_percent)
+{
+    return io::socStorage::read(saved_soc_percent);
+}
```
Lucien950: return a expected here

firmware/hexray/BMS/src/io/io_socStorage.hpp
```diff
@@ -0,0 +1,30 @@
...
+ * Return whether or not SoC persistence storage is available.
+ */
+bool isAvailable();
+
+/**
+ * Read the latest stored SoC value.
+ * @param saved_soc_percent output SoC value in percent
+ * @return true if the read succeeded
+ */
+bool read(float &saved_soc_percent);
```
Lucien950: return expected here

firmware/hexray/BMS/src/io/io_socStorage.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Persist the SoC value to filesystem storage.
+ * @param soc_percent SoC value in percent
+ * @return true if write and sync succeeded
+ */
+bool write(float soc_percent);
```
Lucien950: return expected here

firmware/hexray/BMS/src/io/io_socStorage.hpp
Lucien950, line 1: lowkey move this up to app layer and use io_fileSystem emily is developing

firmware/hexray/BMS/src/tasks.cpp
```diff
@@ -108,12 +122,34 @@
     }
 }
 
+[[noreturn]] static void tasks_runSdCard(void *arg)
+{
+    const uint32_t                 period_ms                = 1000U;
+    const uint32_t                 watchdog_grace_period_ms = 50U;
+    hw::watchdog::WatchdogInstance watchdog_sdCard =
```
Lucien950: lowkey don't watchdog this shit

Lucien950: everything preempts it anyways

firmware/hexray/BMS/src/tasks.cpp
```diff
@@ -122,6 +158,7 @@ void BMS_StartAllTasks()
     Task100Hz.start();
     TaskCanRx.start();
     TaskCanTx.start();
+    TaskSdCardHandle = static_cast<TaskHandle_t>(TaskSdCard.start());
```
Lucien950: does the handle not get stored in the task object? If not it probs should

firmware/hexray/BMS/src/io/io_socStorage.cpp
```diff
@@ -0,0 +1,120 @@
...
+    if (io_fileSystem_read(soc_fd, &last_soc_percent, sizeof(last_soc_percent)) != FILE_OK)
+    {
+        constexpr float invalid_soc_percent = -1.0f;
+        if (!criticalCheck(
+                io_fileSystem_write(soc_fd, (void *)&invalid_soc_percent, sizeof(invalid_soc_percent)) == FILE_OK))
+        {
+            return;
+        }
+        (void)criticalCheck(io_fileSystem_sync(soc_fd) == FILE_OK);
+    }
```
Lucien950: criticalCheck seems like a very bad pattern. just do a bunch of short circuiting the bool conditions and you should be chilling

Lucien950: if(a and b and c and ... ){ happy path }
else { error path }
or
if(not a or not b or not c ...) {error path, return}
happy path

firmware/hexray/BMS/src/io/io_socStorage.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (soc_storage_error_remaining > 0U)
+        {
+            --soc_storage_error_remaining;
+        }
+        LOG_ERROR("SoC storage operation failed");
+        return false;
+    }
+    return true;
+}
```
Lucien950: delete both of these

firmware/hexray/BMS/src/io/io_socStorage.cpp
```diff
@@ -0,0 +1,120 @@
...
+        return false;
+    }
+
+    if (!operationCheck(io_fileSystem_write(soc_fd, (void *)&soc_percent, sizeof(soc_percent)) == FILE_OK))
+    {
+        return false;
+    }
+
+    return operationCheck(io_fileSystem_sync(soc_fd) == FILE_OK);
+}
```
Lucien950: same comment with criticalCheck

firmware/hexray/BMS/src/io/io_socStorage.cpp
Lucien950, line 1: for a wrapper around the file system this is doing way too much

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+        io::irs::positiveState() == app::can_utils::ContactorState::CONTACTOR_STATE_CLOSED)
+    {
+        updateSocStats();
+    }
+}
+
+bool getSocToSave(uint32_t &soc_tenths)
+{
+    const float min_soc_percent = getMinSocPercent();
+    if (!IS_IN_RANGE(0.0f, 100.0f, min_soc_percent))
```
Lucien950: just write it out bluddy 🥀

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// Charge in cell in coulombs.
+double soc_charge_c = -1.0;
+
+// Charge loss at time t-1.
+float soc_prev_current_A = 0.0f;
+
+// Indicates if SOC from SD was corrupt at startup.
+bool soc_is_corrupt = true;
```
Lucien950: can you call it dirty or something corrupt makes it seem like it's busted

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    const float current = app::ts::getCurrent();
+
+    const uint32_t current_time_ms = io::time::getCurrentMs();
+    const uint32_t elapsed_time_ms = current_time_ms - soc_prev_update_ms;
+    soc_prev_update_ms             = current_time_ms;
+
+    const double elapsed_time_s = static_cast<double>(elapsed_time_ms) * MS_TO_S;
+
+    app::math::trapezoidalRule(soc_charge_c, soc_prev_current_A, current, elapsed_time_s);
```
Lucien950, line 112: can you refactor trap rule to += to soc_charge_c
also make sure to update soc_prev_current_A

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void resetSocFromVoltage()
+{
+    // TODO: Use minimum cell voltage once app::segments is migrated in Hexray BMS.
+    soc_is_corrupt = true;
+}
+
+void resetSocCustomValue(const float soc_percent)
+{
+    soc_charge_c = static_cast<double>(soc_percent / 100.0f * SERIES_ELEMENT_FULL_CHARGE_C);
```
Lucien950: lowkey just take in a soc_percent as a double and just do double math

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+float getMinSocCoulombs()
+{
+    return static_cast<float>(soc_charge_c);
+}
+
+float getMinSocPercent()
+{
+    return (static_cast<float>(soc_charge_c) / SERIES_ELEMENT_FULL_CHARGE_C) * 100.0f;
```
Lucien950: i imagine you want to work with doubles until the last possible moment, then float it?

firmware/hexray/BMS/src/app/app_soc.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    float saved_soc_percent = -1.0f;
+    if (readSocFromSd(saved_soc_percent) && IS_IN_RANGE(0.0f, 100.0f, saved_soc_percent))
+    {
+        soc_last_saved_tenths = socPercentToTenths(saved_soc_percent);
+        return soc_last_saved_tenths;
+    }
+
+    return SOC_TENTHS_UNAVAILABLE;
+}
```
Lucien950: is this not duplicated with init code?

firmware/hexray/BMS/src/app/app_soc.hpp
Lucien950: also, writing to sd card and figuring out SOC is two seperate responsibilities. please don't conflate them in one module

Lucien950 (changes requested):
i don't think this is a full implementation, because i don't think the unhappy path is not well covered. please make sure that is implemented and tested

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: aligned

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: fineeeee

firmware/hexray/BMS/CMakeLists.txt
KelmLelm: Yeah I figured, couldn't find

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: you make some strong points

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: for some reason prev current is updated inside the trapazoidal rule function lol

KelmLelm: wait, doesn't trap rule already do += for soc_charge_c

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: I don't really see why you need  to have it as double outside of the accumulation part...

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: Is there something wrong with casting like this? I don't see the gain in making it double

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: ah yes, expected comment

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: Yeah, good point

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: I refuses :))

firmware/hexray/BMS/src/io/io_socStorage.cpp
KelmLelm: Okay, moved up to app layer so it is doing the applying frfr

firmware/hexray/BMS/src/app/app_soc.hpp
KelmLelm: I thought I'd leave the constants here since it can be a pain to redefine them and keep them consistent if we want them for testing

firmware/hexray/BMS/src/app/app_soc.cpp
Lucien950: you lose precision when you divide, which gets rid of all the precision that you build up using doubles in accumulation

firmware/hexray/BMS/src/app/app_soc.cpp
Lucien950: because you want to do all the dividing and whatever in double rather than float

firmware/hexray/BMS/CMakeLists.txt
Lucien950: please use hers when it's done

firmware/hexray/BMS/CMakeLists.txt
```diff
@@ -39,10 +40,10 @@ list(APPEND HW_SRCS
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_spi.cpp"
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_pwmOutput.cpp"
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_can_hx.cpp"
+        "${SHARED_HW_INCLUDE_DIR_CPP}/hw_sd.cpp"
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_watchdog.cpp"
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_bootup.cpp"
         "${SHARED_HW_INCLUDE_DIR_CPP}/hw_resetReason.cpp"
-        "${SHARED_HW_INCLUDE_DIR_CPP}/hw_bootup.cpp"
```
Lucien950, line 45: wut

firmware/hexray/BMS/src/app/app_socStorage.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Return whether or not SoC persistence storage is available.
+ */
+bool isAvailable();
+
+/**
+ * Return the result of the latest filesystem operation attempted by SoC storage.
+ */
+FileSystemError getStatus();
```
Lucien950: this should not return seperately but with the last operation

firmware/hexray/BMS/src/app/app_socStorage.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @brief Get the last SoC value written to SD card in tenths of a percent.
+ */
+uint32_t getLastWrittenSocTenths();
+
+/**
+ * @brief Get the SoC value to save to SD in tenths of a percent when valid.
+ */
+bool convertSocToTenths(float soc_percent, uint32_t &soc_tenths);
```
Lucien950:
```suggestion
/**
 * @brief Get the SoC value to save to SD in tenths of a percent when valid.
 */
std::expected<uint32_t, ErrorCode> convertSocToTenths(float soc_percent);
```

firmware/hexray/BMS/src/app/app_socStorage.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @brief Read SoC value from SD storage.
+ */
+bool readSocFromSd(float &saved_soc_percent);
+
+/**
+ * @brief Write SoC value to SD storage.
+ */
+bool writeSocToSd(float soc_percent);
```
Lucien950: please use expected here as well

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    soc_storage_status = status;
+    if (status != FILE_OK)
+    {
+        LOG_ERROR("File system error: %d", status);
+        return std::unexpected(status);
+    }
+
+    return {};
+}
```
Lucien950: I don't think you need to save this

firmware/hexray/BMS/src/app/app_socStorage.cpp
Lucien950, line 1: rather than saving the error into another variable which you check later (whenever the fuck and not enforceable) please return the error immediately with each function which throws the error

also if you do this you have no idea which function and how it threw the error when debugging, you will only know the type of the error

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+std::expected<void, FileSystemError> read(float &saved_soc_percent)
+{
+    saved_soc_percent = -1.0f;
+
+    if (auto result = requireStorageReady(); !result)
+    {
+        return result;
+    }
```
Lucien950: return if err would work here

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+std::expected<void, FileSystemError> requireStorageReady()
+{
+    if (!soc_storage_ready)
+    {
+        return std::unexpected(soc_storage_status);
+    }
+
+    return {};
+}
```
Lucien950: for example here if the storage is not ready you don't need to know why it's not ready, just that it isn't
also this is highkey just a sanity check, and should not be part of your hot path
and also the error should probably be handled where it is thrown so the type of the error is not of interest to the consumer of this function

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+        (void)recordStatus(FILE_MOUNT_FAILED);
+        LOG_ERROR("SoC storage unavailable: SD card not present");
+        return;
+    }
+
+    if (auto result = recordStatus(io_fileSystem_init()); !result)
+    {
+        LOG_ERROR("SoC storage unavailable: failed to initialize file system");
+        return;
+    }
```
Lucien950: i would decouple this functionality from soc and put it into tasks, because this is a global property

Lucien950: you should still check if this fails (which is kept track of with io_fileSystem) before continuing

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <cstdint>
+#include <expected>
+
+namespace
+{
+constexpr const char *SOC_STORAGE_PATH       = "/soc.txt";
+constexpr uint32_t    SOC_TENTHS_UNAVAILABLE = UINT32_MAX;
+
+FileSystemError soc_storage_status    = FILE_MOUNT_FAILED;
+bool            soc_storage_ready     = false;
```
Lucien950: i think you can implicitly store this by making soc_fd an optional type, which forces an unwrap

firmware/hexray/BMS/src/app/app_socStorage.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    soc_storage_ready = false;
+
+    // TODO: Replace once cpp version of filesystem is implemented
+    if (!sd1.sdPresent())
+    {
+        (void)recordStatus(FILE_MOUNT_FAILED);
+        LOG_ERROR("SoC storage unavailable: SD card not present");
+        return;
+    }
```
Lucien950: yeah you do not need to check this this is a io_filesystem responsibility

firmware/hexray/BMS/src/app/app_socStorage.cpp
Lucien950, line 1: on the lowest of keys what is stopping us from storing the SOC from just inside app_soc?

Lucien950: cause i think this can be massively simplified and just slapped in there

Lucien950 (changes requested):

firmware/hexray/BMS/src/tasks.cpp
```diff
@@ -16,11 +19,26 @@
...
+[[noreturn]] static void tasks_runCanTx(void *arg);
+[[noreturn]] static void tasks_runSdCard(void *arg);
+
+// Define the task with StaticTask template class
+static hw::rtos::StaticTask<512> Task1kHz(osPriorityRealtime, "Task1kHz", tasks_run1kHz);
+static hw::rtos::StaticTask<512> Task1Hz(osPriorityAboveNormal, "Task1Hz", tasks_run1Hz);
+static hw::rtos::StaticTask<512> Task100Hz(osPriorityHigh, "Task100Hz", tasks_run100Hz);
+static hw::rtos::StaticTask<512> TaskCanRx(osPriorityBelowNormal, "TaskCanRx", tasks_runCanRx);
+static hw::rtos::StaticTask<512> TaskCanTx(osPriorityBelowNormal, "TaskCanTx", tasks_runCanTx);
+static hw::rtos::StaticTask<128> TaskSdCard(osPriorityLow, "TaskSdCard", tasks_runSdCard);
```
Lucien950: w cpu monitoring prefire??

firmware/hexray/BMS/src/tasks.cpp
```diff
@@ -123,12 +148,27 @@
...
+        {
+            const uint32_t last_written_soc      = app::socStorage::getLastWrittenSocTenths();
+            const bool     soc_storage_available = app::socStorage::isAvailable();
+            if (soc_storage_available && last_written_soc != rounded_soc)
+            {
+                app::socStorage::writeSocToSd((float)rounded_soc / 10.0f);
+            }
+        }
+        start_ticks += period_ms;
+        osDelayUntil(start_ticks);
```
Lucien950, line 296: you don't need to do this just wait with infinite timeout lmao

firmware/hexray/BMS/src/tasks.cpp
```diff
@@ -123,12 +148,27 @@
...
+    forever
+    {
+        uint32_t rounded_soc = 0U;
+        if (xTaskNotifyWait(0, ULONG_MAX, &rounded_soc, 0) == pdTRUE)
+        {
+            const uint32_t last_written_soc      = app::socStorage::getLastWrittenSocTenths();
+            const bool     soc_storage_available = app::socStorage::isAvailable();
+            if (soc_storage_available && last_written_soc != rounded_soc)
+            {
+                app::socStorage::writeSocToSd((float)rounded_soc / 10.0f);
```
Lucien950: LOG_IF_ERR perhaps?

Lucien950 (changes requested):

Lucien950:
also I swear you are exposing a whole bunch of functions you don't use, if you don't use them please don't expose them

firmware/hexray/BMS/CMakeLists.txt
KelmLelm: yeye

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: The double used for the accumulation though is just to make it possible to have a small coulomb delta change the state of a much larger number by comparison. The precision on the output of the number at any given point does not really matter since the LUT itself is only precise down to the 0.5%.

firmware/hexray/BMS/CMakeLists.txt
KelmLelm: I forgor if it's because it wasn't building or something

firmware/hexray/BMS/src/tasks.cpp
KelmLelm: indeed

firmware/hexray/BMS/src/tasks.cpp
KelmLelm: perhap

firmware/hexray/BMS/src/tasks.cpp
KelmLelm: you don't want a cap on the period this function can run for?

firmware/hexray/BMS/src/app/app_socStorage.hpp
KelmLelm: expected zedwin comment

firmware/hexray/BMS/src/app/app_socStorage.hpp
KelmLelm: Understandable

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: Understand

firmware/hexray/BMS/src/app/app_soc.cpp
KelmLelm: understand

firmware/hexray/BMS/src/app/app_soc.hpp
KelmLelm: getdirty I forsee being needed by current limiting if it is to determine whether or not it can use soc to limit on. Got rid of some unused ones I can't forsee a use for (min ocv from soc). Moved some to internal (getminsoccoulombs) so that they can be broadcasted. Also forgot to broadcast dirty status so I added that too,

firmware/hexray/BMS/src/app/app_soc.hpp
KelmLelm: understand

firmware/hexray/BMS/src/app/app_socStorage.cpp
KelmLelm: oki

firmware/hexray/BMS/src/app/app_socStorage.cpp
KelmLelm: ooki

firmware/hexray/BMS/src/app/app_socStorage.cpp
KelmLelm: you got it chief

firmware/hexray/BMS/src/app/app_socStorage.cpp
KelmLelm: understandable

firmware/hexray/BMS/src/app/app_socStorage.cpp
KelmLelm: wait, the checkmount on the filesystem class is private tho how can I see feedback of the init?

firmware/hexray/BMS/src/jobs.cpp
```diff
@@ -72,11 +73,11 @@ void jobs_init()
     can_rx_queue.init();
     can_tx_queue.init();
 
-    app::can_tx::BMS_Hash_set(GIT_COMMIT_HASH);
```
KelmLelm, line 70: whoops

Lucien950, line 70:
```suggestion
    app::can_tx::BMS_Hash_set(GIT_COMMIT_HASH);
```

Lucien950 (approved):

PR #1982, opened 2026-04-28
Lib Layer

Lucien950:
### Changelist 

The idea of the lib layer is to add everything which must be included with all builds. This way, no boilerplate gets forgetten when bringing up boards. This provides a strong distinction between the items in the hardware layer which interact with peripherals and the standard boilerplate which must be included with every binary.

If you need to #include something, it DOES NOT belong in this layer. This is for the files which weak override only.

### Resolved Tickets

[FIRM-649](https://ubcformulaelectric.atlassian.net/browse/FIRM-649)

[FIRM-649]: https://ubcformulaelectric.atlassian.net/browse/FIRM-649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Aditya-Dhiman4 (approved):
Image

PR #1984, opened 2026-04-30
Power Monitoring Proper Refactor

rijwal:
### Testing Done
Voltage, current, and power readings are all good, OV and UV limits have been tested and confirmed. The only note to keep is that we need to perchance look into a different power monitoring chip because ours can not manage the fact that channels may be off and recognize that itself. if thats even a thing?? We also need to keep a note to seperate the number of pin that we are routing to external interrupts because we cannot configure them together as EXTIs if so, only one can be done. 

### Resolved Tickets

[FIRM-654](https://ubcformulaelectric.atlassian.net/browse/FIRM-654)

[FIRM-654]: https://ubcformulaelectric.atlassian.net/browse/FIRM-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


firmware/hexray/VC/src/app/app_powerMonitoring.cpp
```diff
@@ -18,44 +19,74 @@ void update()
 
     for (int tries = 0; not init_done and tries < 10; tries++)
     {
-        init_done = io::powerMonitoring::init();
+        init_done = io::powerMonitoring::init().has_value();
```
AmirTajaddodi, line 14: init here is not returnig a true or false but a void and errocode so becareful

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
```diff
@@ -18,44 +19,74 @@ void update()
...
     }
     if (not init_done)
     {
         // error handle
```
AmirTajaddodi, line 18: just returing is a little sus coz it makes debugging a nightmare. We want some indication to know where the code failed

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
```diff
@@ -18,44 +19,74 @@ void update()
...
-    app::can_tx::VC_ChannelOnePower_set(ch1_power);
-    app::can_tx::VC_ChannelTwoPower_set(ch2_power);
-    app::can_tx::VC_ChannelThreePower_set(ch3_power);
-    app::can_tx::VC_ChannelFourPower_set(ch3_power);
+    if (!io::powerMonitoring::refresh().has_value())
+    {
+        return;
+    }
+
+    if (io::powerMonitoring::is_alert_asserted())
```
AmirTajaddodi: I thought we are gonna hook up the alert pin on interrupt?

AmirTajaddodi (dismissed):

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
rijwal: We cant do that for either one of the power monitoring or the battery monitoring because they are have the same gpio pin number (15), we can only have one exti pin corresponding to a pin number. We originally had them on separate lines but we needed more adc pins so then things got moved around. I already told deven that for next year we should have them on seperate pins so we can have interrupt driven pins for both.

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
rijwal: hasvalue() checks for if the fucntion returns as expected, so i would assume it would return a true if void and a false if errorcode and i have the check for that right after (ill fix the blank return). ill look into this and make sure

firmware/hexray/VC/src/jobs.cpp
```diff
@@ -4,12 +4,14 @@
...
 #include "app_heartbeatMonitors.hpp"
 #include "app_canRx.hpp"
+#include "app_powerMonitoring.hpp"
 
 #include "io_canMsg.hpp"
 #include "io_canTx.hpp"
 #include "io_canQueues.hpp"
 #include "io_time.hpp"
 #include "io_canReroute.hpp"
+#include "io_powerMonitoring.hpp"
```
Lucien950: not necessary but it's fine

firmware/hexray/VC/src/io/io_powerMonitoring_datatypes.hpp
Lucien950, line 1: bruh

Lucien950: i think this is doing slightly too much

Lucien950: also in general if you have variables in header files they need to be marked inline

can_bus/hexray/VC/VC_tx.json
```diff
@@ -769,6 +769,30 @@
                 "min": 0.0,
                 "max": 800.0,
                 "unit": "W"
+            },
+            "Alert_OV_Channel1": {
```
Lucien950: can you call it CHANNEL1_OV/UV

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
```diff
@@ -1,61 +1,87 @@
...
+    const auto ch2_power = io::powerMonitoring::read_power(CH2);
+    const auto ch3_power = io::powerMonitoring::read_power(CH3);
+    const auto ch4_power = io::powerMonitoring::read_power(CH4);
+
+    if (!ch1_voltage.has_value() || !ch2_voltage.has_value() || !ch3_voltage.has_value() || !ch4_voltage.has_value() ||
+        !ch1_current.has_value() || !ch2_current.has_value() || !ch3_current.has_value() || !ch4_current.has_value() ||
+        !ch1_power.has_value() || !ch2_power.has_value() || !ch3_power.has_value() || !ch4_power.has_value())
+    {
+        return std::unexpected(ErrorCode::ERROR);
+    }
```
Lucien950: individual errors please

firmware/hexray/VC/src/io/io_powerMonitoring.cpp
```diff
@@ -43,90 +40,153 @@ std::expected<void, ErrorCode> write_register(uint16_t reg, std::span<const uint
     return {};
 }
 
-void refresh()
+std::expected<void, ErrorCode> refresh(void)
```
Lucien950: no voids in function arguments for C++

firmware/hexray/VC/src/io/io_powerMonitoring.cpp
```diff
@@ -43,90 +40,153 @@ std::expected<void, ErrorCode> write_register(uint16_t reg, std::span<const uint
...
 {
-    constexpr uint8_t cmd = 0x00;
+    const uint8_t cmd = REG_REFRESH;
     LOG_IF_ERR(pwr_pump.transmit((std::span{ &cmd, 1 })));
-    io::time::delay(1);
+    return {};
 }
 
-bool init()
+std::expected<void, ErrorCode> init(void)
```
Lucien950: can you just keep retrying? cause this is very critical

Lucien950: also this is highkey app layer shit

firmware/hexray/VC/src/io/io_powerMonitoring.cpp
Lucien950, line 1: be a bit more careful about what is in the io layer and in the app layer

Lucien950 (changes requested):

firmware/hexray/VC/src/io/io_powerMonitoring.hpp
```diff
@@ -1,19 +1,29 @@
...
-float read_current(uint8_t ch);
-float read_power(uint8_t ch);
-} // namespace io::powerMonitoring
\ No newline at end of file
+// bitmask: OV (bits 0,1, 2, 3), UV (bits 4, 5, 6, 7)
+std::expected<void, ErrorCode>    refresh();
+std::expected<void, ErrorCode>    init();
+std::expected<float, ErrorCode>   read_voltage(uint8_t ch);
+std::expected<float, ErrorCode>   read_current(uint8_t ch);
+std::expected<float, ErrorCode>   read_power(uint8_t ch);
```
Lucien950: since you already defined the channel above can you just use the enum as the type of the argument for this function?

firmware/hexray/VC/src/app/app_powerMonitoring.cpp
```diff
@@ -2,60 +2,103 @@
...
+    if (!ch4_current.has_value())
+        return std::unexpected(ErrorCode::ERROR);
+    if (!ch1_power.has_value())
+        return std::unexpected(ErrorCode::ERROR);
+    if (!ch2_power.has_value())
+        return std::unexpected(ErrorCode::ERROR);
+    if (!ch3_power.has_value())
+        return std::unexpected(ErrorCode::ERROR);
+    if (!ch4_power.has_value())
+        return std::unexpected(ErrorCode::ERROR);
```
Lucien950: ok no what I meant by this is that you read as many as you can and return error at the end

Lucien950 (changes requested):
two last things

Lucien950 (approved):
W

PR #1992, opened 2026-05-03
Feedforward power limiting

efabiyi:
## Change List

- **motor_table.h** : AMK A2370DD characterization data as compile-time constexpr arrays. 
- **motor_table.cpp**: interpolation logic for the lookup tables
- **current_limiter.hpp/ .cpp**: Current limiting. Call `init()` once at startup, `update()` every control cycle. Takes per-wheel (speed, torque, temperature, measured current) and outputs per-wheel (speed command, positive torque limit, negative torque limit). Assumes arrays are monotonically increasing. The feedback term trims the scaling factor and accounts for steady state drift and bias.

## Known Limitations 
- Prediction accuracy depends heavily on the quality of inverter characterization data, as well as our own state estimation
- Feedback trim gain Kp needs to be tuned


firmware/hexray/VC/src/app/torque_vectoring/controllers/power_limiting/motor_table.hpp
Lucien950, line 1: bomboclatt

Lucien950: maybe put the humungous table into cpp file so that every single hpp which includes this is not immediately ddosd

Lucien950: also do double check that we do have the flash memory to be able to do this

efabiyi: this is very true will do

firmware/hexray/VC/src/app/torque_vectoring/controllers/power_limiting/motor_table.hpp
```diff
@@ -0,0 +1,3668 @@
...
+static constexpr uint16_t N_SPEED   = 201; // number of  speed entries
+static constexpr uint8_t  N_CURRENT = 21;  // number of  current entries
+
+static constexpr float TEMP_POINTS_C[N_TEMP] = { 80.0f, 100.0f, 120.0f };
+static constexpr float SPEED_STEP_RPM        = 100.0f;
+static constexpr float CURRENT_STEP_A        = 5.25f;
+static constexpr float SPEED_MAX_RPM         = 20000.0f;
+static constexpr float CURRENT_MAX_A         = 105.0f;
+
+static constexpr float RPM_TO_RAD_PER_SEC = (2.0f * 3.14159265358979323846f) / 60.0f;
```
Aditya-Dhiman4: we have a M_PI_F macro somewhere

Aditya-Dhiman4 (changes requested):

PR #1994, opened 2026-05-03
VC Inshallah 2.0

AmirTajaddodi:


Lucien950:
2.0 is crazy work

KelmLelm:
We're going for round two? 😭

can_bus/hexray/VC/VC_rx.json
KelmLelm: Power limit to be subbed out by the current limit just by the by

firmware/hexray/VC/src/app/states/app_bmsOnState.cpp
```diff
@@ -0,0 +1,54 @@
...
+    {
+        static const app::powerManager::PowerManagerConfig power_manager_state = { .efuse_configs = { {
+                                                                                       { false, 0, 5 }, // rr_pump
+                                                                                       { false, 0, 5 }, // rl_pump
+                                                                                       { false, 0, 5 }, // r_rad_fan
+                                                                                       { false, 0, 5 }, // l_rad_fan
+                                                                                       { true, 0, 5 },  // f_inv
+                                                                                       { true, 0, 5 },  // r_inv
+                                                                                       { true, 0, 5 },  // rsm
+                                                                                       { true, 0, 5 },  // bms
```
Aditya-Dhiman4: Looking at the previous states, Init and Inverter On both turn on the BMS. If that is intended shouldnt this state be named like BMS ready, else we should not be turning on BMS in Init and Inverter On

firmware/hexray/VC/src/app/states/app_hvInitState.cpp
```diff
@@ -0,0 +1,235 @@
...
+        const bool inverter_dc_quit = app::can_rx::INVFL_bQuitDcOn_get() && app::can_rx::INVFR_bQuitDcOn_get() &&
+                                      app::can_rx::INVRL_bQuitDcOn_get() && app::can_rx::INVRR_bQuitDcOn_get();
+
+        const bool inverter_invOn_quit =
+            app::can_rx::INVFL_bQuitInverterOn_get() && app::can_rx::INVFR_bQuitInverterOn_get() &&
+            app::can_rx::INVRL_bQuitInverterOn_get() && app::can_rx::INVRR_bQuitInverterOn_get();
+
+        for (bool retry_transition = true; retry_transition;)
+        {
+            retry_transition = false;
```
Aditya-Dhiman4: this logic regarding retry_transition can be cleaned up. instead of setting it to false by default, let the loop set it to true at the beginning, and in the conditionals and case statements set it to false

firmware/hexray/VC/src/app/states/app_pcmOnState.cpp
```diff
@@ -0,0 +1,138 @@
...
+                                                                                       { false, 200, 5 }, // r_rad_fan
+                                                                                       { false, 200, 5 }, // l_rad_fan
+                                                                                       { true, 0, 5 },    // f_inv
+                                                                                       { true, 0, 5 },    // r_inv
+                                                                                       { true, 0, 5 },    // rsm
+                                                                                       { true, 0, 5 },    // bms
+                                                                                       { true, 0, 5 },    // dam
+                                                                                       { true, 0, 5 },    // front
+                                                                                   } } };
+        app::powerManager::updateConfig(power_manager_state);
```
setaremalekiii, line 58: Would it be better to use getConfig function and only updateConfig if the efuse table has a different config? I think it might save some time since for example bmsOnState has the same config as pcmOnState

firmware/hexray/VC/src/app/states/app_inverterFaultHandlerState.cpp
```diff
@@ -0,0 +1,56 @@
...
+        {
+            case inverter::FaultHandlerState::INV_FAULT_RECOVERED:
+            {
+                app::StateMachine::set_next_state(inverter::recovery_state());
+                break;
+            }
+            case inverter::FaultHandlerState::INV_FAULT_LOCKOUT:
+            {
+                app::can_tx::VC_Fault_InvLockoutFault_set(true);
+                app::StateMachine::set_next_state(&fault_state);
```
setaremalekiii: do we need a fault_state? for lock out faults theres a need to powercycle the VC anyways so it will start from bmsOnState.

firmware/hexray/VC/src/app/states/app_inverterFaultHandlerState.cpp
```diff
@@ -0,0 +1,56 @@
...
+        app::can_alerts::infos::InverterRetry_set(true);
+    }
+
+    static void runOnTick100Hz(void)
+    {
+        switch (inverter::FaultHandler())
+        {
+            case inverter::FaultHandlerState::INV_FAULT_RECOVERED:
+            {
+                app::StateMachine::set_next_state(inverter::recovery_state());
```
setaremalekiii: recovery_state is not updating unless you call FaultCheck idk if this will work

firmware/hexray/VC/src/app/states/app_driveState.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    static void runOnTick100Hz(void) {}
+
+    static void runOnExit(void) {}
+
+} // namespace driveState
+
+State drive_state = { .name              = "Drive State",
+                      .run_on_entry      = driveState::runOnEntry,
+                      .run_on_tick_1Hz   = nullptr,
```
setaremalekiii, line 102: if this is nullptr for every state, do we really need it as part of the state object? Maybe we can just delete it completely

Lucien950 (approved):

PR #2019, opened 2026-05-13
Historical Data

jadenbn:
things of note:
- this needs to be merged in tandem with jack's branch to pull in appropraite backend changes
- the endpoint will not work currently because i've coded it to use a GET /api/v1/signal/sessions/{start}/{end} but the implementation is GET /api/v1/signal/sessions/. jack is working on the {start}/{end}
- timezone is set by environment variable


jadenbn:
moved to new pr

PR #2020, opened 2026-05-14
Frontend Enum / Alert API Patch

Eroxl:
Fixes the discrepancies between the frontend enum / alert types and the new rust backend types. Additionally adds two new mock data producers to the backend to make it easier to test against Enum and Alert signal types.

https://github.com/user-attachments/assets/0b326370-18ee-4bc3-9999-7342fca5d704


software/tracksight/frontend/lib/signals/LiveSignalStore.tsx
```diff
@@ -32,14 +32,30 @@ class LiveSignalStore extends SignalStore {
     this.waveletBuffers = new Map();
 
     socket.on("data", (payload) => {
-      const { name: signalName, timestamp, value } = payload;
+      const { name: signalName, timestamp, value, signal_type } = payload as {
```
eriklsy, line 35: Im not sure if Jaden is actually dealing with this but I think enumeration makes more sense to handle this: our backend also sends the possible signal_type of "boolean" so this should be handled and probably interpreted as an enum? 

Image

software/tracksight/frontend/lib/signals/LiveSignalStore.tsx
eriklsy, line 1: Image This is something you have to check out in your enum selector. Please confirm this compatibility 

Image

eriklsy (changes requested):
pls fix and verify ty

software/tracksight/frontend/lib/signals/LiveSignalStore.tsx
Eroxl: Sorry building didn't catch this for some reason.

software/tracksight/frontend/lib/signals/LiveSignalStore.tsx
Eroxl: I added the Boolean type to the LiveSignalStore and it'll just be parsed like an enum. The EnumSelector doesn't let you select booleans though I dont think, I can probably change that easily if you want.

eriklsy: Ya probably have the enums selector select booleans pls. Doesn't really make sense to be something numerical deals with so it prob makes sense like that for now

eriklsy (changes requested):
Comment thread

Eroxl:
image

Booleans are included w enums in the timeline now.

eriklsy (approved):

PR #2034, opened 2026-05-22
updated filter for codegen to only trigger on dam

eriklsy:
kept legacy behaviour of basic codegen file but enabled all boards from db on dam


can_bus/hexray/bus.json
```diff
@@ -66,5 +66,8 @@
   ],
   "loggers": [
     "DAMLogger"
+  ],
+  "data_capture_nodes": [
+    "DAM"
   ]
```
Lucien950, line 69: wait we already got a loggers property 😭

Lucien950 (approved):
ok but i would just use the loggers property which already exists

can_bus/hexray/bus.json
eriklsy: real!!!!!! oops

PR #2035, opened 2026-05-22
fix tiling query and source stringify

j-ackyao:
so basically, `serde_json::to_string` adds literal quotations to the string, which breaks the query. also adding the source query struct (rahhh i wanted inline so bad) to signal tiling endpoint


eriklsy (approved):
igbro

PR #2036, opened 2026-05-22
restart on disconnect/anything

eriklsy:
yurp

### DIFF
--- a/software/tracksight/rust_backend/src/main.rs
+++ b/software/tracksight/rust_backend/src/main.rs
@@ -107,7 +107,7 @@ async fn main() {
         let base_can_queue_tx = can_queue_tx.clone();
         let base_can_db = can_db.clone();
 
-        move |tasks: &mut JoinSet<(Task, Result<(), JoinError>)>| {
+        move |tasks: &mut JoinSet<(Task, Result<(), JoinError>)>, is_restart: bool| {
             let shutdown_rx_clone = base_shutdown_rx.resubscribe();
             let hc_tx_clone = base_hc_tx.clone();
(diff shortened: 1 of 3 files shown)


software/tracksight/rust_backend/src/tasks/serial_handler.rs
```diff
@@ -31,6 +32,10 @@ pub async fn run_serial_task(
     .open_native_async()
     .unwrap_or_fail_health_check(&health_check_tx, Task::SerialHandler).await;
 
+    if is_restart {
+        println!("{}", green("SerialHandler successfully restarted."));
+    }
+
```
j-ackyao: idk if this should be here, if anything it should be below where the health check is sent.

software/tracksight/rust_backend/src/main.rs
```diff
@@ -190,7 +191,7 @@ async fn main() {
                         // also worst case isn't bad, can still force quit
                         println!("Restarting {task:?} in {TASK_RESTART_DELAY_MS}ms...");
                         sleep(Duration::from_millis(TASK_RESTART_DELAY_MS)).await;
-                        serial_spawner(&mut tasks);
+                        serial_spawner(&mut tasks, true);
```
j-ackyao: let's not pass through `is_restart` as a parameter for spawning serial task. the state of whether or not this is a restart just for the sake of printing is not critical to spawning the task itself.
would recommend, instead to add a health check right after this spawn and listen for serial health check specifically, then printing out that it restarted successfully, this way you won't need to change anything in `serial_handler` regarding restarts

j-ackyao (approved):
lgtm

PR #2038, opened 2026-05-22
vbat enable

eriklsy:
rtc needed to be switched to LSE from LSI and modifications were required to the prescalers


firmware/hexray/DAM/src/tasks.cpp
Lucien950, line 1: revert?

Lucien950 (approved):
agree

firmware/hexray/DAM/src/tasks.cpp
eriklsy: oopsies!

PR #2042, opened 2026-05-26
TSIM machine + tests + wd move

eriklsy:
blink blink blink


firmware/hexray/DAM/src/app/app_tsim.cpp
```diff
@@ -0,0 +1,90 @@
...
+void init()
+{
+    io::tsim::set_off();
+    curr_color = TsimColor::OFF;
+    next_color = TsimColor::RED;
+    toggle_timer.stop();
+    bootup_ignore_timer.restart();
+}
+
+void tick()
```
Aditya-Dhiman4, line 64: Consider when would we wanna start reading imd, bspd, and bms oks. I think these only propagate a meaningful value once we reach HV Init state on the VC state machine (verify this plz). I don't know if by default these propagate OK before we reach that state.

firmware/hexray/DAM/src/app/app_tsim.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    if (no_fault)
+    {
+        io::tsim::set_green();
+        curr_color = TsimColor::GREEN;
+        toggle_timer.stop();
+        return;
+    }
+
+    switch (toggle_timer.updateAndGetState())
```
Aditya-Dhiman4, line 79: DRY Breh

Aditya-Dhiman4 (dismissed):

firmware/hexray/DAM/src/app/app_tsim.cpp
eriklsy: Fatimah says faults should trigger the light regardless of state

firmware/hexray/DAM/src/app/app_tsim.cpp
```diff
@@ -0,0 +1,104 @@
...
+    apply_tsim_color(TsimColor::OFF);
+    curr_color = TsimColor::OFF;
+    next_color = TsimColor::RED;
+    toggle_timer.stop();
+    bootup_ignore_timer.restart();
+}
+
+void tick()
+{
+    const bool ignore_fault_on_bootup = bootup_ignore_timer.updateAndGetState() == app::Timer::TimerState::RUNNING;
```
AmirTajaddodi: which faults are we specifically ignoring from the bms on bootup. It would be add to nice some comments here but I agree with the implementation

firmware/hexray/DAM/src/app/app_tsim.cpp
```diff
@@ -0,0 +1,104 @@
...
+    const bool no_fault               = (app::can_rx::BMS_BmsLatchOk_get() && app::can_rx::BMS_ImdLatchOk_get() &&
+                           app::can_rx::BMS_BspdLatchOk_get()) ||
+                          ignore_fault_on_bootup;
+
+    if (no_fault)
```
AmirTajaddodi:
```suggestion
                           app::can_rx::BMS_BspdLatchOk_get());

    if (no_fault || ignore_fault_on_bootup)
```

AmirTajaddodi: this is super nit but this is just more readable imo

firmware/hexray/DAM/test/io_tsimFake.cpp
```diff
@@ -0,0 +1,48 @@
...
+    s_last_color = Color::OFF;
+    s_call_count = 0;
+}
+
+Color lastColor()
+{
+    return s_last_color;
+}
+
+unsigned int callCount()
```
AmirTajaddodi:
```suggestion
uint32_t callCount()
```

AmirTajaddodi (changes requested):

AmirTajaddodi (approved):

PR #2043, opened 2026-05-27
Historical Data v2

jadenbn:
same as previous PR, just rebased onto master much cleaner than the previous branch could have been.


eriklsy (commented):
w historical

eriklsy (approved):
y

PR #2045, opened 2026-05-28
boot files and metadata updating

emilysxng:
- Moved SD fs init into DAM app layer
- File metadata updates with updated time upon fs init and NTP button press. Consistent with historical formatting
- 1 file per boot , named boot_x_{hash}.bin where x = boot number and hash = hash string generated by hash function
- Radio fixes (error spam fixed, won't fault on removal/bump, etc)


eriklsy (approved):
i pray

PR #2048, opened 2026-05-28
changes for mdns

eriklsy:
updated instance name and our build for frontend so we also forward our frontend, w mdns!!!!!!!!!!!


j-ackyao (approved):
yeah whatever lgtm 👍

Lucien950:
the gaoat

PR #2061, opened 2026-06-07
ADC Calibration

pranay-o:


pranay-o (commented):
not sure if filter tick should be in broadcast

firmware/hexray/BMS/src/app/app_tractiveSystem.cpp
```diff
@@ -2,19 +2,39 @@
...
 namespace app::ts
 {
 static Timer overcurrent_warning_timer{ TS_OVERCURRENT_DEBOUNCE_DURATION_MS };
 
-float getVoltage()
+static ExponentialFilter ts_voltage_filter =
+    ExponentialFilter::withCutoffFrequency(TS_VOLTAGE_CUTOFF_HZ, TS_VOLTAGE_SAMPLE_RATE_HZ);
+static float ts_filtered_voltage = 0.0f;
+
+void init()
```
Aditya-Dhiman4: you highkey do not even need an init step

firmware/hexray/BMS/src/app/app_tractiveSystem.cpp
```diff
@@ -2,19 +2,39 @@
...
+{
+    ts_filtered_voltage = io::ts::getVoltage();
+    ts_voltage_filter =
+        ExponentialFilter::withCutoffFrequency(TS_VOLTAGE_CUTOFF_HZ, TS_VOLTAGE_SAMPLE_RATE_HZ, ts_filtered_voltage);
+}
+
+void filteredVoltageTick()
```
Aditya-Dhiman4: you would have to call filtered voltage tick in jobs and broadcast, why not consolidate this into broadcast and set that static variable in broadcast., then you wont have 2 different functions you call in jobs

Lucien950 (approved):
think about HAL_ADCEx_Calibration_Start as well

pranay-o (commented):
should i calibrate all adcs?
@Lucien950

PR #2065, opened 2026-06-07
Jadenbn/telem marker

jadenbn:
add markers on the chart for can


eriklsy (approved):
lesgo telem button on!!!!

PR #2066, opened 2026-06-07
Tracksight Historical Alert Bar

Eroxl:
Restructure historical data so it's behavior is more similar to the live and mock signal stores allowing alerts to be displayed and discovered with historical data. Adds a new tag (signal type) to points in the influx database so we can query for all alerts at once.

Restyled the UI for selecting a session so it's more inline with the SD card dumping page, it's now in the top right with it's own UI and modal.

https://github.com/user-attachments/assets/e83e2901-a7e9-4a43-a534-8f931f03f2d5

Additionally fixes two bugs one related to the tool tip bar being de synced when the widgets we're mounted out of order and another with level of details not being applied to live enum data.


eriklsy:
@copilot resolve the merge conflicts in this pull request, put the error bar further right than the historical selection stuff

eriklsy:
what a waste of credits lol

eriklsy (approved):
better alert LOD + small stufff

PR #2076, opened 2026-06-11
More ADBMS!!

pranay-o:


firmware/hexray/BMS/src/jobs.cpp
KelmLelm: are we sure we want balancing state to have priority over fault state

KelmLelm (approved):
LOOKING GOOD

firmware/hexray/BMS/src/app/app_segments.hpp
```diff
@@ -1,19 +1,25 @@
...
 
+
+
 extern io::semaphore spi_bus_lock;
 extern io::semaphore health_lock;
+extern io::semaphore shared_lock;
+extern io::semaphore internal_lock;
 
 namespace app::segments
 {
```
KelmLelm, line 23: please extern fault and warning temps and voltages

PR #2095, opened 2026-06-17
BMS Temp Mux Hotfix

KelmLelm:


Lucien950:
i don't think this is the issue, this is also the perspective of elec team

PR #2100, opened 2026-06-18
adding inv error to crit screen

setaremalekiii:


firmware/hexray/CRIT/src/app/screens/app_screen_alert.cpp
Lucien950, line 1: great changes

firmware/hexray/CRIT/src/app/screens/app_screen_invError.cpp
```diff
@@ -0,0 +1,89 @@
...
+    { io::seven_seg::a, io::seven_seg::l, io::seven_seg::t, io::seven_seg::dot, io::seven_seg::dot, io::seven_seg::dot,
+      io::seven_seg::dot, io::seven_seg::dot, io::seven_seg::dot }
+};
+
+static uint16_t prev_fr = 0u;
+static uint16_t prev_rr = 0u;
+static uint16_t prev_fl = 0u;
+static uint16_t prev_rl = 0u;
+
+static void update()
```
Lucien950, line 26: this will not work, because the moment fr is set to something it will auto override everything else. please keep track of the errorcode/errored inverter in one variable

firmware/hexray/CRIT/src/app/screens/app_screen_invError.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    // priority of error codes is based on which wheels travel longer dist:
+    // 1. fr
+    // 2. rr
+    // 3. fl
+    // 4. rl
+    // If higher priority error code is on we'll just keep that on first two segs are the
+    // inverter last 5 the code from CAN
+
+    if (fr_inv_error && fr_inv_error != prev_fr)
```
Lucien950: this will display the error for one tick, then ignore it

Lucien950 (changes requested):

PR #2102, opened 2026-06-18
shooting blanks

rijwal:


firmware/hexray/BMS/src/cubemx/Inc/stm32h7xx_hal_conf.h
Lucien950, line 1: this one you have to revert to pass chimera

Lucien950 (dismissed):
remember to do crit

setaremalekiii (approved):
we really be shooting blanks!!

PR #2103, opened 2026-06-19
i love sd cards real

eriklsy:


firmware/hexray/DAM/src/app/app_sd.cpp
```diff
@@ -130,6 +132,30 @@ void requestSync()
...
+
+    // Base was already anchored in jobs_init(); the metadata write just reads it via getFastBase().
+    requestMetadataUpdate();
+    LOG_INFO("initLogFs: Metadata update requested in init");
+
+    if (const auto err = io::sd::upgrade(); !err.has_value())
+    {
+        LOG_ERROR("initLogFs: io::sd::upgrade failed: %d", static_cast<int>(err.error()));
+        return;
+    }
```
Aditya-Dhiman4, line 154: do we have to hard fail if upgrade fails? why not use 1 bit bus and if that fails then fully fail?

Aditya-Dhiman4 (changes requested):
bing bong

Aditya-Dhiman4 (approved):

PR #2118, opened 2026-07-10
ADBMS reversible isospi & command count

pranay-o:


firmware/hexray/BMS/src/io/adbms/io_adbms_commandCount.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "io_adbms.hpp"
+#include "io_semaphore.hpp"
+
+
+namespace {
+    using namespace io::adbms;
+
+    uint8_t expected_count = 0U;
+    Segments<uint32_t> mismatch_counts{};
+    io::semaphore command_count_lock { true };
```
pranay-o: The way this lock is used is kinda stupid. easy to cause a deadlock with spi lock

firmware/hexray/BMS/src/io/adbms/io_adbms_commandCount.cpp
```diff
@@ -0,0 +1,93 @@
...
+        expected_count = nextCount(expected_count);
+    }
+
+    bool check(const uint8_t seg, const uint8_t received) {
+        const io::unique_semaphore lock{ command_count_lock };
+        if (received != expected_count) {
+            mismatch_counts[seg]++;
+            return false;
+        }
+        return true;
```
pranay-o, line 82: idk if it should resync or reset. if i try to reset, it causes deadlock so keep this way or think of smth better

firmware/hexray/BMS/src/io/adbms/io_adbms_internal.cpp
```diff
@@ -208,9 +215,14 @@ Segments<result<RegBuffer>> readRegGroup(const uint16_t cmd)
             regs[segment] = unexpected(ErrorCode::CHECKSUM_FAIL);
             continue;
         }
+        
+        if (not commandCount::check(static_cast<uint8_t>(segment), segment_reg_group.cmd_count())) {
```
pranay-o, line 229: if i unplug the segment, bus will go high so would i expect 0x3F here?

pranay-o: do I care if ts happens cuz this a different issue. it would keep on trying to correct it but it will always stay 0x3F so mismatch number will keep on increasing

PR #2120, opened 2026-07-12
Compiler Optimizations for Third Party Libraries

Jugveer-Sandher:
### Changelist 
- The non third party path referenced `${ELF_NAME}` which isn't a parameter of this function (it's only defined locally in embedded_binary) :(
- Changed the optimization flag to optimize-for-size `-Os` instead of unoptimized `-O0`.
  - The optimizations could fuck with the debugger but this is only for JCAN and HAL, so maybe it's okay.
  - Could consider doing something less extreme like `-Og` instead.

### Testing Done
No optimization:
Screenshot 2026-07-12 at 1 35 14 PM

With optimization:
Screenshot 2026-07-12 at 1 33 57 PM

### DIFF
--- a/firmware/cmake/embedded.cmake
+++ b/firmware/cmake/embedded.cmake
@@ -105,14 +105,14 @@ function(embedded_object_library
                 PUBLIC
                 ${LIB_INCLUDE_DIRS}
         )
-        target_compile_options(${LIB_NAME} PRIVATE ${SHARED_GNU_COMPILER_CHECKS} -O0)
+        target_compile_options(${LIB_NAME} PRIVATE ${SHARED_GNU_COMPILER_CHECKS} -Os)
     ELSE ()
         target_include_directories(${LIB_NAME} PUBLIC ${LIB_INCLUDE_DIRS})
         target_compile_options(${LIB_NAME} PRIVATE ${SHARED_GNU_COMPILER_CHECKS_STRICT})
         if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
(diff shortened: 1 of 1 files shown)


firmware/cmake/embedded.cmake
```diff
@@ -105,14 +105,14 @@ function(embedded_object_library
                 PUBLIC
                 ${LIB_INCLUDE_DIRS}
         )
-        target_compile_options(${LIB_NAME} PRIVATE ${SHARED_GNU_COMPILER_CHECKS} -O0)
+        target_compile_options(${LIB_NAME} PRIVATE ${SHARED_GNU_COMPILER_CHECKS} -Os)
```
AmirTajaddodi, line 108: For -Os we should be rly careful coz a lot of the code base doesn't account for compiler optimizations and shit we honestly aren't very diligent with a bunch of the stuff. Would be an interesting task to give to someone to do a objdump of the binary and check see the diff of certain operations we know for a fact are not safe

Jugveer-Sandher: This should be pretty safe, the only downside would be with using a debugger and stepping through the code. Plus we are already doing this for the app code.

Could consider doing something less extreme like `-Og` it would probably still help a decent amount.

Lucien950: as long as you are not introducing undefined behaviours in your code the compiler should guarentee correct behaviour

Jugveer-Sandher: Screenshot 2026-07-12 at 8 27 00 PM

Get this monkey out of the repo!!!!

PR #2122, opened 2026-07-16
Amir Doubted Us

pranay-o:

### DIFF
--- a/firmware/hexray/BMS/src/app/states/app_prechargeDriveState.cpp
+++ b/firmware/hexray/BMS/src/app/states/app_prechargeDriveState.cpp
@@ -38,7 +38,9 @@ namespace prechargeDriveState
                 break;
             case app::precharge::State::SUCCESS:
                 io::irs::setPositive(app::can_utils::ContactorState::CONTACTOR_STATE_CLOSED);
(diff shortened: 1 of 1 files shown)


pranay-o (commented):
ur precharge is working sir

rijwal:
Mashallah

AmirTajaddodi (commented):
I cant believe we didnt have this

Lucien950 (commented):
ong is this not just a gpio write then read? like obviously it should be high? unless there is some hardware issue preventing air+ gpio from going high?

AmirTajaddodi (approved):

Lucien950:
wait gangy i am very surprised this is an issue please keep looking into rootier causes cuz lowkey i don't believe this

AmirTajaddodi:
> wait gangy i am very surprised this is an issue please keep looking into rootier causes cuz lowkey i don't believe this

nah but if think abt it is a relay not a massive contactor so it takes a while longer to close

pranay-o:
> close

wait this fix lowkey doesn’t make sense 
we need to have a way to check if the contractor actually closes. i dont think the contractor prevents the enable pin from going high

